Issue #1, opened 2021-02-18
CLI Specification


jotamusik:
WIP
**Setting up a context**
The user will be able to set a specification file once and all actions after setting it, will be related to that specification file

## Validation
Setup an option to be able to watch a specification file and reload the output when the file change

## Some examples (from @derberg):
```
asyncapi set context ./asyncapi.yml
asyncapi validate --context=./asyncapi.yml
asyncapi validate #in case context was set
asyncapi create --context=./asyncapi.yml 
asyncapi get channels --context=./asyncapi.yml
asyncapi add channel
asyncapi new #new is not really a verb nor action but makes more sense than create I think

asyncapi generate --template=@asyncapi/html-template --context=./asyncapi.yml --output=./folder
asyncapi generate --model=java --context=./asyncapi.yml --output=./folder
asyncapi generate --model=java --context=./asyncapi.yml#/components/schemas/userCreate --output=./folder
asyncapi generate --model=java --context=(asyncapi get schema userCreate) --output=./folder
```

fmvilas:
Love it so far. Simple and clean. For `asyncapi validate`, I think we could just use the existing [Generator](https://www.github.com/asyncapi/generator) syntax. It would be like the following:

```
Usage: asyncapi generate [options] <asyncapi> <template>

- <asyncapi>: Local path or URL pointing to AsyncAPI specification file
- <template>: Name of the generator template like for example @asyncapi/html-template or https://github.com/asyncapi/html-template

Options:
  -V, --version                  output the version number
  -d, --disable-hook [hooks...]  disable a specific hook type or hooks from given hook type
  --debug                        enable more specific errors in the console
  -i, --install                  installs the template and its dependencies (defaults to false)
  -n, --no-overwrite <glob>      glob or path of the file(s) to skip when regenerating
  -o, --output <outputDir>       directory where to put the generated files (defaults to current directory)
  -p, --param <name=value>       additional param to pass to templates
  --force-write                  force writing of the generated files to given directory even if it is a git repo with unstaged files or not empty dir (defaults to false)
  --watch-template               watches the template directory and the AsyncAPI document, and re-generate the files when changes occur. Ignores the output directory. This flag should be used only for template development.
  -h, --help                     display help for command
```

derberg:
That was fast man, thank you for taking a lead here!

> If the input is a local file, there will be an option to automatictly detect file changes and auto reload the output

you mean like a `--watch` flag that keeps the cli running and watching for changes and reacting?

[...]

PR #3, opened 2021-04-12
setting up commands

Souvikns:
---
> Still not complete figuring out all the commands
---

**Description**

- Just installed commander and had set it up

**Related issue(s)**
#1

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,2 @@
+node_modules
+package-lock.json
--- a/bin/asyncapi.js
+++ b/bin/asyncapi.js
@@ -25,12 +25,18 @@ program
     .description("validate the spec file")
     .option('-v, --version <asyncapi-version>', "version of asyncapi spec")
     .option('-c, --context <spec>', "path to the spec file")
+    .action((command) => {
+
+    });
 
 
 program
     .command("create")
     .description("create spec file")
     .option('-c, --context <spec>', "path to the spec file")
+    .action((command) => {
+
+    });
 
 
 program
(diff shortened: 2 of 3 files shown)


GouravSardana:
@jonaslagoni can we add codacy checks here for code quality ?
+ Add sonar cloud bot too. It will be easy for us to maintain the code quality from starting ;)

jonaslagoni:
> @jonaslagoni can we add codacy checks here for code quality ?
> 
>     * Add sonar cloud bot too. It will be easy for us to maintain the code quality from starting ;)

We can yes, I would suggest you add a linter as well from the start - similar to https://github.com/asyncapi/parser-js/blob/2b1d06fef5600caf4893638c58b568fe8d71ced4/package.json#L19, we learned it the hard way and had to refactor a bunch 😄. I will add a PR for the GitHub actions.

derberg:
yes, eslint for static code analysis, 
sonar cloud enabled just now

derberg:
@Souvikns we do not want to use commander but https://github.com/vadimdemedes/ink, we use commander in current generator CLI, but want better user experience that Ink can bring. @jotamusik is working on introducing it.

@jotamusik looking at the interest from the community, maybe you could already open a PR, even if not yet ready, folks that want to help could already jump in and provide feedback, what do you think?

Souvikns:
@derberg ok so let me read [ink](https://github.com/vadimdemedes/ink) and then can update the pr.

jotamusik:
>@jotamusik looking at the interest from the community, maybe you could already open a PR, even if not yet ready, folks that want to help could already jump in and provide feedback, what do you think?
Okey! I will do it this afternoon! :P

PR #4, opened 2021-04-13
chore: add community files

jonaslagoni:
**Description**
This PR adds all the files from https://github.com/asyncapi/.github/
Requested in https://github.com/asyncapi/cli/pull/3#issuecomment-818576916
@derberg I guess you have to set up some extra things afterward?


derberg:
no need to move templates, they are supported by `.github` functionality, meaning they can be stored only in `.github` repo and will work in all repos that do not have own templates, just have a look https://github.com/asyncapi/cli/issues/new/choose

workflows just need to be copied, they would be already but I have one bug that I need to solve in my global workflows action

jonaslagoni:
> no need to move templates, they are supported by `.github` functionality, meaning they can be stored only in `.github` repo and will work in all repos that do not have own templates, just have a look https://github.com/asyncapi/cli/issues/new/choose

Removed.

> workflows just need to be copied, they would be already but I have one bug that I need to solve in my global workflows action

So you have a blocker which prevents us from adding the workflow files as they are now? 🤔

derberg:
@jonaslagoni no, I just mean that now you create those files manually, but since we created CLI repo, there were some updates in global workflows that were not pushed here due to my bug. 

Normally, you would have to copy those files anyway if you would like to have them right after creating the repo (like in robot)

you can merge

derberg (approved):

PR #14, opened 2021-06-18
feat: add new context command to make it easier to work with multiple files

### LINKED ISSUE #8: Persisting context for multiple projects
> context should be optional, and I still should be able to pass context with a flag in case I'm doing some automation, or working with multiple files.

> context should also be smart in the way that if I do not have context set, and I do not pass --context command, CLI check if there are files called asyncapi.json/yml/yaml in workdir, and picks them up.

> context should most probably be persisted in a file.

I think we should store the context as a key:value pair of the directory the command was invoked to the context path 

```
asyncapi context /path/to/spec/file
```

this command will save the {process.cwd(): "/path/to/spec/file"}, this way we can persist
the context for that specific folder/workspaces.

Souvikns:
**Description**
- Create commands to interact with the CLI for context operations
- Successfully create and maintain a `.asyncapi` file at the home directory to persist the context. 
- Create functions or layers to support other commands to use the saved context for their operation. 

**Related issue(s)**

Fixes #8


Souvikns:
@derberg it's not complete, can you check if it is going in the right direction.

src/cli.ts
```diff
@@ -14,6 +14,11 @@ const cli = meow(`
 			Options
 				-c --context  Your AsyncAPI specification
 				-w --watch Enable watchMode (not implemented yet)
+		context
+			current                show the current set context
+			list     			   list all the contexts
+			remove  			   remove the current set context 
+			use <fileName|URL>     set/update context
```
derberg, line 21: looking at our discussion in the issue, we still need `context set <context-name> <filePath | URL>` right? and the `use` should rather be `use <context-name>`?

src/cli.ts
```diff
@@ -22,7 +27,7 @@ const cli = meow(`
 		context: {
 			alias: 'c',
 			type: 'string',
-			isRequired: true
+			isRequired: false
```
derberg, line 30: context needs rename to `file`, as an alternative to the context you set with `context` command, right?

src/cli.ts
```diff
@@ -14,6 +14,11 @@ const cli = meow(`
 			Options
 				-c --context  Your AsyncAPI specification
 				-w --watch Enable watchMode (not implemented yet)
+		context
+			current                show the current set context
+			list     			   list all the contexts
+			remove  			   remove the current set context
```
derberg, line 20: how about making it optional that you can also remove specific context by providing its name?

src/hooks/context/contextService.spec.ts
```diff
@@ -0,0 +1,32 @@
+import { ContextService } from './ContextService';
+import * as os from 'os';
+import * as path from 'path';
+import * as fs from 'fs';
+import { ContextFile } from './models';
+
+describe('Context Service should', () => {
+	beforeEach(() => {
+		const contextFilePath = path.resolve(os.homedir(), '.asyncapi');
```
derberg, line 9: `.asyncapi` string must be in some separate file, so we specify it just once and can reuse in main code and tests. Something like https://github.com/asyncapi/asyncapi-react/blob/master/library/src/constants.ts. Makes sense?

actually `path.resolve(os.homedir(), '.asyncapi');` could also be wrappend in a reusable util

derberg (commented):
looking awesome, so clean! just left some comments.

@jotamusik man, you did great job with initial setup

src/cli.ts
Souvikns: I was thinking that CLI will load the context file, but `--context` could override the existing context for that command execution.

derberg: ok, this makes sense

adding `--file` can be done separately

> I was thinking that CLI will load the context file

it would also be awesome that in case context is not there, the CLI also check if in the execution path there are files like `asyncapi.yml`, `asyncapi.yaml` or `asyncapi.json` and load them automatically.

Souvikns: ok, so you mean, in the working directory. This would be awesome. 

So 
- First check if there is any `asyncapi.yml`, `asyncapi.yaml`, `asyncapi.json` in the working directory 
- If not then check globally for the `.asyncapi` file

derberg: I'm not really sure about the priority here, no strong opinion

> First check if there is any asyncapi.yml, asyncapi.yaml, asyncapi.json in the working directory

this could be first to promote proper filename pattern, but we should also clearly log information about what file is used by the CLI

> If not then check globally for the .asyncapi file

this could also be first as it could be confusing for a user that he set a context but a different file is read.

better DX would be I think if we follow your priority order

Souvikns:
I think we should persist in the context not globally in the `home` directory but rather project level.

### Scenario 
Suppose I have multiple separate projects in my system, and I have installed the CLI as a dev dependency on all my projects. So I am using individual CLI instances for my different projects. They all might have different versions. Saving the context `.asyncapi` file in the home directory, now all the CLI instances need to access that one file. I think the user will need to save the update the file frequently if changing projects often. But if we save the `.asyncapi` file in the project level that is in the `node_module/@asyncapi/CLI/.asyncapi` then the users don't have to update even if changes project. 

what do you think @derberg

derberg:
@Souvikns this must be validated with tools like `brew` or `chocolatey`. I don't know if `node_module/@asyncapi/CLI/.asyncapi` could be supported if you provide a distro for `brew` for example.

Thing is that CLI users are not JS only, and there are many non-JS devs that consider it  a shame to install nodejs 🤷🏼 I'm not gonna judge 😄 we need to respect that and to get wider CLI adoption, try making in available without forcing people to use `npm` or `npx` or anything like that. Thus the `home` location instead of `node_modules`, because I assume later is not supported with my scenario.

src/hooks/context/models.ts
```diff
@@ -0,0 +1,68 @@
...
+export interface Contexts {
+	[name: string]: string
+}
+
+export interface ContextContent {
+	_current: string,
+	_contexts: Contexts
+}
+
+export class ContextFile {
```
jotamusik, line 16: Maybe I would rename this class as StoragedContexts because reading it right now, for me it's like the file that represents a single context. For example, context X that represents the file path/asyncapi.yml 🤷🏻‍♂️

src/hooks/context/hook.tsx
```diff
@@ -0,0 +1,15 @@
+import { container } from 'tsyringe';
+
+import { ContextService } from './contextService';
+import { ContextFile } from './models';
+
+export function useContextFile() {
+	const contextService: ContextService = container.resolve(ContextService);
+
+	return {
+		list: () => {
```
jotamusik, line 10: I would segregate this, make a list method that transform the storaged data to make it ready to use on the ListContextComponent and one method that transform the necessary data to make it ready to use on the ShowCurrentContextComponent 🤔

src/CommandsRouter.spec.tsx
```diff
@@ -23,7 +23,7 @@ describe('CommandsRouter should', () => {
 
     setTimeout(() => {
       expect(lastFrame()).toBe(chalk.green(`File: ${file.getSpecificationName()} successfully validated!`));
-      done();
     }, 200);
+    done();
```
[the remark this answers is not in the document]
Souvikns: I am using windows and it was failing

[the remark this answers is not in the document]
jotamusik: did you try to make the timeout bigger? 🤔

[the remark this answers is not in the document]
Souvikns: yeah, it was still failing.

src/hooks/context/models.ts
Souvikns: today I changed most of the service files, will keep in mind.

src/components/Context/Context.tsx
```diff
@@ -0,0 +1,31 @@
...
+		case 'remove':
+			return <Text>Removing the current set context</Text>
+		case 'use':
+			return <Text>Using a context</Text>
+		case 'create':
+			return <Text></Text>
+		default:
+			return <Text>Unsupported command</Text>
+	}
+}
```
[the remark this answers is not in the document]
Souvikns: Will be writing subcommand router for this 👍🏻

Souvikns:
@jotamusik I tried something with the command router can you check once if it's fine.

Souvikns:
@derberg I have changed the architecture of the hook 

now every function will return a response object and an error object accordingly if there won't be any error then the error object will be undefined and if there will be an error the response object will be undefined. 
```tsx
let {list, err} = useContexFilet().list();

let {specFile, err} = useContextFile().loadSpecFile();

let {contextFile, err} = useContextFile().addContext(key, value);

let {contextFile, err} = useContextFile().changeCurrent("home");

let {contextFile, err} = useContextFile().remove();

```

derberg:
Hey, sorry for the late review but was super busy with 2.1 spec release.
I have some comments:

- If I have no context set, I should get a clear info about it when doing `asyncapi context current`. Now there is no message, just silent execution. I should also get a suggestion how to set a context
- When I do `asyncapi context list` and do not have context file, I should get a nice message that I have no context, user should not care if context are in some file or no, this is `Context file not found in the home directory` technical information that is not relevant, more relevant is that I have no context and how to add one
- `asyncapi context remove` exacutes silently if I have no context. I should get a clear message + we should support removing of another context from the list, not just the current. 
- we have option `set` but in code we always call it `add`, rename to `add` is better imho
- I don't think set works as expected:
    ```
    $ asyncapi context set test https://raw.githubusercontent.com/asyncapi/generator/master/test/docs/dummy.yml
    $ asyncapi context list
    Context file not found in the home directory
    $ asyncapi context current
    $ 
    ```

I believe you are missing ContextService and context related components are not really used

Souvikns:
@derberg Context Releated component I am yet to hook up with the router was am waiting to hear from @jotamusik as he suggested changes for the router logic to be shifted from the context components itself. 
In the meantime, I am fixing all other issues and creating better messages.

derberg:
@Souvikns sound good, ping me when you need me

Souvikns:
@derberg I have moved to ContextService, having doubt about the naming convention for the removal of the context and the current. 
- removing the current context 
- removing an existing context key-value pair.

derberg:
@Souvikns I think I need more data here, what exactly you have a challenge with?

Souvikns:
I just need some opinion from you about the command naming with deletion of an context and clearing the currnent

derberg:
I personally prefer `remove` as this is the word that plays well with `list`, and you remove item from the list. Delete sounds like you delete all, just my non-native speaker opinion.

it is fine to do this imho:
```
context remove <context-name|current>
```

Souvikns:
The tests are passing on my computer. 
<details>
<Summary>Screenshot</Summary>

image

</details>

derberg:
@Souvikns tests are failing but the problem is not tests but building project, it doesn't build on my laptop too, same reason. I' on mac

derberg:
I think it is about case sensitiveness, you need to rename all `from './context'` to `from './Context'`

I suspect [this](https://github.com/sindresorhus/tsconfig/blob/main/tsconfig.json#L31) is the reason

Souvikns:
@derberg I have removed the logic from the `model.ts` file to `contextService.ts` so `model.spec.ts` is kind of redundant I only kept it for review. Shall I remove the tests and the code in the `model.ts` file.

derberg:
if there are duplicates or stuff that is not for the merge then yes, please cleanup. 

I'm not looking into code anyway, doing manual tests to see if flows are right

derberg:
This is current help:
```
        context
                current  show the current set context
                list  list all the contexts
                remove  remove the current set context 
                set <context-name> <filePath | URL>  Set/update contexts
                use <fileName|URL>  set/update context 
```

### `current`

```
$ asyncapi context current

home : home/asynapi.yml
```

Why am I getting a context that points to home and to non existing file. This causes an error later

```
$ asyncapi validate

  ERROR  The "path" argument must be of type string. Received undefined

 node:internal/errors:363:5

 - NodeError (node:internal/errors:363:5)
 - validateString (node:internal/validators:119:11)
 - Object.resolve (node:path:1098:7)
 - SpecificationFile (dist/hooks/validation/models.js:34:26)
 - Validate (dist/components/Validate/Validate.js:27:15)
 - renderWithHooks (node_modules/react-reconciler/cjs/react-reconciler.development.js:6036:18)
 - mountIndeterminateComponent (node_modules/react-reconciler/cjs/react-reconciler.development.js:8570:13)
 - beginWork$1 (node_modules/react-reconciler/cjs/react-reconciler.development.js:9938:16)
 - Object.invokeGuardedCallbackImpl (node_modules/react-reconciler/cjs/react-reconciler.development.js:11563:10)
 - invokeGuardedCallback (node_modules/react-reconciler/cjs/react-reconciler.development.js:11740:31)
```

`asyncapi validate` should work if context is set. This `asyncapi validate --context home` should also work, `context` option should support context name now and not path to file, and in next PR we need to provide support for `--file`

There should be no default context/ We discussed that when I do `asyncapi validate` with no context, then CLI tries to pick up `asyncapi.yml|yaml|json` from current workdir.

### `set`

When I do `asyncapi context set dummy https://raw.githubusercontent.com/asyncapi/generator/master/test/docs/dummy.yml` there is no error or confirmation and after doing `asyncapi context list` I don't see dummy context there

When I do `asyncapi context set dummy /path/sources/generator/test/docs/dummy.yml` it is the same as above.

### Bug not related to this PR

in `models.ts` we have bug here because `e.location` is not `null` but `undefined`:
```
const errorHasLocation = e.location !== null;
        if (errorHasLocation) {
          return errorsInfo.push(`${e.title} ${e.location.startLine}:${e.location.startColumn}`);
        }
```

please change to `const errorHasLocation = !!e.location;`

### `use`

help says `use <fileName|URL>  set/update context `

but `use` should allow `<context-name>` only, and it is for setting context only, not updating, just write `set current context`

Souvikns:
@derberg for the validate command to work with the new context commands I have to make changes to the validate component, should we add those changes in this PR or have a separate PR to do that. 

I am asking this because of https://github.com/asyncapi/cli/pull/14#discussion_r656939103 where I feel this could be a separate module.

derberg:
@Souvikns yes and no 😄 so `--file` can be left for later but if we do not support `--context <context-name>` instead of `<filePath>` then it might feel weird to introduce a feature that contradicts names of options, you know what I mean?

Souvikns:
@derberg I have made the changes you pointed out. 🤞🏻

derberg:
- the bug I mentioned in my previous comment is still there, please have a look, the one about `e.location`
- I don't think you implemented fetching if the spec if someone points to a URL when setting context
- validating against current context doesn't work
  ```
  $ asyncapi context use dummy
  dummy : /Users/wookiee/sources/go-template/test/asyncapi.yaml
  $ asyncapi context current
  dummy : /Users/wookiee/sources/go-template/test/asyncapi.yaml
  $ asyncapi validate
  This context key does not exist.
  ```
- please remove these as they are only confusing during review, like any test code that is not part of tests:
  ```
  home : /home/projects/asyncapi.yml
  code : /home/projects/asyncapi.yaml
  ```
- context removal didn't work for me
- how about picking up asyncapi.yml|yaml|json from current working dir functionality? do you consider it in scope here?



`asyncapi validate --context dummy` - worked, so that's is a good start

please also manually test errors I reported above before you give it for review 😉

derberg:
ignore my comment about `home` and `code`, I think they probably stayed there from previous version of cli, I just had them in `.asyncapi` I guess. When I removed and recreated the file they are not added to the file anymore 👍🏼

Souvikns:
> how about picking up asyncapi.yml|yaml|json from current working dir functionality? do you consider it in scope here?

I have created the logic for autodetecting, but I feel It should have a separate module because it will be used by every other command.

derberg:
> but I feel It should have a separate module because it will be used by every other command.

yes, it definitely will, it is like a global default behaviour

Souvikns:
As we had a conversation about the auto-detection of spec file in the working directory or fallback to saved context. 
I was thinking of a hook that takes the `file and context ` parameter that we pass in the flag and loads the spec path from there, but if both are undefined then it looks for spec file in the working directory, and if that is not found then load current context and if that is not found then throw an error. 
So basically all other commands just need to call this hook to get their spec file path. That's why I was thinking of having a separate hook for this, rather than the `useContextFile` hook.

derberg:
yup, and doesn't matter what is the order as long as it is documented:
1. `--file`
2. `--context`
3. current context
4. working dir

this sounds like the best we could get

Souvikns:
So @derberg what do you think should we implement this feature in another module or this one.

derberg:
because of it's generic nature feels like it should be separate

derberg:
ok then, autodetection of files or context will be done in a separate PR. I see but fixes are solved. The URL fetching I would drop for now

README.md
```diff
@@ -23,9 +23,17 @@ Usage
...
+  context
+    current show the current set context
+    list show the list of all stored contexts
+    remove <context-name> remove a context from the store
+    use <context-name> set any context from store as current
+    add <context-name> <filepath> add/update new context
 
 Examples
-  $ asyncapi validate --context=specification.yml
+  $ asyncapi context add dummy ./specification.yml
```
derberg, line 37: let us always promote naming with `asyncapi`
```suggestion
  $ asyncapi context add dummy ./asyncapi.yml
```

also please rename example spec in this repo

src/CliModels.ts
```diff
@@ -26,14 +27,20 @@ export class CliInput {
...
-    const help = meowOutput.help;
+    const [command, ...args] = meowOutput.input;
     const { context, watch } = meowOutput.flags;
-    return new CliInput(command || 'help', { context, watch }, help);
+    return new CliInput(command || 'help', { context, watch }, args);
+  }
+
+  static createSubCommand(cliInput: CliInput): CliInput {
+    const [command, ...args] = cliInput.arguments;
+    return new CliInput(command || "help", cliInput.options, args);
```
derberg, line 43: there is no such thing as `help` command, right? so even if no command is passed, nothing will really happen, help will not show up really

Souvikns: Yeah, but as I was reusing the `CliInput` class and it needs a string that's why passed the help command, it actually does nothing as help text si handled by `meow`

derberg: feel wierd, let us leave it as it is, just use `'` instead of `"` and please create a followup issue to clear that out, maybe we actually should have `help` as command and not flag 🤔

src/cli.ts
```diff
@@ -12,17 +12,24 @@ const cli = meow(`
...
-	  $ asyncapi validate --context=specification.yml
+	  $ asyncapi context add dummy ./specification.yml
+	  $ asyncapi validate --context=dummy
 `, {
 	flags: {
 		context: {
 			alias: 'c',
 			type: 'string',
-			isRequired: true
+			isRequired: false
```
derberg, line 32: for now, because we do not automatically read current context, this must be set to true because it is always needed

src/components/Context/context.spec.tsx
```diff
@@ -0,0 +1,60 @@
...
+import { ListContexts, ShowCurrentContext, AddContext, SetCurrent } from './Context';
+import { TestingVariables } from '../../constants';
+
+let testing = new TestingVariables();
+
+describe('listing contexts', () => {
+	test("should render error when no context file found", () => {
+		testing.deleteDummyContextFile();
+		let { lastFrame } = render(<ListContexts />);
+		expect(lastFrame()).toMatch("No contexts saved yet.");
```
derberg, line 12: please create another followup issue to turn these messages into constants so we reuse them in components and tests, something like https://github.com/asyncapi/asyncapi-react/blob/master/library/src/constants.ts

no need to fix it now, not related to this pr

src/constants.ts
```diff
@@ -0,0 +1,38 @@
+import * as os from 'os';
+import * as path from 'path';
+import * as fs from 'fs';
+import { SpecificationFile } from './hooks/validation';
+import { Context } from './hooks/context/models';
+
+export const CONTEXTFILE_PATH = path.resolve(os.homedir(), '.asyncapi');
+
+export class TestingVariables {
```
derberg, line 9: it cannot work the way it works now. You are creating real context for the sake of the integration tests I presume, but you do it on the original `.asyncapi` file, so basically me as a developer I have bad user experience because what I have in my context is lost because of these tests.

maybe approach with `.asyncapi` should be configurable so in tests you can say the context is in some temporary folder that is in gitignore. Thoughts?

Souvikns: The context service loads from the `.asyncapi` file that is in the home directory, and when I was developing, It was kind of annoying because after the tests run I always had a new .asyncapi file. If our service takes the global context file's path as an argument I can create other dummy files for testing.

src/hooks/context/contextService.ts
```diff
@@ -0,0 +1,46 @@
...
+		if (!fs.existsSync(CONTEXTFILE_PATH)) throw new ContextFileNotFoundError();
+		return JSON.parse(fs.readFileSync(CONTEXTFILE_PATH, 'utf-8')) as Context;
+	}
+
+	deleteContextFile() {
+		if(fs.existsSync(CONTEXTFILE_PATH)) fs.unlinkSync(CONTEXTFILE_PATH);
+	}
+
+	save(context: Context) {
+		fs.writeFileSync(CONTEXTFILE_PATH, JSON.stringify(context), { encoding: 'utf-8' });
```
derberg, line 20: we should also validate if file exists when saving the context, otherwise it doesn't make sense to save it

Souvikns: This I had a long thought when I started with this, but then I came to the conclusion that context will only persist path, and we have other features to check if the file exists or not because validate hook is basically doing that.

src/constants.ts
derberg: exactly, now I know where I was getting this home and code contexts from 😄 

yeah, I think we need a way to override `home/.asyncapi` for tests purposes otherwise we will get frustrated quite often

src/hooks/context/contextService.ts
derberg: yeah, but the problem is that if I made a mistake in path when adding it, I learn about it when validating, so too late, I should learn about it the moment I create it

Souvikns: Yeah, so I should have a check that if that file exists.

derberg (changes requested):
- manual tests 🚀 
- there are some issues that are not much related to this PR or can be done later, I wrote about followup issues, please create them 💪🏼 
- what we definitely need to fix before merge is
   - making sure we check file exists before we add it to the context
   - make sure tests do not messup with global context

so far so good!

src/constants.ts
jotamusik, line 9: For me TestingVariables is not such a good name. I will name it something like `ContextFixtures`

Fixtures is that kind of background data that you need on a test in order to make it works. That's why I think that this name could be better 😅

src/hooks/context/hooks.tsx
```diff
@@ -0,0 +1,103 @@
...
+
+
+	return {
+		list: () => {
+			try {
+				let ctx: Context = contextService.loadContextFile();
+				let response = Object.keys(ctx.store).map(c => ({ key: c, path: ctx.store[c] }));
+				return { response };
+			} catch (error) {
+				return { undefined, error };
```
jotamusik, line 17: I will type this kind of returned object in order to give some behaviour like the ValidationResponses

jotamusik (changes requested):

jotamusik:
Also, be aware that the default tab size is 2 spaces and there are some new files that have a tab size of 4 instead. 😅

src/constants.ts
derberg: just have test in name so it is clear what that's for

derberg:
@jotamusik 

> Also, be aware that the default tab size is 2 spaces and there are some new files that have a tab size of 4 instead. 😅

you sure you wanna do a tab-fight on Friday 😆

src/components/Context/Context.tsx
```diff
@@ -0,0 +1,87 @@
...
+
+	return <Text>{response}</Text>
+}
+
+export const SetCurrent: FunctionComponent<{ options: any, args: string[] }> = ({ args }) => {
+	let [key,] = args;
+
+	if (!key) {
+		return <ContextError error={new Error("missing arguments")} />
+	}
```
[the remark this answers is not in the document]
Souvikns: My hook accepts `string` and that's why I have checked this here, will it be better to have my hook accept `undefined` also than check there.

derberg (approved):
LGTM 🚀

derberg:
@Souvikns please do not forget to create all the follow-up issues that were mentioned here in the comments and also on slack. Remember that first improvement that we need to do here is to make sure tests do not operate on global `.asyncapi` context. It should be possible to manipulate it and I think it will be super easy just to address it through environment variable. So basically `.asyncapi` at home dir is default but one can override it with env like `CLI_HOME_PATH=./path`, so I run it as `CLI_HOME_PATH=./path asyncapi context current`. This should do the trick

Issue #35, opened 2021-07-26
Validation of default files in CWD does not work when I have no files in the context


derberg:
This is how you can reproduce it:
```
$ asyncapi context list
myapp1 : /Users/wookiee/sources/cli/test/spec.yaml
$ asyncapi context current
No context is set as current, please set a current context.
$ asyncapi validate
File: /Users/wookiee/sources/cli/asyncapi.yaml successfully validated!
$ asyncapi context remove myapp1
context deleted successfully
$ asyncapi context list
No contexts saved yet.
$ asyncapi validate
No contexts saved yet, run asyncapi --help to know more.
```

in case I do `asyncapi validate` and, flags like `--context` or `--files` are not passed, and there is no default context + no contexts at all, the validation of default files from CWD should work

Souvikns:
@derberg shall I start with this issue next ?

derberg:
yeap

Souvikns:
## Error 
When the global context file is absent, so no default context is present, and also when flags are missing but there is an `asyncapi.yml` file in the working directory, It should load that file automatically, but it is throwing `No contexts saved yet, run asyncapi --help to know more.` error. 

## Cause of Error 
This is happening because of https://github.com/asyncapi/cli/blob/b459c1c79f27253524def11576298e0aee45831d/src/hooks/context/hooks.tsx#L136 loading the context file before checking for auto-detection. It is needed because to read the `--context` flag we need the global context config. 

## Two Possible Solution 

1. If we give auto-detection first priority then in this scenario the CLI will auto-detect files first and not load the global config. But doing this will change the default behavior as it will not look for global current context first, but only if there is no `asyncapi.yml` file in the working directory. This could be beneficial in some scenarios. 

2. After we catch the `ContextFileNotFoundError` we check for the `asyncapi.yml` file in the working directory and throw an error if it is not found. This way the default behavior will not change but fix the issue with this scenario. 

> @derberg what do you think, which approach should I follow. 
---
 
https://github.com/asyncapi/cli/blob/b459c1c79f27253524def11576298e0aee45831d/src/hooks/context/hooks.tsx#L126-L162

derberg:
I'm definitely with 2, just make sure message thrown to the user clearly explains cause and next steps

PR #40, opened 2021-07-27
fix: improve CLI messages and move them into separate file

### LINKED ISSUE #21: Turn CLI messages to constants
please create another followup issue to turn these messages into constants so we reuse them in components and tests, something like https://github.com/asyncapi/asyncapi-react/blob/master/library/src/constants.ts

no need to fix it now, not related to this pr

_Originally posted by @derberg in https://github.com/asyncapi/cli/pull/14#discussion_r666844302_

Souvikns:
**Description**
Turning CLI messages into constant so that it would be easy to update in the future. 

**Related issue(s)**
Resolves #21


Souvikns:
## Breaking the help commands into different parts. 
Taking inspiration from `commander.js` we can divide every command to have top options and its subcommands. 
so the top-level command would look something like this. 

#### `asyncapi --help`
```
Usage: asyncapi [options] [command]

Options:
  -V, --version       output the version number
  -h, --help          display help for command

Commands:
  context [options]   Manage contexts
  validate [options]  validate spec file
```

#### `asyncapi context --help` 
```
Usage: asyncapi context [options] [command]

Store key-value pair of contexts and the CLI will load your contexts 
automatically making your commands shorter. 

Options:
  -h, --help          display help for command

Commands:
  list                list all saved contexts
  current             see current context
  use <context-name>  set the context as current context
  add <context-name> <spec-file-path>   add/update new context
  remove <context-name>  remove a context
```

#### `asyncapi validate --help`
```
Usage: asyncapi validate [options]

validate spec file

Options:
  -f, --file <spec-file-path>         Path of the spec file
  -c, --context <saved-context-name>  context name to use
  -w, --watch                         watch mode
  -h, --help                          display help for command

Example:
 $ asyncapi validate --file=./asyncapi.yml
 $ asyncapi validate  // if context is set. 
```

derberg:
looks great, but remember, not in scope of this PR 😉 

> Store key-value pair of contexts and the CLI will load your contexts 
automatically making your commands shorter. 

we need also a sentence that explains what context is

> validate spec file

Rather `Validate an AsyncAPI file`

> Also I was thinking we should add an external link to the documentation of how CLI loads spec path automatically

I don't think it should be external. when I run `asyncapi validate` I don't want to see error that I have no contexts, I want to just get an info that I need to point to AsyncAPI file and a detailed explanation how I can do that, right there in the error, that I can use file flag with link to file, or context name with info how to get more help about context, or default context or default filename in CWD

Souvikns:
@derberg I have converted all the error messages into constants and created a `Help` class to generate help messages as needed.

src/messages.tsx
```diff
@@ -0,0 +1,78 @@
...
+      '-f, --file <spec-file-path>  Path of the spec file',
+      '-c, --context <saved-context-name>  Context name to use',
+      '-w, --watch  watch mode',
+      this._helpFlag
+    ]
+  }
+
+  private context: CommandHelp = {
+    name: 'context',
+    usage: 'asyncapi context [options] [command]',
```
derberg, line 47:
```suggestion
    usage: 'asyncapi context [command] [options]',
```

src/messages.tsx
```diff
@@ -0,0 +1,78 @@
...
+    shortDescription: 'Manage contexts',
+    flags: [
```
derberg, line 49:
```suggestion
    longDescription: 'Context is what makes it easier for you to work with multiple AsyncAPI files. You can add multiple different files to a context. This way you do not have to pass --file flag with path to the file every time but just --context flag with reference name. You can also set a default context, so neither --file nor --context flags are needed.
    flags: [
```

+ enable support for `longDescription` as atm it is not consumed

derberg (commented):
@Souvikns nice, looks super clean, just left 2 comments

src/help-message.ts
```diff
@@ -0,0 +1,100 @@
...
+      }
+    }
+
+    return helpText;
+  }
+
+  showCommandHelp(command: CommandName) {
+    let helpText = '';
+    // eslint-disable-next-line security/detect-object-injection
+    const commandHelpObject = this.helpMessage.commands[command];
```
magicmatatjahu, line 78:
```suggestion
    const commandHelpObject = this.helpMessage.commands[String(command)];
```

derberg (commented):
what about messages from https://github.com/asyncapi/cli/blob/master/src/components/Context/contexterror.tsx ? 

ultimate goal should be you never have to do things like this in tests: `expect(lastFrame()).toMatch('No contexts saved yet.');` so basically never explicitly hardcode the message in the test but use a constant

src/help-message.ts
Souvikns: `const commandHelpObject = this.helpMessage.commands[String(command)];` wasn't working but 
`const commandHelpObject = this.helpMessage.commands[command as CommandName];`  did.

magicmatatjahu: If you can remove the `eslint-disable...` by doing it, then change it with `as`

Souvikns:
@derberg I have converted the error messages into constants. 

Some context commands have string responses like when a context is deleted. Shall I convert those into constant?

derberg:
@Souvikns basically any message printed to the user must be a constant as it must be tested, thus it is used in code and tests, and should not be repeated. There are some tests for context and validation that are still using "hardcoded" messages, like `Error Thrown`, `Error in validation`, `context deleted successfully` or `File: ${file.getSpecificationName()} successfully validated!`

Souvikns:
I was thinking of creating a complete class for responses, just like I did for the help message text.

derberg:
For HelpMessage it makes sense as there is a lot of reusability, the other messages are pretty independent. Class would be an overhead imho

Souvikns:
@derberg can we run the tests again, all the test cases are running in my local setup. 
image

derberg:
@Souvikns look at the error:
```
Expected substring: "No contexts saved yet, run asyncapi --help to know more"
    Received string:    "home : /Users/runner/work/cli/cli/test/specification.yml
    code : /Users/runner/work/cli/cli/test/specification.yml"
```

looks like test picked up some old instance that has `.asyncapi` file with some old test data. I rerun, but might be it won't work.

derberg:
Still some messages are there hardcoded:
- https://github.com/Souvikns/cli/blob/better-cli-messages/src/hooks/validation/hook.tsx#L13
- https://github.com/Souvikns/cli/blob/better-cli-messages/src/hooks/validation/hook.tsx#L17

Souvikns:
Again the macos-tests are failing. 

```tsx
  test('return validation service errors when the validation has failed', async () => {
    ValidationServiceMock().makeReturn(ValidationResponse.createError({ detail: 'Error in validation' }));

    const useValidateResponse = await useValidate().validate(fileThatExistsValidationInput);
    expect(useValidateResponse.success).toBeFalsy();
    expect(useValidateResponse.errors[0]).toEqual('Error in validation');
  });
```

There are some test cases that are checking for error messages from the parser. Should we need to make these errors constants.

src/messages.tsx
```diff
@@ -0,0 +1,20 @@
+export const NO_CONTEXTS_SAVED = 'No contexts saved yet, run asyncapi --help to know more';
+
+export const CONTEXT_NOT_FOUND = 'This context key does not exists.';
+
+export const KEY_NOT_FOUND = 'Key not found';
+
+export const DELETING_CURRENT_CONTEXT = 'You are tyring to delete a context that is currently in use.';
```
derberg, line 7: spelling error and also context name should be in the message

`Context {CONTEXT_NAME} is set as current and cannot be removed`

src/messages.tsx
```diff
@@ -0,0 +1,20 @@
...
+
+export const MISSING_CURRENT_CONTEXT = 'No context is set as current, please set a current context.';
```
derberg, line 9: I don't think `MissingCurrentContextError` is ever used

src/messages.tsx
```diff
@@ -0,0 +1,20 @@
...
+
+export const MISSING_ARGUMENTS = 'Missing arguments.';
```
derberg, line 11: I don't think `MissingArgumentstError` is ever used

src/messages.tsx
```diff
@@ -0,0 +1,20 @@
...
+
+export const NEW_CONTEXT_ADDED = 'New context added.';
```
derberg, line 13: How about?

`New context added. \nYou can set it as your current context:\n      asyncapi context use {CONTEXT_NAME}\nYou can use this context when needed with --context flag`

derberg (commented):
since we have now all the messages in one place I made some suggestions to make them sound better

src/messages.tsx
Souvikns: `MissingCurrentContextError` is used here https://github.com/asyncapi/cli/blob/2f69e066f00c1dcb8336b1f1f2d8aac09d23884b/src/hooks/context/hooks.tsx#L27

derberg: oh, but such message should never be printed to the user. When user did not pass `file` or `context` option, and also did not specify a default context, we should look for default files in CWD, and if they are missing, we should print a nice message that explains possible options

derberg: can be fixed with https://github.com/asyncapi/cli/issues/35 though

Souvikns: Yeah I will be fixing this just after this PR.

src/messages.tsx
Souvikns: @derberg I was making these changes, and I think It would look good if the command example has a different color, but it is hard to make them in different colors in a string. But as we are using ink I think making all the text messages a component would be nice.

derberg: yup, definitely, one of the reasons we use Ink. I guess that you would prefer to do it in a followup PR to not complicate this one?

Souvikns: Ok so in this PR I will just keep it as it is and update the texts. Then in another follow-up pr, we can convert all string constants to components.

src/messages.tsx
Souvikns: This is used when any input variables are missing 

https://github.com/asyncapi/cli/blob/2f69e066f00c1dcb8336b1f1f2d8aac09d23884b/src/hooks/context/models.ts#L50-L55

src/messages.tsx
Souvikns: This one is currently not in use. I created it for a warning case if the user is deleting any context that is set as current but didn't end up using it. Should I remove this?

derberg: yup

src/messages.tsx
derberg: let's leave it as it is for now

derberg:
ready for last review round?

Souvikns:
Two minor things left to update, then It will be ready.

Souvikns:
> ready for last review round?

@derberg I am ready for the last review round. 🤞🏻

derberg (commented):
one more message left

```
export class SpecFileNotFoundError extends Error {
  constructor() {
    super();
    this.message = 'specification file not found in that path.';
  }
}
```

should be `No AsyncAPI file found in {PATH}`

best reuse message from validation, `File: ${filePath} does not exists or is not a file!`

src/messages.tsx
```diff
@@ -0,0 +1,16 @@
+export const NO_CONTEXTS_SAVED = 'No contexts saved yet, run asyncapi --help to learn more';
+
+export const CONTEXT_NOT_FOUND = (contextName: string) => `Context ${contextName} does not exists.`;
+
+export const MISSING_CURRENT_CONTEXT = 'No context is set as current, please set a current context.';
+
+export const MISSING_ARGUMENTS = 'Missing arguments.';
+
+export const NEW_CONTEXT_ADDED = (contextName: string) => `New context added.\nYou can set it as your current context:\n asyncapi context use ${contextName}\nYou can use this context when needed with --context flag`;
```
derberg, line 9:
```suggestion
export const NEW_CONTEXT_ADDED = (contextName: string) => `New context added.\n\nYou can set it as your current context: asyncapi context use ${contextName}\nYou can use this context when needed with --context flag: asyncapi validate --context ${contextName}`;
```

derberg (commented):
left just 2 comments and additional question: meow should be updated and use new help message right?

PR #52, opened 2021-08-18
feat: improving help command

### LINKED ISSUE #22: Improvement with help command
feel wierd, let us leave it as it is, just use `'` instead of `"` and please create a followup issue to clear that out, maybe we actually should have `help` as command and not flag 🤔

_Originally posted by @derberg in https://github.com/asyncapi/cli/pull/14#discussion_r666839222_

Souvikns:
**Description**

- Creating support for root and command level help components. 
- Using ink to create beautiful help components. 

**Related issue(s)**

Fixes #22

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -27,6 +27,7 @@ const cli = meow(`
 		$ asyncapi validate --context=dummy
 		$ asyncapi validate --file=./asyncapi.yml
 `, {
+  autoHelp: false,
   flags: {
     context: {
       alias: 'c',
@@ -43,6 +44,9 @@ const cli = meow(`
       alias: 'f',
       type: 'string',
       isRequired: false
+    },
+    help: {
+      alias: 'h'
     }
   }
 });
--- a/src/help-message.tsx
+++ b/src/help-message.tsx
@@ -1,4 +1,6 @@
 import { injectable, container } from 'tsyringe';
+import React, { FunctionComponent } from 'react';
+import {Text} from 'ink';
 
 export type CommandName = 'validate' | 'context';
 
@@ -52,6 +54,10 @@ export class HelpMessage {
 
 export class HelpMessageBuilder {
   private helpMessage: HelpMessage = container.resolve(HelpMessage);
+
+  HelpComponent: FunctionComponent<any> = () => {
+    return <Text>{this.showHelp()}</Text>;
+  }
 
   showHelp() {
     let helpText = '';
--- a/src/CliModels.ts
+++ b/src/CliModels.ts
@@ -1,4 +1,4 @@
-export type Command = string;
+export type Command = string | undefined;
 export type HelpMessage = string;
 export type Arguments = string[];
 
@@ -12,11 +12,13 @@ export class CliInput {
   private readonly _command: Command
   private readonly _options: Options
   private readonly _arguments: Arguments
+  private readonly _help: boolean | undefined
 
-  private constructor(command: Command, options: Options, args: Arguments) {
+  private constructor(command: Command, options: Options, args: Arguments, help?:boolean) {
     this._command = command;
     this._options = options;
     this._arguments = args;
+    this._help = help;
   }
 
   get command(): Command {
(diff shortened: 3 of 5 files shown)


Souvikns:
## All help messages

### `asyncapi --help`
image

### `asyncapi validate --help`
image

### `asyncapi context --help`
image

Souvikns:
@derberg pr is ready for review.

derberg:
love it mate!

how about in such cases:
```
$ asyncapi validate

- You can use --file flag to provide a path to the AsyncAPI file: asyncapi validate --file path/to/file/asyncapi.yml
- You can use --context flag to provide a name of the context that points to your AsyncAPI file: asyncapi validate 
--context mycontext
- In case you did not specify a context that you want to use, the CLI checks if there is a default context and uses 
it. To set default context run: asyncapi context use mycontext
- In case you did not provide any reference to AsyncAPI file and there is no default context, the CLI detects if in 
your current working directory you have files like asyncapi.json, asyncapi.yaml, asyncapi.yml. Just rename your file
 accordingly.
```

or will it followup as now you just focused on part related to `help`?

src/help-message.tsx
```diff
@@ -0,0 +1,129 @@
...
+
+  readonly commands: Command = {
+    validate: {
+      usage: 'asyncapi validate [options]',
+      shortDescription: 'Validate asyncapi file',
+      flags: [
+        this.helpFlag,
+        '-f, --file <spec-file-path>  Path of the AsyncAPI file',
+        '-c, --context <saved-context-name>  Context to use',
+        '-w, --watch  Watch mode'
```
derberg, line 38:
```suggestion
        '-w, --watch  Enable watch mode (not implemented yet)'
```

src/cli.ts
```diff
@@ -27,6 +27,7 @@ const cli = meow(`
 		$ asyncapi validate --context=dummy
```
derberg, line 27: how is this part working now, when meow is used? it kinda duplicates other messages right?

Souvikns: Since `autoHelp` is false the old string help message does nothing we can remove that.

Souvikns:
Not all terminals have similar support for colors, so in some terminals, it is not looking good and also have visibility issue. Like in CMD - 
image

src/cli.ts
derberg: go ahead then

derberg:
@Souvikns I think it is ok if it is not that good in windows commander, I checked in basic mac terminal and it looks good. Imho it is acceptable

derberg (approved):

src/cli.ts
```diff
@@ -43,6 +23,9 @@ const cli = meow(`
       alias: 'f',
       type: 'string',
       isRequired: false
+    },
+    help: {
+      alias: 'h'
```
derberg, line 28: shouldn't `version` be here too? also I noticed `-v` yields help and not version

derberg:
@Souvikns I approved but there is one more comment?

src/cli.ts
Souvikns: The `--version` works automatically by meow. But now when I tested it, `-v` does not work should I close `autoVersion` and write custom logic for it.

derberg: open a follow up issue for it please and for now remove from help messages references for `-v`

derberg (commented):
@Souvikns what about [this](https://github.com/asyncapi/cli/pull/52#issuecomment-903776427) comment + I see you removed the entire info about the version, but the version works fine, just problem is that only `-v` alias doesn't work out of the box, so please document just `version`

Souvikns:
Then let me update it in the readme and as well as in the cli messages.

Souvikns:
@derberg actually the way I have written the code it splits the flag alias and flag param for highlighting so it is coming something like this, does this look ok?

image

derberg:
@Souvikns it would make sense to highlight `--help` too, so always highlight all flags imho

Souvikns:
ok sure then.

Souvikns:
The initial help message has been changed how shall we update the readme file to support this. Should I put a screenshot or just make some kind of walkthrough in the usage to how to access help messages for different commands.

derberg:
make it easy now, we can beautify readme later, just screen shots of output from each help, or copy-paste of help text

Souvikns:
Then should I update the readme in this PR?

derberg:
Yeah, just simplemupdate is needed of what we have now to reflect reality. Better docs, descriptions and some walkthrough can be done in a followup

Souvikns:
@derberg does this look good. I guess adding an image for the root help we would have to update the image every time we add new commands. 
asyncapi-help

derberg:
@Souvikns I definitely agree that at this stage doing images would be to much. We should just update `CLI` section though, what you see after `asyncapi --help` and what after `asyncapi context --help` and `asyncapi validate --help`

Issue #37, opened 2021-07-26
Always follow the same structure (verb + noun, noun + verb, or namespace + verb + noun)


fmvilas:
I think we should be following a structure, regardless of the command at hand. For instance:

```
asyncapi validate file
```

Is following the `verb + noun` approach. Whereas:

```
asyncapi context add
```

Is reversing it to `noun + verb`.

In #1, I suggested we use `verb + noun` because it reads naturally. However, after thinking about how to flip contexts and how to occasionally evolve them to accept multiple asyncapi files, it seems the `noun + verb` might work better here. Just like the Github CLI. So for instance, if I want to add one more asyncapi file to a context, how would it be?

```
asyncapi add context # that sounds like adding a context not "adding to" a context
asyncapi add-to context # clear but ugly
```

So maybe it's a better idea to go for something like:

```
asyncapi context add # Creates a context
asyncapi context add file # Adds a file to a context
```

In this case, the structure would be something like `noun + verb + [noun]` (being the last noun optional). To avoid confusion, I suggest we call it `namespace + verb + [noun]` because the first noun actually works like a namespace.

derberg:
I suggested `asyncapi context add` and others as it feels very natural when I use the same approach with GitHub CLI. I just think the verbs might be very different per noun, so `verb + noun` would be very strange here.

the problem is that sometimes there is just no `noun`, like `validate` or `generate`, so general rule sound like `noun? + verb`

fmvilas:
Yeah, that's true. Maybe we should still consider the `namespace` and make it something like `[namespace] + verb + [noun]`? Just throwing ideas.

derberg:
I have to say I'm not fully getting the `namespace` idea. What would it be?

fmvilas:
Like in this case:

```
asyncapi context add file
```

It's different than:

```
asyncapi context add
```

The first one adds an AsyncAPI file to the context, the second one adds a context. The first one is, for instance, if we want to have tools like https://github.com/asyncapi/app-relations-discovery in the CLI. In that case, you don't want to manually specify all the files because they could potentially be hundreds. Does it make sense?

derberg:
I case kubectl or github cli, stuff similar to context, go under `config` verb actually so it is either `kubectl config get-context` or `gh config set editor vim`. I think we should follow, as in our case context is pure config related, having it on the "root" may confuse it has anything to do with the AsyncAPI-related functionalities


[...]

PR #59, opened 2021-08-31
feat: change from flags to arguments when passing path to a file or context name

Souvikns:
**Description**
In this PR I am trying to shift from `--file` flag to input so now the validate command will look something like this - `asyncapi validate <file-path|urlcontext-name>`. As of now, we were relying on the `useSpecFile` hook to load the spec file according to availability. It was built in a way that any command in the future could use it load the spec file as all commands depend on the spec file in some way since we were using the `--file` flag. Now when we are shifting from flags to input, this hook is not needed.
Also, other commands will have different needs for accessing inputs. So I think we would be needing to create different hooks to load the spec file according to that specific command. 

Amidst all this, some logic would be reusable like
- If no input is passed and if the command is loading a single spec file then usage of the current context or autodetecting spec file from working directory.
- If the input is passed checking if it is a valid **local spec file path**, **URL**, or **context name**. 

**Related issue(s)**

See also #37


derberg:
@Souvikns fyi GitHub has option to create a `draft` pull request

Souvikns:
@derberg should I make this a draft now?

derberg:
Nah, it's ok, just remember next time, so you do not have to add any disclaimer in the description that PR is wip

Souvikns:
@derberg I am ready for review. 

Working on this PR I think the `ContextService` has a problem, but I don't know if that is a legit problem its something I feel, should I open an issue talking about this.

derberg:
@Souvikns don't forget about docs and help commands 😉 

as for the `ContextService` -> you definitely need to create an issue as I have no idea what are you referring too exactly 😄

derberg:
- when I do below, I don't get an error that the given option is not available and instead my default context is picked.
  ```
  asyncapi validate --context test2
    File: /xxx/test/docs/dummy.yml successfully validated!
   ```
- when I do `asyncapi validate /wrong/path/dummy.yml` I get error with outdated help text
- when I do `asyncapi context add test /correct/path/dummy.yml`I get help message that has outdated help text

search in project for any mention of `--file` and `--context`

README.md
```diff
@@ -40,13 +40,11 @@ context  Manage contexts
 ```
 USAGE
 
-asyncapi validate [options]
+asyncapi validate <specPath | context-name> [options]
```
derberg, line 43:
```suggestion
asyncapi validate <specPath | contextName> [options]
```
or
```suggestion
asyncapi validate <spec-file-path | context-name> [options]
```
for consistency, and make sure it is changed everywhere needed

Souvikns:
When you do `asyncapi validate --context test2` it loads the current context because now cli is loading from input and test2 is in options and the input is undefined which leads it load context from current. 

Now `--file` and `--context` does not work if you want to load any specific context then you can do it like this `asyncapi validate test2`. I think removing every mention of `--file` and `--context` flags would help the users.

derberg:
> When you do asyncapi validate --context test2 it loads the current context because now cli is loading from input and test2 is in options and the input is undefined which leads it load context from current.

yes, but we need to have in place validation of options and notify users if they use not existing options, so we do not do what we think is better for users but tell what's wrong. You know what I mean? can be done as follow-up as this could be supported also by https://www.npmjs.com/package/levenshtein-edit-distance to not only tell users the command is wrong, but also try suggesting the correct one

derberg:
@Souvikns awesome, lemme know once you need me for another review round

Souvikns:
> @Souvikns awesome, lemme know once you need me for another review round

@derberg it is ready to be reviewed.

derberg (approved):
🔥 🚀

You can merge it on your own. Make sure the commit that you will use when using `squash&merge` reflects the one from PR title
:shipit:

Issue #38, opened 2021-07-26
Make contexts shareable


fmvilas:
#### Reason/Context
Currently, contexts are stored in the user's home directory. I think it would be cool to store the contexts in the repo directory so they can be shared.

#### Description
This would work the same way package.json works. The `asyncapi` CLI would seek for the file (let's call it `.asyncapi`) in the CWD and, if not found, it would look for it in the parent directory, then the parent of the parent, etc.

The cool thing is that if you don't want to share it, you could just place the file in your home dir, effectively behaving as it is now (global). Also, the global context would be the fallback if a "local" one is not found in a specific project.

Also, this would remove the need to set up contexts yourself manually. Since the file is already in the repo, running `asyncapi` will already load the context automatically.

Additionally, this eliminates the problem where a user doesn't have permission to write to disk. Very likely to happen in private built in-house CI systems (this was the case at New Relic).

derberg:
I like this idea a lot!

Souvikns:
would like to start by talking about the data structure for context. All context is doing right now is storing file paths that supposedly point to a spec path. Storing it in a key-value pair. So context in general looks something like this.

```ts
interface Context {
  current?: string,
  store: {
    [name: string]: string
  }
}
```

`ContextService` is loading this data and providing functions to read and manipulate this. Now the problem is there is a case that the global `.asyncapi` might be missing. For this case, the loader is throwing a `ContextFileNotFoundError` which serves the purpose, but this means we have to check for this error every time we are using this. This is making the code messy, I would like to give some examples - 

While autoloading the spec file, there was a case where when there was no global `.asyncapi` file it was unable to autodetect spec files in the working directory as it was throwing the `ContextFileNotFoundError`. Had to work around like this. 
https://github.com/asyncapi/cli/blob/87970d12bb9a3dadc392805beb0ab3fd025c9048/src/hooks/context/hooks.tsx#L158-L168

Was playing around with the idea that we don't throw an error rather return an object with undefined context. 

<details>
<summary>Something like this. </summary>

```ts
import * as fs from 'fs';
import * as path from 'path';


const SPEC_FILE_PATH = path.resolve(__dirname, 'asyncapi');

interface Context {
  current?: string,
  store: {
    [name: string]: string
  }
}

const saveContext = (ctx: Context) => {
  fs.writeFileSync(SPEC_FILE_PATH, JSON.stringify(ctx), {encoding: 'utf-8'});
}

class ContextService{
[...]

PR #72, opened 2021-09-20
refactor: updated and decoupled contextService logic

Souvikns:
**Description**
This is the start of the process to [move the context commands under config](https://github.com/asyncapi/cli/issues/37#issuecomment-908137433). In this PR I have focused on updating the `contextService` because of the issue discussed [here](https://github.com/asyncapi/cli/issues/38#issuecomment-912377247). I would like to clarify what I have done here. 

- I have recreated context service but this time decoupling as much as possible. 
- Context does a good job to support the `validate` command as it needs only one spec file path. But to support libraries like [cupid](https://github.com/asyncapi/cupid) and [diff](https://github.com/asyncapi/diff) where more than one spec file path is needed context shall evolve as well. For that, I have created a separate `IContext` interface that could be easily updated. 
- To support the future ambition of loading context files [from the root directory and to be sharable](https://github.com/asyncapi/cli/issues/38#issue-952937304) I have created a separate class `ContextAllocator` to load context files accordingly. For now, it is loading the context file from the root file itself. This also helps in testing allowing us to create mock classes to check functionalities. 
- Now `ContextService` does not throw an error if the contextFile is not present in the root dir instead it loads undefined context. 

**Related issue(s)**

See also #38 and #37

### DIFF
--- a/src/config/context/model.ts
+++ b/src/config/context/model.ts
@@ -1,4 +1,4 @@
-import { injectable, registry, } from 'tsyringe';
+import { injectable } from 'tsyringe';
 import fs from 'fs';
 import { CONTEXTFILE_PATH } from '../../constants';
 
@@ -30,12 +30,6 @@ export interface IContextAllocator {
 }
 
 @injectable()
-@registry([
-  {
-    token: 'IContextAllocator',
-    useToken: ContextAllocator
-  }
-])
 export class ContextAllocator implements IContextAllocator {
   contextFilePath = CONTEXTFILE_PATH;
   load(): Context | undefined {
@@ -48,7 +42,10 @@ export class ContextAllocator implements IContextAllocator {
 
   save(context: Context) {
     try {
-      fs.writeFileSync(this.contextFilePath, JSON.stringify(context), { encoding: 'utf8' });
+      fs.writeFileSync(this.contextFilePath, JSON.stringify({
+        current: context.current,
+        store: context.store
+      }), { encoding: 'utf8' });
       return true;
     } catch (error) {
       return false;
--- a/src/config/context/model.ts
+++ b/src/config/context/model.ts
@@ -1,5 +1,3 @@
-/* eslint-disable no-undef */
-/* eslint-disable no-console */
 import { injectable } from 'tsyringe';
 import fs from 'fs';
 import { CONTEXTFILE_PATH } from '../../constants';
@@ -38,6 +36,7 @@ export class ContextAllocator implements IContextAllocator {
     try {
       return new Context(JSON.parse(fs.readFileSync(this.contextFilePath, 'utf8')) as IContext);
     } catch (error) {
+      console.warn(error);// eslint-disable-line no-undef, no-console
       return undefined;
     }
   }
@@ -50,7 +49,7 @@ export class ContextAllocator implements IContextAllocator {
       }), { encoding: 'utf8' });
       return context;
     } catch (error) {
-      console.warn(error);
+      console.warn(error);// eslint-disable-line no-undef, no-console
       return undefined;
     }
   }
--- a/src/config/context/model.ts
+++ b/src/config/context/model.ts
@@ -1,3 +1,5 @@
+/* eslint-disable no-undef */
+/* eslint-disable no-console */
 import { injectable } from 'tsyringe';
 import fs from 'fs';
 import { CONTEXTFILE_PATH } from '../../constants';
@@ -48,6 +50,7 @@ export class ContextAllocator implements IContextAllocator {
       }), { encoding: 'utf8' });
       return context;
     } catch (error) {
+      console.warn(error);
       return undefined;
     }
   }
--- a/src/config/context/model.ts
+++ b/src/config/context/model.ts
@@ -1,0 +1,58 @@
+import { injectable, registry, } from 'tsyringe';
+import fs from 'fs';
+import { CONTEXTFILE_PATH } from '../../constants';
+
+export interface IContext {
+  readonly current?: string,
+  readonly store: {
+    [name: string]: string
+  }
+}
+
+export class Context implements IContext {
+  current?: string | undefined;
+  store: { [name: string]: string; };
+
+  constructor(ctx: IContext) {
+    this.current = ctx.current;
+    this.store = ctx.store;
+  }
+
+  getContext(contextName: string) {
+    return this.store[contextName as string];
+  }
+}
+
+export interface IContextAllocator {
+  contextFilePath?: string
+  load: () => Context | undefined
+  save: (context: Context) => boolean
+}
+
+@injectable()
+@registry([
+  {
+    token: 'IContextAllocator',
+    useToken: ContextAllocator
+  }
+])
+export class ContextAllocator implements IContextAllocator {
+  contextFilePath = CONTEXTFILE_PATH;
+  load(): Context | undefined {
+    try {
+      return new Context(JSON.parse(fs.readFileSync(this.contextFilePath, 'utf8')) as IContext);
+    } catch (error) {
+      return undefined;
+    }
+  }
+
+  save(context: Context) {
+    try {
+      fs.writeFileSync(this.contextFilePath, JSON.stringify(context), { encoding: 'utf8' });
+      return true;
+    } catch (error) {
+      return false;
+    }
+  }
+}
+
--- a/src/config/context/model.ts
+++ b/src/config/context/model.ts
@@ -26,7 +26,7 @@ export class Context implements IContext {
 export interface IContextAllocator {
   contextFilePath?: string
   load: () => Context | undefined
-  save: (context: Context) => boolean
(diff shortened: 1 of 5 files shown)


src/config/context/model.ts
```diff
@@ -0,0 +1,58 @@
...
+  contextFilePath?: string
+  load: () => Context | undefined
+  save: (context: Context) => boolean
+}
+
+@injectable()
+@registry([
+  {
+    token: 'IContextAllocator',
+    useToken: ContextAllocator
```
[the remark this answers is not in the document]
magicmatatjahu: Also, you wrote in the PR description:

> To support the future ambition of loading context files from the root directory and to be sharable I have created a separate class ContextAllocator to load context files accordingly. For now, it is loading the context file from the root file itself.

so in the future, will the `IContextAllocator` interface be implemented by other classes that can also be injected? I don't see the functionality in the `tsyringe` to have multiple implementations for a single token. Really sorry, but I don't understand this sentence:

> To support the future ambition of loading context files from the root directory and to be sharable I have created a separate class ContextAllocator to load context files accordingly

to accomplish goal by `IContextAllocator` token.

src/config/context/model.ts
```diff
@@ -0,0 +1,58 @@
...
+    try {
+      return new Context(JSON.parse(fs.readFileSync(this.contextFilePath, 'utf8')) as IContext);
+    } catch (error) {
+      return undefined;
+    }
+  }
+
+  save(context: Context) {
+    try {
+      fs.writeFileSync(this.contextFilePath, JSON.stringify(context), { encoding: 'utf8' });
```
magicmatatjahu, line 51: I prefer to explicit stringify only the properties from context `current` and `store`, not whole `context` object. It's safer. Why? Because if someone will write the function in class with arrow notation, then it will be treated as reference to the function, not method in the prototype of class. See example:

image

As you can see, arrow function is treated as normal property, not as method, so please change code to the:

```suggestion
      fs.writeFileSync(this.contextFilePath, JSON.stringify({ current: context.current, store: context.store }), { encoding: 'utf8' });
```

src/config/context/model.ts
```diff
@@ -0,0 +1,58 @@
...
+      return true;
+    } catch (error) {
+      return false;
```
magicmatatjahu, line 54: Why do you return `boolean` value, not undefined/throw Error?

magicmatatjahu (commented):
Overall good, but I have problem with IContextAllocator idea.

src/config/context/model.ts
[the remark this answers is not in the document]
Souvikns: So what I was thinking is that there are two cases where the context is used, one is to lookup context for loading input (only read functions) and the other is in the context commands where the context store is manipulated. So `ContextAllocator.load` is returning the `context` class which allows only to read data. And creating instance of the `ContextService` allows to read and manipulate the context. 

> To support the future ambition of loading context files from the root directory and to be sharable I have created a separate class ContextAllocator to load context files accordingly. For now, it is loading the context file from the root file itself.

The idea is still hazy but, this is something like having a config file, somewhat like what `jest.config.js` has where the user creates a `.asyncapi` file in the working directory writing all context. And this file could also be read over the internet. I was thinking that making any change in logic for locating the context file would be easy if implemented like this.

src/config/context/model.ts
Souvikns: Would it be a good idea to return the updated context if successful and undefined if unsuccessful?

src/config/context/model.ts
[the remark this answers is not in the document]
magicmatatjahu: As I understand it, the implementation currently allows to read single configs and will be further developed to e.g. support operation on two contexts? Am I corect?

Here comes my question. Why should you treat it as an interface injection (by `IContextAllocator` token), and not as a simple `ContextAllocator` token? Interface injection is always used when there are several implementations for this interface, which we want to inject in appropriate places in the application. Could you describe this decision. Thanks!

src/config/context/model.ts
magicmatatjahu: In case when you load context, return context/undefined is good approach, but for saving I think that it will be better when you return, as you mentioned, also context/undefined, but please make `console.warn` in the catch case, because user should know that he/she cannot save context, e.g. due to fact that hasn't permission. Please also use the `console.warn` (with error information) in the catch of `load` function.

src/config/context/model.ts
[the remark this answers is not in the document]
Souvikns: I am new to the DI concepts and I don't know if this is the right way to do this. I injected the `IContextAllocator` token because I wanted to pass in mock `ContextAllocator` during tests. when I was passing as a Simple ContextAllocator token it was not working. 

So I did something like this 

```
class MockPositiveContextAllocator implements IContextAllocator {
  load() {
    return new Context({
      current: 'test',
      store: {
        test: './test/specification.yml',
        check: './test/specfication.yml'
      }
    });
  }

  save(context: Context) {
    return context instanceof Context;
  }
}

```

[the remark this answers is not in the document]
magicmatatjahu: Ok I see

> I am new to the DI concepts and I don't know if this is the right way to do this.

No worries, good thing we explained to each other :) As I see in the code, you can still make this:

```ts
@injectable()
export class ContextService {
  private _context: Context | undefined
  constructor(
    @inject(ContextAllocator) private contextAllocator: IContextAllocator
  ) {
    this._context = this.contextAllocator.load();
  }
  
  ...
}
```

and then remove `@registry` decorator above `ContextAllocator` class. In this approach you can still use the `MockPositiveContextAllocator` mock in the test with

```tx
ctxService = new ContextService(new MockPositiveContextAllocator());
```

:) I don't know why it wasn't work for you in previous implementation, but I didn't see it, so I cannot say why 😅 

Please try this approach :)

src/config/context/model.ts
Souvikns: The lint is shouting for using `console.warn` also I was thinking that currently, the starting point is in the `cli.ts` file `render(commandsRouter(cli));`  Could we wrap it in an async function and add a catch clause there hooking another ink component for all the Error messages there. So in the whole code we can just throw error messages handle all of them in one single place.

magicmatatjahu: Even `console.info`, or `console.error`? I have seen the latest changes - we can go with `Context | undefined`, but at least give some feedback to users.

> The lint is shouting for using console.warn 

We can also change linter rules if we need. For CLI app we can enable some `console.*` methods.

> So in the whole code we can just throw error messages handle all of them in one single place.

You could have several errors at once in one command and I would prefer to buffer these errors and shown at the end - but this should be implemented in the future.

Souvikns: >You could have several errors at once in one command and I would prefer to buffer these errors and shown at the end - but this should be implemented in the future.

could work on this in the next PR.

magicmatatjahu: > could work on this in the next PR.

Make for it issue, really, because it can sound easy, but it probably won't 😅

src/config/context/model.ts
```diff
@@ -0,0 +1,58 @@
...
+
+  save(context: Context) {
+    try {
+      fs.writeFileSync(this.contextFilePath, JSON.stringify({
+        current: context.current,
+        store: context.store
+      }), { encoding: 'utf8' });
+      return context;
+    } catch (error) {
+      console.warn(error);
```
magicmatatjahu, line 53: You can disable only one line:

```suggestion
      console.warn(error); // eslint-disable-line no-undef, no-console
```

and then please remove 

```
/* eslint-disable no-undef */
/* eslint-disable no-console */
```

from the beginning of file :)

magicmatatjahu (changes requested):
Overall good, only this problem with `console.warn` Please wait for Łukasz. We'll see what he says. 

Please also change the title of PR to the `chore:` or `refactor:` because as I see, you didn't use new services in source code, but only define it, so it's not feat/fix :)

src/config/context/model.ts
magicmatatjahu: Please also add: 

```
console.warn(error);
```

in the `load` function :)

Souvikns: This still gives error 
image

magicmatatjahu: You have to write `eslint-disable-line` in this same line:

```js
console.warn(error); // eslint-disable-line no-undef, no-console
```

Souvikns: Yeah, my bad just figured that.

derberg:
I see that @magicmatatjahu reviews, so no need to wait for me with approval as it is pure refactor.

also, if it is pure refactor, nothing touching the functionality, this should not trigger a release but be a `refactor: ` commit

magicmatatjahu (approved):

magicmatatjahu:
@Souvikns Please change title and merge commit with `refactor:` prefix.

PR #83, opened 2021-09-27
Alternative example implementation using oclif

fmvilas:
I wanted to open an issue and explain the goods we'd get if we start using Oclif instead of coding and maintaining our own solution but it was too complicated to explain and too easy to actually implement the current behavior using Oclif. Bear with me, this implementation is an example and it's not fully replacing all the existing features.

I think I was one of the firsts to actually get excited with Ink in this project but the more I see it, the less I think we need it. Ink is cool but creating a great CLI experience is something that has been done many times in the past by different people and I think we can benefit from it. Ink is nothing else than a "view" framework for the CLI. I'm sure that, if we need it, we should be able to make it work together with Oclif. And here's my point, I don't think we'll ever need it. Have a look at all the [Oclif plugins](https://github.com/oclif/cli-ux) and the [Oclif documentation](https://oclif.io/docs/introduction) itself. Not to mention all the cool stuff we can do with [Inquirer](https://github.com/SBoudrias/Inquirer.js) too.

So my proposal is that we get rid of React, Ink, and other stuff and we choose this path. Implementing a new command is super easy, battle-tested by Heroku and others in different operating systems, and it supports plugins for all the stuff we may need in the future. Thoughts?


alt/asyncapi/README.md
```diff
@@ -0,0 +1,81 @@
...
+@asyncapi/cli/0.1.0 darwin-x64 node-v14.15.1
+$ asyncapi --help [COMMAND]
+USAGE
+  $ asyncapi COMMAND
+...
+```
+<!-- usagestop -->
+# Commands
+<!-- commands -->
+* [`asyncapi context [FILE]`](#asyncapi-context-file)
```
fmvilas, line 31: List of commands is automatically updated by Oclif.

alt/asyncapi/README.md
```diff
@@ -0,0 +1,81 @@
...
+$ asyncapi (-v|--version|version)
+@asyncapi/cli/0.1.0 darwin-x64 node-v14.15.1
+$ asyncapi --help [COMMAND]
+USAGE
+  $ asyncapi COMMAND
+...
+```
+<!-- usagestop -->
+# Commands
+<!-- commands -->
```
fmvilas, line 30: List of commands is automatically updated by Oclif when you use their CLI tool.

alt/asyncapi/src/commands/context/add.ts
```diff
@@ -0,0 +1,47 @@
+import {Command, flags} from '@oclif/command'
```
fmvilas, line 1: Oclif supports nested files and folders inside the `src/commands` folder. You want a subcommand? Nest a file inside a folder and you get it 💯

alt/asyncapi/src/commands/validate.ts
```diff
@@ -0,0 +1,26 @@
...
+  static flags = {
+    help: flags.help({char: 'h'}),
+  }
+
+  static args = [{name: 'file'}]
+
+  async run() {
+    const {args} = this.parse(Validate)
+
+    const file = new SpecificationFile(args.file || 'asyncapi.yaml')
```
fmvilas, line 17: To match the current implementation, this should fall back to context when a file is not specified. Didn't want to get that far for this example.

alt/asyncapi/test/commands/context.test.ts
```diff
@@ -0,0 +1,17 @@
+import {expect, test} from '@oclif/test'
```
fmvilas, line 1: Please forget about all the test files. They're generated and I didn't update them.

alt/asyncapi/README.md
```diff
@@ -0,0 +1,155 @@
...
+* [`asyncapi context`](#asyncapi-context)
+* [`asyncapi context:add [CONTEXT-NAME] [ASYNCAPI-FILE-PATH]`](#asyncapi-contextadd-context-name-asyncapi-file-path)
```
fmvilas, line 32: Commands are showing here with a colon (`:`). That's something we'll have to fix in this README. The rest of the code is working with spaces instead of colons.

Souvikns:
I think we should switch over to oclif ✅ but will miss ink. Though I think we can build features faster with oclif.

fmvilas:
Yeah, I think Ink is cool too. In case it helps, the Apollo GraphQL folks have a draft PR to integrate Oclif and Ink. Although it's 2 years old 😄 https://github.com/apollographql/apollo-tooling/pull/1697

derberg:
I did not have much time to look at all the changes and the CLI library, but I see you share only the advantages of the solution 😄 

My question is:
Now I have just this https://github.com/asyncapi/cli/blob/master/src/help-message.tsx#L104-L143 and as a result I can build a very nice colorful help like 👇🏼 for any command

Screenshot 2021-09-29 at 08 42 26

How would it work with oclif?

fmvilas:
Everything in Oclif works with plugins, even the help: https://github.com/oclif/plugin-help. If we want to customize it, we just need to [extend it](https://oclif.io/docs/help_classes) or create our own plugin, and add color using something like [Chalk](https://www.npmjs.com/package/chalk).

And speaking about colors, I don't think what we currently have is better than this non-colored help. Just see how it looks for me:

Captura de pantalla 2021-09-29 a las 9 19 17

"Commands" almost doesn't have contrast. You gotta be careful with colors, especially in the background. There's a way to make it look better with Chalk but Ink is not using Chalk properly, it seems. I had the same issue when coding the initial version of Studio. It's easy to fix (you have to call `chalk.reset()` first) but you can't do it with Ink.

And also something that was bothering me but didn't want to mention, is the fact that Ink is clearing up the terminal. That completely removes all the past output you had and you can't scroll up to see it. Didn't want to mention anything yet because that's probably something we can easily change with Ink.

Summarizing, if we want to create a great experience, we better rely on something battle-tested than rolling our own stuff. As I said, I love Ink but I think we're misusing it in this project. And nothing prevents us from integrating Oclif and Ink in the future.

Souvikns:
> I think we should resolve this is early, because later if we decide that we are going with oclif then more feature has to be ported. Here are some of my thoughts on the matter. 

### Custom Help 
Oclif has options for [custom help](https://oclif.io/docs/help_classes). So we can make colorful help strings, yeah but it is not consistent, maybe we choose the test with some other colors that stay consistent in different terminals. One problem with the custom help we have right now is that it is a lot of code and now again I have to change it because we are moving `context` command under `config`.

### Command routing 
I was checking out some multi command cli like `nest-cli` and they all use some kind of command parsing library like `commander` or such. We were building this on our own and it works fine but I guess we can push new features faster with oclif as we don't need to update the routing logic every time we add a new command. 

### The auto-documentation 
Also, I think this could help. 

### And maybe we can use Ink though I don't know if it is a antipattern 
We might still have ink components and then `render` them in the `run` function in oclif's command classes. Something like this - 


```jsx
import React from 'react';
import {Command, flags} from '@oclif/command'

import { IContext, loadContextFile } from '../../models/Context'
 
import {render} from 'ink'
import {list} from './components'

export default class ContextList extends Command {
  static description = 'list all saved contexts'

  static flags = {
    help: flags.help({char: 'h'}),
  }

  async run() {
    const {args, flags} = this.parse(ContextList)
    const ctx: IContext = loadContextFile()
    const response = Object.keys(ctx.store).map(c => ({ key: c, path: ctx.store[String(c)] }))
    render(<list response={response} />);
  }
}

```
> I haven't tested this, it is just a thought.

derberg:
Sorry but I don't have much time to look through the PR atm so need your help to answer these basic questions

- what are we exactly fixing by using `oclify`? After reading comments I assume some of our code is a boilerplate that could be replaced with out-of-the-box stuff from `oclify`. Or is it just command routing that you do not like and that we can replace with oclify?
- >  `Ink but I think we're misusing it in this project` Which part? how is this CLI different than others built with Ink?
- regarding contrast issues with help text, they won't be solved by just introducing Oclif. We will have to write our own solution for that anyway. Did I get it right?
- @Souvikns did you have a look at `test` part of the change? I think this is also important part to discuss, will it be easier to test commands or harder

I just want to make sure we do not want to switch just because there is just another cool kid on the block that we learned about.

I like oclify has CLI to do some code generation. Nice DX IMHO, but is it really a deciding factor? not really 😄

Souvikns:
> @Souvikns how much code from https://github.com/asyncapi/cli/pull/79/files you could get rid of when using oclify?

I believe we can reuse the whole thing, and just update the router and help generation part, other than that everything is reusable. 

> @Souvikns did you have a look at test part of the change? I think this is also important part to discuss, will it be easier to test commands or harder 


Also, we have individual unit tests for the logic of the tasks, so I can tell that logic will not break. but [oclif](https://oclif.io/docs/testing) has some other way of testing. I don't know much about it. I will get back once I read about it more, maybe @fmvilas can talk about this more. 

One more thing I would like to point out is that I and @magicmatatjahu were talking about [improving the error handling](https://github.com/asyncapi/cli/issues/76) that oclif has some interesting take on this.

magicmatatjahu:
I looked through the documentation and some examples and I like ocliff itself. The idea with commands as classes and static fields for flags etc is very similar to https://github.com/adonisjs/ace I think ocliff is an older project and ace was based on it - I do not know this.

As for the documentation, it's a big plus, clear with examples, but I miss the tutorial on how to write plugins from scratch.  Apparently there is a command to create a plugin structure, but exactly how to write it? 

As for DI, I think we'll still be able to use it, of course if we still want 😄 

I myself have an indifferent opinion about oclif and ink. I leave the decision to those who actively contribute to cli (@Souvikns 
 @jotamusik) and the developers themselves should decide. Switching to it and doing a review for another tool, will not be a problem for me. However, the problem is PR https://github.com/asyncapi/cli/pull/79, because if we go with ocliff there is no point in doing a review for it.

magicmatatjahu:
I looked through the documentation and some examples and I like ocliff itself. The idea with commands as classes and static fields for flags etc is very similar to https://github.com/adonisjs/ace I think ocliff is an older project and ace was based on it - I do not know this.

As for the documentation, it's a big plus, clear with examples, but I miss the tutorial on how to write plugins from scratch.  Apparently there is a command to create a plugin structure, but exactly how to write it? 

As for DI, I think we'll still be able to use it, of course if we still want 😄 

I myself have an indifferent opinion about oclif and ink. I leave the decision to those who actively contribute to cli (@Souvikns 
 @jotamusik) and the developers themselves should decide. Switching to it and doing a review for another tool, will not be a problem for me. However, the problem is PR https://github.com/asyncapi/cli/pull/79, because if we go with ocliff there is no point in doing a review for it.

fmvilas:
Note: I removed a duplicate comment from @magicmatatjahu.

Souvikns:
> I myself have an indifferent opinion about oclif and ink. I leave the decision to those who actively contribute to cli (@Souvikns
@jotamusik) and the developers themselves should decide. Switching to it and doing a review for another tool, will not be a problem for me. However, the problem is PR #79, because if we go with ocliff there is no point in doing a review for it.

My vote goes to oclif, as we can port most of the code over to oclif, I think it will not take that much time.

derberg:
@Souvikns thanks for looking into this 🙇🏼 
I share @magicmatatjahu view and for me most important is you opinion and @jotamusik 

Let us just give @jotamusik time until the end of the week to at least say `folks wait, I need more time on review` and then we can go forward.

jotamusik:
I will take a look today! Thanks for waiting! 🤗

jotamusik:
I don't see any problem, but I would like to make enphasis that the first structure of the cli, were thought in order to make changes like this easy. 
So, keeping the hexagonal architecture of the project, and using Ink or oclif only as a view layer or command router, it should be achievable. But again, we need to make an effort on keep things separately, we need to segregate the view part of the cli from it's logic in order to make this change easy. 😝 

I think we can do it. About the testing part of the ecuation, even if the test utils that comes with oclif don't match so well, on unit tests, we can tests the classes isolated, at the end, they are just classes with it's constructor so we can create instances in order to tests them.

I think that oclif it's a good option to make those people not confortable with react or frontend in general, feel more capable of doing things. So by my side.... Why not?? xD



For me Ink is not a problem, but if this makes easy the way to introduce commands and help people to feel more comfortable in the project, I don't have any problem. I don't have an opinion, the project was started with Ink, it wasn't a business requirement.

I see this, as a chance to make a clean start (obviously recycling some code base), and be sure that we are on the same page 😋

Souvikns:
So how should we proceed?

derberg:
@Souvikns I think this is a priority now before other things are added. Only thing that makes me think is https://github.com/asyncapi/cli/pull/79. I guess we should merge it first, right?

Souvikns:
> @Souvikns I think this is a priority now before other things are added. Only thing that makes me think is #79. I guess we should merge it first, right?

I think we shouldn't merge #79 but merge the changes from branch https://github.com/asyncapi/cli/tree/context-patch where the context service is fixed, in #79 I was updating the command routes which is mostly ink logic and which is not needed as we are going with oclif. 

I had created a clean oclif install in my forked branch, but I think doing this we have to add all the previous code and it would take time, rather we add oclif in the existing codebase and made the necessary changes. 

> open to suggestions.

derberg:
> we add oclif in the existing codebase and made the necessary changes.

👍🏼 

Looks like you know better what are the next steps. Go ahead 🚀

fmvilas:
@derberg I think this can be closed. @Souvikns I didn't code these tests myself, they're generated automatically by Oclif when you use their CLI: https://oclif.io/docs/generator_commands.

alt/asyncapi/README.md
Gabriel-Gotti: why dont they ever work? one thing i need to know is what is the prompts to check into my home network to see who is in it

PR #88, opened 2021-10-04
refactor: updated and decoupled contextService logic (#72)

Souvikns:
* feat: updated and decoupled contextService logic

* feat: creating new context when root context file is missing.

* feat: removed registry & stringify issue

* chore: returning context or undefined in contextAllocator save method

* feat: added warning for saving file.

* feat: lint fixes

**Description**

- This PR fixes the initial issue of context service, where it threw an error when the context file was loaded every time the context file was missing. Now it simply returns `undefined`. 
- This does not break anything and only creates new context service logic is not yet used anywhere. 

**Related issue(s)**

see also #83


Souvikns:
@magicmatatjahu has once reviewed this, just need the approval to merge.

derberg (approved):
derberg via Axolo. :shipit:

PR #95, opened 2021-10-11
refactor: porting logic from ink routers to oclif routers.

Souvikns:
**Description**

- Ported logic from ink to oclif, so now `./bin/run` has all the functionalities previously supported. 
- Ink code is still present so nothing should break yet. 

> Need opinion on deleting the ink code in this PR or the next PR. 

**Related issue(s)**

see also #83


Souvikns:
I am having some issues with setting up `oclif/test`, so shall I push setting up tests to the next PR where I remove ink logic and update the test. 

> Ready for review @derberg @magicmatatjahu

PR #110, opened 2021-11-02
feat: created specificationUrl class

### LINKED ISSUE #7: Support passing url to AsyncAPI file
#### Reason/Context

The initial version of CLI that we soon merge in supports files from the file system. We should support remote files as parser and generator already support it. 

#### Description
- User passes `--context` as `https://raw.githubusercontent.com/asyncapi/asyncapi/master/examples/2.0.0/simple.yml`.
- Even though the parser has `parseFromURL` option, it doesn't mean that all tools in the figure will support it from day 1. For example, the diff library doesn't have to include any dependency to `fetch` to pull spec from the remote. This means we need to have independent functionality that fetches the spec from remote if we detect that context points to URL

Souvikns:
**Description**
This PR aims to add support to fetch specification file from url. 

- Ultimately the cli will be able to take url as a parameter to resovle commands 
- Save url in context 
- Fetch specification file from url. 

**Related issue(s)**

Resolves #7


Souvikns:
`load` function is returning `SpecificationFile` instance but to add support for URL paths it feels weird to have fetch function in the `SpecificationFile`. We could create another `SpecificationUrl` class that will fetch the data, but I think we should change `SpecificationFile` to `Specification` and have two separate functions `loadFromFile` and `loadFromUrl` to load the spec file accordingly. 

So our load function would look something like this 
```ts
const spec = await load(filePath);
await parser.parser(spec.text());
```

src/models/SpecificationUrl.ts
```diff
@@ -0,0 +1,30 @@
+import fetch from 'node-fetch';
+
+export default class SpecificationURL {
```
fmvilas, line 3: Isn't SpecificationURL class just checking a URL? I mean, it has nothing to do with the spec but just fetching from a URL and validating a URL. Maybe we can just add these functions to a "helper" library? Also, I noticed you're not using this class anywhere in the code. Is it because it's a Draft PR?

src/models/SpecificationFile.ts
```diff
@@ -13,13 +14,24 @@ const allowedFileNames: string[] = [
 const TYPE_CONTEXT_NAME = 'context-name';
 const TYPE_FILE_PATH = 'file-path';
 
+export class Specification {
```
fmvilas, line 17: Agree with removing the `File` part 👍 That's a good point.

src/models/SpecificationUrl.ts
Souvikns: will not have the `SpecificationURL` class just checking different approaches

fmvilas:
In general, my advice is that we should be reducing the cognitive load of the code, and I think creating more classes and more modules that depend on each other only makes it harder to understand. Remember what we talked about the "hops" you have to do when reading a code. If the only logic that's really different is validating a URL and fetching from a URL, then probably all you need are just "helpers" or external modules that already do that.

Souvikns:
@fmvilas should `startStudio` function take file path as input or spec as input. To create URL support I was thinking of making it in a way that every command which needs to read specification from somewhere can just use the load function but it will not get file path or URL or context but rather the specification string or json itself. Studio needs `filepath` and port to start so it does not support URL as of now. 

Also `Should studio load specification from URL` as it won't be able to update it anyway.

fmvilas:
Yeah, I don't think Studio needs to be able to open a URL, at least for now. As you said, you'll not be able to edit it anyway so what's the point 🤷‍♂️

test/commands/validate.test.ts
```diff
@@ -34,6 +34,16 @@ describe('validate', () => {
         expect(ctx.stderr).to.equals('ValidationError: There is no file or context with name "./test/not-found.yml".\n');
         done();
       });
+
+    test
+      .stderr()
+      .stdout()
+      .command(['validate', 'https://bit.ly/asyncapi'])
```
derberg, line 41: I see that @fmvilas already reviews, so from my point of view I only suggest to log a followup issue, something good for the first contribution, to make sure tests do not rely on external resources. Basically, external resources are risky to use in tests and can cause tests to be flaky, like randomly failing just because the external resource is not available. So follow up should describe that tests should actually start independent server hosting asyncapi file, like we do in parser -> https://github.com/asyncapi/parser-js/blob/master/package.json#L20

derberg:
From my point of view important is that help message `spec path or context-name` is updated

src/commands/validate.ts
```diff
@@ -13,14 +13,14 @@ export default class Validate extends Command {
   }
 
   static args = [
-    { name: 'spec-file', description: 'spec path or context-name', required: false },
+    { name: 'spec-file', description: 'spec path, context-name or spec url', required: false },
```
fmvilas, line 16:
```suggestion
    { name: 'spec-file', description: 'spec path, url, or context-name', required: false },
```

src/commands/validate.ts
```diff
@@ -34,8 +34,14 @@ export default class Validate extends Command {
       }
     }
     try {
-      await parser.parse(await specFile.read());
-      this.log(`File ${specFile.getPath()} successfully validated!`);
+      if (specFile.getFilePath()) {
+        await parser.parse(specFile.text());
+        this.log(`File ${specFile.getFilePath()} successfully validated!`);
+      }
+      if (specFile.getURLPath()) {
```
fmvilas, line 41:
```suggestion
      } else if (specFile.getURLPath()) {
```

src/models/SpecificationFile.ts
```diff
@@ -65,10 +108,21 @@ export async function nameType(name: string): Promise<string> {
...
   } catch (e) {
+    if (await isURL(name)) {return TYPE_URL_PATH;}
     return TYPE_CONTEXT_NAME;
   }
 }
 
+export async function isURL(urlpath: string): Promise<boolean> {
+  try {
+    const url = new URL(urlpath);
+    if (url.protocol === 'http:' || url.protocol === 'https:') {return true;}
```
fmvilas, line 119:
```suggestion
    return (url.protocol === 'http:' || url.protocol === 'https:');
```

This way you can get rid of the `return false;` in line 123 (not the one inside the catch).

src/models/SpecificationFile.ts
```diff
@@ -12,14 +13,52 @@ const allowedFileNames: string[] = [
...
+
+  text() {
+    return this.spec;
+  }
+
+  getFilePath() {
+    return this.filePath;
+  }
+
+  getURLPath() {
```
fmvilas, line 36: I think you shouldn't call this "URL path". That's a URL.

I mean, this is a URL: `https://www.asynapi.com/blog`.
And this is the URL path: `/blog`.

They're not the same and I think here you mean the full URL, not just the path part.

Why not `getFileURL()`?

```suggestion
  getFileURL() {
```

If you change it, you'd have to change everything else following similar notation, like `const TYPE_URL_PATH = 'url-path';`

fmvilas (changes requested):

derberg (approved):
All good from my side but wait for @fmvilas as he made some good quality review

also you have some merge conflicts in package.json

derberg:
@fmvilas as you initially requested changes you need to approve before we can merge

magicmatatjahu:
Looking at the changes, it's ok for me. You just have to wait for the Fran :)

fmvilas (approved):

fmvilas:
Sorry for the delay. Looks good to me 🚀

Issue #111, opened 2021-11-02
allow user to pick a template from the `new` command


boyney123:
#### Reason/Context
Please try answering few of those questions

At the moment the `new` command generates a basic AsyncAPI for the user, but it would be nice if we asked them some more questions like:

- What protocol would you like to use (MQTT, WebSockets, HTTP, None)
- Based on what they select it generates using the correct template.

## How will this change help?

I think just more flexibility for the CLI tool and give people more options, users will come from all backgrounds, giving them a helping hand with their chosen protocol might really help.

boyney123:
Added code in #131

derberg:
@boyney123 this one is implemented, right?

boyney123:
Yeah merged part of #131 , will close 👍

PR #113, opened 2021-11-02
feat: allow user to select template on `new` command

boyney123:
**Description**

This feature will ask the user what template they would like to use.

When creating a new asyncapi file using the `new` command, they will be asked if they want to use one of our templates (as a starting point).

This will hopefully allow people to get started easier using mqtt, websockets, kafka for example.

Adding future templates should be straight forward too.

Related issues #111


Souvikns (approved):
LGTM 🚀

src/commands/new.ts
```diff
@@ -7,13 +7,15 @@ import { resolve } from 'path';
 
 const { writeFile, readFile } = fPromises;
 const DEFAULT_ASYNCAPI_FILE_NAME = 'asyncapi.yaml';
+const DEFAULT_ASYNCAPI_TEMPLATE = 'basic.yaml';
```
fmvilas, line 10: Did you mean `simple.yaml`?

```suggestion
const DEFAULT_ASYNCAPI_TEMPLATE = 'simple.yaml';
```

src/commands/new.ts
```diff
@@ -7,13 +7,15 @@ import { resolve } from 'path';
...
 
 export default class New extends Command {
   static description = 'creates a new asyncapi file';
 
   static flags = {
     help: flags.help({ char: 'h' }),
     'file-name': flags.string({ char: 'n', description: 'name of the file' }),
+    'selected-template': flags.string({ char: 'n', description: 'name of the template to use' }),
```
fmvilas, line 18: A few things on this:

1. Shall we call it `template` or `example`? I think we're not giving the user a template they can fill in the gaps but an example they can modify.
2. I'd just go for `example` (or `template`, if we prefer). It's implicit that you're selecting it because you're passing the flag.
3. The char can't be `n` as it's already being used for the `file-name` flag. Maybe we can go for `e` (example) or `t` (template).

So here's my suggestion you can apply if you like it:

```suggestion
    'example': flags.string({ char: 'e', description: 'name of the example to use' }),
```

fmvilas (changes requested):

src/commands/new.ts
boyney123: Thanks @fmvilas good spot!

src/commands/new.ts
boyney123: Yeah good point @fmvilas , yeah will make the change to move to `example` a few places to change so will make them seperate. Thanks

boyney123:
Thanks, Fran, changes have been made to the PR. 

If you are happy, maybe we can merge @Souvikns ?

Souvikns (approved):

Souvikns:
@boyney123 it is out-of-date with the base branch

boyney123:
ah sorry, @Souvikns thanks! Just updated

Souvikns:
nice now once @fmvilas approves we can merge it.

derberg:
aren't we rushing too much here folks? A similar feature I see is introduced in the studio and official examples are duplicated there too. So we basically duplicate lots of content here and there and we should slow down and think how to do it better, otherwise, spec release 2.3 and all next ones will get even more complicated 😅 

Unless you already talked about it and I missed it, so sorry for party-pooping 😄

boyney123:
> So we basically duplicate lots of content here and there and we should slow down and think how to do it better, otherwise, spec release 2.3 and all next ones will get even more complicated

Yeah understand what you mean, and at what point should we automate this stuff? I guess at the moment if 2.3 came out, it would be a 5min job to update these ones, and if that happens often then we automate that part later?

Guess it comes down to at what time is a good time to automate things, I tend to do it when it becomes painful or I'm replicating a lot of effort.

But yeah I guess another option here, would be the push model you mentioned before @derberg, having the examples pushed from the main spec repo and consumed here? Is that the kind of pattern you mean?

derberg:
@boyney123 I'm always advocating for push model and automating stuff that we already know need automation. As a [release coordinator](https://github.com/asyncapi/spec/blob/master/RELEASE_PROCESS.md) of the last 2 releases, I assure you that we need to go towards "full" release automation, and definitely not "5min here" and "5min there" 😄 as then we have 10min 😆 I'm sure you know what I mean.

We have some push model automation in `spec` repo, I think it is just enough to extend it with new repos, and probably only support push on merge to master in spec repo 🤔 except of the fact that it is not just a simple switch of examples, but a release of new studio and cli that also support new spec version 🤔 

yeah, not so trivial, might be good to have some discussion about it, especially involving @magicmatatjahu. I was even thinking if playground/studio + cli should not be mandatory repos people should contribute to when working on new spec feature proposals

magicmatatjahu:
@derberg 

> We have some push model automation in spec repo, I think it is just enough to extend it with new repos, and probably only support push on merge to master in spec repo 🤔 except of the fact that it is not just a simple switch of examples, but a release of new studio and cli that also support new spec version 🤔

This can be problematic, but there is nothing to stop us to just copying an examples to studio/cli and on the studio/cli side use the examples from copied folder. If it comes to new version, I don't see any problem, because studio/cli itself should be prepared for new spec version before merging it to master on spec repo - same as 2.1 and 2.2 we did.

> yeah, not so trivial, might be good to have some discussion about it, especially involving @magicmatatjahu. I was even thinking if playground/studio + cli should not be mandatory repos people should contribute to when working on new spec feature proposals

I don't think so that it's a good idea to force contribute to that tools when someone want proposes something new in spec. For studio it can ends with updating:
- react-component
- in some cases also html-template
- server-api
- visualiser
- in future builder
- etc.

so as you can see it will be huge contribution. Let's not complicate people's lives 😄

derberg:
> This can be problematic

in what way?

> I don't see any problem, because studio/cli itself should be prepared for new spec version before merging it to master on spec repo

it is true in case of studio because of converter integration (although we should also think on making it automated), but nothing to do in case of the CLI, just bump of the parser that is anyway done in automated way.

so yeah, looking on previous projects I worked in before AsyncAPI and looking on AsyncAPI now where we want to hand over release coordination to anyone for the community -> it must be super simple, involve as little amount of people as possible, and as small amount of steps as possible. You did a release with me last time, you know how complex it is, and it should not be.

So yeah, what is the `problematic` part. I guess sync `push` from repo with bump of dependencies?

Also how these examples should work, I guess we agree that always the latest version should be supported here? which makes automation easier

boyney123:
> so yeah, looking on previous projects I worked in before AsyncAPI and looking on AsyncAPI now where we want to hand over release coordination to anyone for the community -> it must be super simple, involve as little amount of people as possible, and as small amount of steps as possible. You did a release with me last time, you know how complex it is, and it should not be.

Yeah, that's a fair point. 👍 

> Also how these examples should work, I guess we agree that always the latest version should be supported here? which makes automation easier

I think so, to be honest, I think we should support the latest versions until we get a request not to do that, or support older versions. 

@derberg pushing these from the website repo sounds like a plan, are you OK to pair with me on this? Not sure how to do that to be honest....

derberg:
I'm just having second thoughts if the `push` strategy we have for the docs should also be used here. Maybe in case of libraries and code this should be a part of the build/publish process.

for example, in some projects we update API.md or Readme ToC on a build level, as a separate workflow that updates stuff that humans normally forget about. Maybe we should treat examples the same way, especially that they are part of distro 🤔 Maybe we need a separate issue and also involve others to get different ideas

magicmatatjahu:
> So yeah, what is the problematic part. I guess sync push from repo with bump of dependencies?

Yeah, I thought about that.

> Also how these examples should work, I guess we agree that always the latest version should be supported here? which makes automation easier

If we will go with `3.0.0` version in the April/June next year, I think that we should always have examples for latest major version, it means e.g. `2.3` and `3.1`. What do you think?

boyney123:
I'm going to close the PR in favour for #131, which allows users to select a template, if templates are there and also uses a script for (pull) model to allow users to select examples from `website` repo.

Issue #114, opened 2021-11-02
No file extension when user defines the name of the file (using new command)


boyney123:
#### Describe the bug

When user uses the `new` command they can provide the filename of the file.

For example, if I put `myNewFile` as a value, the file is created OK, but no file extension is added.

I think it might be a better experience if we detect if no extension is on the file and just put the `yaml` extension of the file for them?

jjung99:
Hi, I would like to take a look at this issue

boyney123:
Hey @jjung99 ,

Sorry I managed to find some time earlier and got a PR raised, but be good if you fancy having a look and review it? https://github.com/asyncapi/cli/pull/117

Or if you think of a better way to do it?

jjung99:
It is fine! I just wanted to take a look at others' projects and try to fix a bug if any for learning purposes. Thanks for the comment!

PR #117, opened 2021-11-02
feat: adding default file extension for users

boyney123:
**Description**

When user uses the `new` command they are asked for the `filename`. 

If the user does not provide a file extension the file is created.

This PR will check to see if any file extension has been made, and will add `.yaml` onto the filename if no extension has been provided.

### Why

Just think it's a slightly better developer experience than allowing them to have files without extensions as async API files.

Related issue #114

### DIFF
--- a/src/commands/new.ts
+++ b/src/commands/new.ts
@@ -81,7 +81,7 @@ export default class New extends Command {
   async createAsyncapiFile(fileName:string) {
     const defaultAsyncapiFile = await readFile(resolve(__dirname, '../../assets/asyncapi.yaml'), { encoding: 'utf8' });
 
-    const fileNameHasFileExtension = fileName.includes('.')
+    const fileNameHasFileExtension = fileName.includes('.');
     const fileNameToWriteToDisk = fileNameHasFileExtension ? fileName : `${fileName}.yaml`;
 
     try {
--- a/src/commands/new.ts
+++ b/src/commands/new.ts
@@ -81,7 +81,7 @@ export default class New extends Command {
   async createAsyncapiFile(fileName:string) {
     const defaultAsyncapiFile = await readFile(resolve(__dirname, '../../assets/asyncapi.yaml'), { encoding: 'utf8' });
 
-    const fileNameHasFileExtension = fileName.split('.').pop() !== fileName;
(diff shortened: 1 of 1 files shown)


src/commands/new.ts
```diff
@@ -81,17 +81,20 @@ export default class New extends Command {
   async createAsyncapiFile(fileName:string) {
     const defaultAsyncapiFile = await readFile(resolve(__dirname, '../../assets/asyncapi.yaml'), { encoding: 'utf8' });
 
+    const fileNameHasFileExtension = fileName.split('.').pop() !== fileName;
```
fmvilas, line 84: What about this?

```suggestion
    const fileNameHasFileExtension = fileName.includes('.')
```

Creating an array and removing elements is an expensive operation. I think we can just check for the presence of a `.` char.

fmvilas (changes requested):

src/commands/new.ts
boyney123: Yeah happy to go with that, would only mean things like `application.random.yaml` would not be valid, although I have no idea who would do this stuff.... pop means we always get the last one regardless of the amount of `.` in the filename.

What do you think? Am I over thinking it :D?

fmvilas: But if someone does `application.random.yaml` then it means `application.random` is the name, isn't it?

fmvilas: None of our solutions will cover that case, I think. Unless we obtain the extension and check if it's `yml`, `yaml`, or `json` and therefore dicard `random` or any other string 🤔

boyney123: > None of our solutions will cover that case, I think. Unless we obtain the extension and check if it's yml, yaml, or json and therefore dicard random or any other string 🤔

Yeah think you are right. Hmmm think let's go for includes and see if we get any issues about it, and in future if we do we can sniff out the extensions we want

fmvilas (approved):

boyney123:
Thanks for the review @fmvilas , if you are cool @Souvikns maybe it's good to go?

Souvikns:
Yeah I think this is good to go, I am merging it.

Souvikns (approved):

Issue #127, opened 2021-11-03
should `start studio` try and find a local file?


boyney123:
#### Reason/Context

Was playing around with the CLI tool and trying to get the studio running on my machine, but when I ran `asyncapi start studio` it was asking me for `context` to be set ( I assume that is a CLI thing ).

When I run `asyncapi start studio --file=asyncapi.yml` it works as expected.

I wonder if the `asyncapi start studio` should at least try and find a local asyncapi file in the directory before falling back to that `context` solution?

### Why we need this improvement?

Just means we could tell people to run `asyncapi start studio` in any directory with an AsyncAPI file. Which I think is a better developer experience vs having to know about this `context` stuff.

boyney123:
Any thoughts on this @jonaslagoni @Souvikns?

Or maybe I'm not understanding how context works

jonaslagoni:
Good question, I see pros and cons for both tbh.

I guess the question falls back to, what is the best user experience when you need to perform actions with specific (or multiple) AsyncAPI documents.

I could see it as being a mix tbh. Where if you want to use context use that, or if you want to specifically provide an input to a specific command, you should be able to do that.

Regarding this issue, I would not assume the CLI to automatically look for a local file, as it is quite error-prone, cause I think it would rarely be called `asyncapi.yml`, so going through all files would be, too much IMO 😄

derberg:
`asyncapi start studio` should be consistent with the rest of the CLI. There was and I hope still is 😄 a fallback mechanism for loading AsyncAPI file, it basically looks for context, default file `asyncapi.yml` in the context of where command is run, etc!

so yes, we should do automatic for the use, same as in other commands, but also allow to pass `--file` or `--context` flags to overwrite defaults. And yes, we should not force user to learn about `context` concept but throw a full help message that explains what different options user can pick, now I see `ContextError: No context is set as current, please set a current context.` but I'm 100% sure there was a better error displayed with full help text, but don't see it anymore 😢

boyney123:
Ahh yeah I just was reading around and found that is already does fall back to files (well tries too).

https://github.com/asyncapi/cli/blob/master/src/models/SpecificationFile.ts#L43

Looks like this line is happening before https://github.com/asyncapi/cli/blob/master/src/models/SpecificationFile.ts#L50

When nothing it passed it will go to load context and just fail if nothing is there, I think this fix is moving those two around.

So flow:

- Use file/context if provided
- Check for local file

I think the current flow is :

[...]

PR #130, opened 2021-11-04
feat: set fallback to read local dir for asyncapi file if no context is present

boyney123:
**Description**

A fix for #127 it now will check the file directory before finally checking the context again (and throwing errors).


Souvikns:
This would be a big decision, as the overall functionality would change, I think we need to ask @derberg and @fmvilas for this. Because when we decided on the order of the fallback we agreed on this because the functionality was for the power users who would want to be using the current context if he has spent time setting it up.

boyney123:
ah OK.

```
if (filePathOrContextName) {
    const type = await nameType(filePathOrContextName);
    if (type === TYPE_CONTEXT_NAME) {
      return loadFromContext(filePathOrContextName);
    } 
    await fileExists(filePathOrContextName);
    return new SpecificationFile(filePathOrContextName);
  }
```

Isn't this the part that would check the context? Happens before the fallback?

Or are you talking about the users that setup context, but don't pass in the filepath or context name?

Souvikns:
That path figures out what that input is. Is the input a file path or a context name this does not run as we are not passing any input as the command is `asyncapi start studio`

fmvilas:
Yeah, this same bug hit me during my QCon presentation. I typed `asyncapi start studio` and it complained there wasn't any context but I had a file in the current working directory. I think this should change.

derberg (changes requested):
- there is come console.log in code
- make sure PR title follows imperative mode. Anyway title also needs review to reflect implementation
- context is still important, it is just that in case of no default context, there should not be an error like we get now, but fallback mechanism should work further and in the end the file from root should be picked
- some tests are failing

boyney123:
Thanks, @derberg. 

> there is come console.log in code 

😅👀 let's pretend you didn't see anything 😂 

> make sure PR title follows imperative mode. Anyway title also needs review to reflect implementation

Do you mean [imperative mood](https://en.wikipedia.org/wiki/Imperative_mood)? The first time I heard of this, can you give an example of what you would expect? I changed it, let me know what you think :).

> context is still important, it is just that in case of no default context, there should not be an error like we get now, but the fallback mechanism should work further and in the end, the file from root should be picked

OK makes sense. I changed the flow so now will check context, if that fails it will then try and read the local dir for any files. Is this what you mean?

derberg:
> let's pretend you didn't see anything

see what?

> Do you mean imperative mood

yes 😅 basically follow https://chris.beams.io/posts/git-commit/#imperative so we stay consistent and we don't get one commit like `adding.....` and right after `add.....`. Best would be imho `feat: set fallback to read local dir for asyncapi file if no context is present`.  It is of any command and also rather improvement and not fix. What do you think?

> OK makes sense. I changed the flow so now will check context, if that fails it will then try and read the local dir for any files. Is this what you mean?

yup, exactly

boyney123:
> chris.beams.io/posts/git-commit/#imperative

Thanks, @derberg this is great and helps loads! I will amend the title 🙇

derberg (approved):
LGTM
@Souvikns thoughts?

We still have the problem of course that if no local file is present then `ContextError: No context is set as current, please set a current context.` is shown. But I think there is a bigger problem with how we are passing errors, as so far I see `https://github.com/asyncapi/cli/blob/master/src/errors/validation-error.ts#L19` will never get called even if I do `asyncapi validate` without context or local file

Souvikns:
> We still have the problem of course that if no local file is present then `ContextError: No context is set as current, please set a current context.` is shown. 

If no file is present in the working directory and also there is no current context set, then `ContextError: No context is set as current, please set a current context.` will be thrown, also now it is picking the asyncAPI file from the working directory if there is any, thanks to @boyney123  so I think this problem is solved. 

> But I think there is a bigger problem with how we are passing errors, as so far I see https://github.com/asyncapi/cli/blob/master/src/errors/validation-error.ts#L19 will never get called even if I do asyncapi validate without context or local file

Yeah, this is not something that is supposed to happen. Ok so let me write about the issue in more details:

So when I run `asyncapi validate` and there is no file in the working directory and no current context is set, previously it threw `no-spec-found` error which had a message like this - 
```
  - You can provide a path to the AsyncAPI file: asyncapi validate path/to/file/asyncapi.yml
  - You can also pass a saved context that points to your AsyncAPI file: asyncapi validate mycontext
  - In case you did not specify a context that you want to use, the CLI checks if there is a default context and uses it. To set default context run: asyncapi context use mycontext
```

But now this message never gets printed, and instead of this we have a error message like this 
```
ContextError: No context is set as current, please set a current context.
```
This error is getting thrown here: 
https://github.com/asyncapi/cli/blob/b9cc9388d0aac81caa8d5bd9322b4c430127cb53/src/models/SpecificationFile.ts#L43

When we do not find any current context we are throwing `SpecificationFileNotFound` error. 

> @derberg I think we should open a new issue for this.

derberg:
@Souvikns yeah, definitely a separate issue, created it here https://github.com/asyncapi/cli/issues/148

@boyney123 just update with latest master and we get that one merged 🥳

boyney123:
Cool thanks @derberg @Souvikns 

All updated, good to go 👍

PR #131, opened 2021-11-05
feat: on `asyncapi new` give users the ability to select an example as a starting point

boyney123:
## Feature

`asyncapi new` at the moment creates a new asyncapi file only using the `simple.yaml` example.

This feature allows users to use an example (from here https://github.com/asyncapi/spec/tree/master/examples) as a starting point (which is an extra optional question).

### Why

Just gives people more options and ways to start with AsyncAPI vs getting started with the default file. 

For example if the user is interested in WebSockets they could select our web sockets example as a starting point.

### How it works

image

- new script has been added which will fetch a zip file from https://github.com/asyncapi/spec, idea from @derberg 
- new script will unzip the contents, find the examples and make this data accessible to the user on the CLI
- the [parser](https://github.com/asyncapi/parser-js) is also used to display user friendly info based on templates.

If you want to watch 📹 it in action: https://asyncapi.slack.com/files/U020GN9C6FM/F02L5152LTG/cli-with-examples.mp4

#### Notes

The whole feature is optional, the user can still continue without picking any examples, and if no examples have been fetched the CLI will continue to work and user will be not be asked the question.


boyney123:
Hey @derberg any thoughts on this pull model for the examples?

boyney123:
Just copying over convo from slack: https://asyncapi.slack.com/archives/CQVJXFNQL/p1636358757198500?thread_ts=1636124132.191400&cid=CQVJXFNQL

image

scripts/fetch-asyncapi-example.js
```diff
@@ -0,0 +1,93 @@
...
+const TEMP_ZIP_NAME = 'spec-examples.zip';
+
+const fetchAsyncAPIExamplesFromExternalURL = () => {
+  return new Promise((resolve) => {
+    request(SPEC_EXAMPLES_ZIP_URL)
+      .pipe(fs.createWriteStream(TEMP_ZIP_NAME))
+      .on('close', () => {
+        console.log('Fetched ZIP file');
+        resolve();
+      });
```
fmvilas, line 20: Should we add error handling here?

```suggestion
      })
      .on('error', () => {
        reject();
      });
```

*Note: not sure if this code is correct.*

src/commands/new.ts
```diff
@@ -7,13 +7,16 @@ import { resolve } from 'path';
...
 const DEFAULT_ASYNCAPI_FILE_NAME = 'asyncapi.yaml';
+const DEFAULT_ASYNCAPI_TEMPLATE = 'simple.yaml';
 
 export default class New extends Command {
   static description = 'creates a new asyncapi file';
 
   static flags = {
     help: flags.help({ char: 'h' }),
     'file-name': flags.string({ char: 'n', description: 'name of the file' }),
+    'use-example': flags.string({ char: 'u', description: 'boolean if they want to explore examples'}),
```
fmvilas, line 18: Three things here:

1. `boolean if they want to explore examples`. Who are "they"? My recommendation is that you always try to avoid directing to anyone (see other flags) or, if needed, direct your message to the reader using second person singular, i.e., "you(r)(s)". I think this one can become `ask for examples in the interactive terminal`.
2. That said, I'd completely remove this flag and would always show it unless the `example` flag is passed. Just like we do with `port`, for instance.
3. You're not using this flag at all in your code 😄

src/commands/new.ts
```diff
@@ -57,6 +64,31 @@ export default class New extends Command {
...
+      const exampleFiles = await readFile(resolve(__dirname, '../../assets/examples/examples.json'), { encoding: 'utf8' });
+      examples = JSON.parse(exampleFiles);
+    } catch (error) {
+      // no examples found
+    }
+
+    if (examples.length > 0) {
+      questions.push({
+        name: 'use-example',
+        message: 'would you like to start your new file from one of our examples? (https://github.com/asyncapi/spec/tree/master/examples)',
```
fmvilas, line 77: Can we make this message shorter?

fmvilas (changes requested):

src/commands/new.ts
boyney123: Yeah sure, any ideas? What do you think of having the link in there to show them what examples we are talking about, or not bothering? I'm in two minds about it....

fmvilas: Yeah, not sure. Maybe just remove the link. Something we can do is show a message after an example is selected saying: "You can find more examples at https://github.com/asyncapi/spec/tree/master/examples".

boyney123: Cool thanks, will take a look mate

src/commands/new.ts
boyney123: Yeah, thanks, a mistake from me, maybe copy and paste job, but yeah we don't need this as a flag at all.

magicmatatjahu: My 2 cents: I think that it should be `show examples in the interactive terminal` or `explore examples...` or something similar, because for me `ask for...` indicates 3rd person, not 2nd (cli user).

src/commands/new.ts
boyney123: Yeah, I just removed the URL for now completely going to keep it simple for now.

scripts/fetch-asyncapi-example.js
boyney123: Thanks, will add it, yeah I believe it is from what I read about the `createReadStream` api, thanks!

src/commands/new.ts
magicmatatjahu: Ok, nevermind, I see that param is removed 😅

scripts/fetch-asyncapi-example.js
```diff
@@ -0,0 +1,95 @@
...
+      const parsedSpec = await parser.parse(exampleContent);
+      const title = parsedSpec.info().title();
+      const protocols = listAllProtocolsForFile(parsedSpec);
+      return {
+        name: protocols ? `${title} - (protocols: ${protocols})` : title,
+        value: example
+      };
+    } catch (error) {
+      // Failed for somereason to parse this spec file, ignore for now
+    }
```
magicmatatjahu, line 74: You should at least log that error

```suggestion
    } catch (error) {
      // Failed for somereason to parse this spec file, ignore for now
      console.error(error);
    }
```

src/commands/new.ts
```diff
@@ -57,6 +63,31 @@ export default class New extends Command {
...
     }
 
+    try {
+      const exampleFiles = await readFile(resolve(__dirname, '../../assets/examples/examples.json'), { encoding: 'utf8' });
+      examples = JSON.parse(exampleFiles);
+    } catch (error) {
+      // no examples found
+    }
+
+    if (examples.length > 0) {
```
fmvilas, line 73: If flag `--example` is passed, we should not be asking for an example.

```suggestion
    if (!selectedTemplate && examples.length > 0) {
```

fmvilas (changes requested):

magicmatatjahu (commented):
Only one suggestion :)

scripts/fetch-asyncapi-example.js
boyney123: Cool thanks @magicmatatjahu

src/commands/new.ts
boyney123: ah yes, thanks!

src/commands/new.ts
```diff
@@ -68,18 +99,25 @@ export default class New extends Command {
...
     if (questions.length) {
       const answers: any = await inquirer.prompt(questions);
+
       if (!fileName) {fileName = answers.filename as string;}
+      if (!selectedTemplate) {selectedTemplate = answers.selectedTemplate as string;}
       if (openStudio === undefined) {openStudio = answers.studio;}
     } else {
       fileName = DEFAULT_ASYNCAPI_FILE_NAME;
+      selectedTemplate = DEFAULT_ASYNCAPI_TEMPLATE;
     }
```
magicmatatjahu, line 109: I think that here should be made fallback to the default values, outside else:

```suggestion
    }
    fileName = fileName || DEFAULT_ASYNCAPI_FILE_NAME;
    selectedTemplate = selectedTemplate || DEFAULT_ASYNCAPI_TEMPLATE;
```

magicmatatjahu (changes requested):

scripts/fetch-asyncapi-example.js
fmvilas: This is still not fixed (I'm saying because you asked me for review again 😄)

fmvilas (changes requested):

Souvikns (commented):
Should we add the asset folder to the files object in the `package.json`

scripts/fetch-asyncapi-example.js
boyney123: Ah, thanks my bad 😅

src/commands/new.ts
boyney123: Cool thanks @magicmatatjahu , made the change 👍

scripts/fetch-asyncapi-example.js
boyney123: Ok, think it's done now, sorry lol, thought I made the change but looks like I did'nt

boyney123:
> Should we add the asset folder to the files object in the package.json

Thanks @Souvikns! Forgot about that, all done 🙇

scripts/fetch-asyncapi-example.js
```diff
@@ -0,0 +1,98 @@
...
+const fetchAsyncAPIExamplesFromExternalURL = () => {
+  return new Promise((resolve, reject) => {
+    request(SPEC_EXAMPLES_ZIP_URL)
+      .pipe(fs.createWriteStream(TEMP_ZIP_NAME))
+      .on('close', () => {
+        console.log('Fetched ZIP file');
+        resolve();
+      }).on('error', () => {
+        reject();
+      });
```
fmvilas, line 22: Optional suggestion 😄 

```suggestion
      }).on('error', reject);
```

magicmatatjahu (dismissed):

Souvikns (approved):

magicmatatjahu (dismissed):

fmvilas (dismissed):

magicmatatjahu (approved):
LGTM! 🚀

PR #133, opened 2021-11-10
fix: fixing some injection sink issues

boyney123:
**Description**

We are getting feedback from our tools saying we have some `injection sink` issues

image

This small PR resolves the issues we have.


magicmatatjahu (approved):
LGTM!

boyney123:
@Souvikns if you are happy, here is another small fix 👍

Issue #58, opened 2021-08-30
Integrate `asyncapi/diff` in the CLI


aayushmau5:
Since the [`asyncapi/diff`](https://github.com/asyncapi/diff/) library is, now, usable, we can integrate it into the AsyncAPI CLI.

My idea for the command usage:
```
asyncapi diff --files <file1.yml> <file2.yml>
asyncapi diff --contexts file1 file2
```

So, we need to think about some things:
- How to output the data
- Where to output the data
- The flags which will specify the type of output 

**How to output the data:**

For this, I'm thinking of having a flag `--<format>`. For started, we can provide output in JSON format. Then, in the future, we can go to HTML based output, markdown, etc.

```
asyncapi diff --files <file1.yml> <file2.yml> --json
asyncapi diff --files <file1.yml> <file2.yml> --html
```
**Where to output the data:**

Now the question arises, where do we output the data. In a file? In the terminal?
Looking at OpenAPI [diff](https://github.com/OpenAPITools/openapi-diff) tool, we can also provide an **optional** `<file>` option along with `--<format>` flag. `--<format> <file>`. The default output will be in the terminal.

```
asyncapi diff --files <file1.yml> <file2.yml> --json <file-output>
asyncapi diff --files <file1.yml> <file2.yml> --html <file-output>
```

**The flags which will specify the type of output:**

Since the diff library provides a bunch of helpers function which provide different output(ex. only `breaking` changes, only `non-breaking` changes), we can provide a flag to do this through the CLI. The default will provide the full output.
ex.
```
asyncapi diff --files <file1.yml> <file2.yml> --json <file> --breaking # for breaking changes
asyncapi diff --files <file1.yml> <file2.yml> --json <file> --non-breaking # for non-breaking changes
```

Any thoughts on this idea?

derberg:
Hey, there is a topics covered here that are important no matter if we work on diff or not: the output

atm, library outputs on terminal only. There is no `--output` option so far. This is independent from `diff` as we need output to file functionality in many different commands, for relations finding or for optimizations. This needs to be implemented separately

regarding optional flags, we so far tend to always have name, so instead of `--json` do `--format=json`

as if it comes to `asyncapi diff --files <file1.yml> <file2.yml>`, we have some discussions in other issue that are kind of going the direction of getting rid of `--file` or `--files` flag, as flags are options, they are optional, while in case of file, it is must have, not optional, it is kinda a center of all the things

aayushmau5:
[...]

PR #135, opened 2021-11-12
feat: integrate `asyncapi diff` library using the `diff` command

aayushmau5:
**Description**

This PR integrates the [`asyncapi-dff`](https://github.com/asyncapi/diff/) library using the `diff` command.

At the moment, the command looks something like this.

```
asyncapi diff <file-path/context-1> <file-path/context-2>
```

And it has two flags:

`--format`(default and only acceptable right now value: `json`) which determines the format of the output.
`--type`(`all` (default) , `breaking`, `non-breaking`, `unclassified`) which determines the type of the output we want.

At the moment, the output is being logged to `stdout`. The support for outputting data into files is to be added(which is most likely to be done by me, whether in this PR or another one depends on the review :) ).

**Related issue(s)**

See https://github.com/asyncapi/cli/issues/58


src/commands/diff.ts
```diff
@@ -0,0 +1,109 @@
...
+    const firstDocumentPath = args['first-spec-file'];
+    const secondDocumentPath = args['second-spec-file'];
+
+    const outputFormat = flags['format'];
+    const outputType = flags['type'];
+
+    let firstDocument, secondDocument;
+
+    try {
+      firstDocument = await load(firstDocumentPath);
```
Souvikns, line 50: awesome 👏🏻 
As you are loading spec files using the `load` function it does support loading spec files from context. so after setting appropriate contexts the command looks something like this 
```
node bin/run diff v1 v2  
```

but I think auto-loading context files does not work correctly, also I think we don't need autoloading contexts for this command anyway.

Souvikns (dismissed):

src/commands/diff.ts
```diff
@@ -0,0 +1,109 @@
...
+  static flags = {
+    help: flags.help({ char: 'h' }),
+    format: flags.string({
+      char: 'o',
+      description: 'output format',
+      default: 'json',
+    }),
+    type: flags.string({
+      char: 't',
+      description: 'the type of output',
```
derberg, line 21: what types are out there? should be clear in help what user can provide

src/commands/diff.ts
```diff
@@ -0,0 +1,109 @@
...
+import { ValidationError } from '../errors/validation-error';
+import { SpecificationFileNotFound } from '../errors/specification-file';
+
+export default class Diff extends Command {
+  static description = 'find diff between two asyncapi files';
+
+  static flags = {
+    help: flags.help({ char: 'h' }),
+    format: flags.string({
+      char: 'o',
```
derberg, line 15: sounds strange to me that main command is `format` and short version is `o` and not `f`

src/commands/diff.ts
```diff
@@ -0,0 +1,109 @@
...
+      description: 'output format',
```
derberg, line 16: what other formats are supported? tried yaml and failed

src/commands/diff.ts
```diff
@@ -0,0 +1,109 @@
...
+      char: 't',
+      description: 'the type of output',
+      default: 'all',
+    }),
+  };
+
+  static args = [
+    {
+      name: 'first-spec-file',
+      description: 'first spec path or context-name',
```
derberg, line 29: I think help should be clear on explaining which file is checked against which. I mean for me it sounds like 1st is new and 2nd is old, but help should be clear about it with no doubts

src/commands/diff.ts
derberg: what do you mean by autoloading context?

src/commands/diff.ts
```diff
@@ -0,0 +1,109 @@
...
+      );
+      const secondDocumentParsed = await parser.parse(
+        await secondDocument.read()
+      );
+      const diffOutput = diff.diff(
+        firstDocumentParsed.json(),
+        secondDocumentParsed.json()
+      );
+
+      if (outputFormat === 'json') {
```
derberg, line 91: please have a look at this article https://thecodebarbarian.com/pretty-json-stringify-output.html

we should lot out json in a pretty formatted way and not block of text, it would be much nicer :)

src/commands/diff.ts
derberg: maybe `FIRST-SPEC-FILE` and `SECOND-SPEC-FILE` should be called differently? check maybe other diff tolls on how they make it clear

derberg (changes requested):
Awesome to see it comming!
Please have a look at some possible improvements to developer experience

src/commands/diff.ts
Souvikns: By autoloading context I mean by the feature when there is no context set but there is an asyncapi file in the working directory cli loads that right, but in this case where more than two spec file paths are required so that would not work as intended.

derberg: ah, yeah, you mean fallback to `current` context. Yes, not a case here

src/commands/diff.ts
derberg: also there is no help message when I use other formats, like `not supported, this is the list we support atm: bla bla`

aayushmau5:
Thanks for the reviews. Will make the changes ASAP.

src/commands/diff.ts
aayushmau5: oops, yeah it should be `f`. Will fix that.

src/commands/diff.ts
aayushmau5: got it. 
Also, this made me wonder, right now only JSON output is supported, but we have plan for adding HTML/YAML/etc. types of output. 
Means we need to convert the JSON data to those formats. I was wondering whether we should have this "convert" function in diff library itself or we implement those standalone in the CLI?

src/commands/diff.ts
aayushmau5: Checked open api diff CLI, and they are using `openapidiff <old> <new>`. I will checkout other formats as well.

src/commands/diff.ts
derberg: imho this is for diff library to have this functionality of different outputs

src/commands/diff.ts
derberg: `<old> <new>` sounds pretty straightforward

aayushmau5: Cool. I will go with `<old> <new>` then. Looked at a couple of more libraries, and they had the similar format, so that works.

src/commands/diff.ts
aayushmau5: Got it. I will start working on different output formats after this PR.

derberg:
@aayushmau5 hey, tests are failing because of linter.

Souvikns:
This is something new 

image

aayushmau5:
Yeah, confused as to why this is happening. Let me check if this has something to do with the context I'm setting in the tests.

derberg:
@aayushmau5 you need a hand? or two? 😄

aayushmau5:
Yeah :sweat_smile: Not sure why the test fails in windows even though I have not used any context in the tests for diff command :confused:

derberg:
looking at logs `Error: EPERM: operation not permitted, open 'C:\Users\runneradmin\.test.asyncapi'` looks like tests fail as there is no access to the config file that keeps context, and test that removes it, cannot do it because it has no rights to do it.

super strange that it is happening right now, they had to do some updates to VM or something.

The problem is I have no idea how to solve it, other than adding a NODE_ENV that you can use in tests to override default context location. This way you could provide location that can be easily modified by the workflow. 

@Souvikns @boyney123 @magicmatatjahu thoughts?

to summarize, I'm thinking about modifying these 3 lines -> https://github.com/asyncapi/cli/blob/master/src/models/Context.ts#L8-L10
```
const isTestEnv = !!process.env.TEST;
const CONTEXT_FILENAME = isTestEnv ? '.test.asyncapi' : '.asyncapi';
export const DEFAULT_CONTEXT_FILE_PATH = path.resolve(os.homedir(), CONTEXT_FILENAME);
```

with something like 
```
const CONTEXT_FILENAME = process.env.CONTEXT_FILENAME || '.asyncapi';
export const DEFAULT_CONTEXT_FILE_PATH = process.env.CONTEXT_FILE_PATH || path.resolve(os.homedir(), CONTEXT_FILENAME);
```

and then just update `test` script in `package.json` to provide custom name and path location

Souvikns:
```
const isTestEnv = !!process.env.TEST;
const CONTEXT_FILENAME = isTestEnv ? '.test.asyncapi' : '.asyncapi';
export const DEFAULT_CONTEXT_FILE_PATH = path.resolve(os.homedir(), CONTEXT_FILENAME);
```

we have this so that while testing we don't overwrite the global config file. I think we should just check if it is test env then we should choose `.test.asyncapi` or choose something from the env variable. 

```
const CONTEXT_FILENAME = isTestEnv ? process.env.context_file_path || '.test.asyncapi' : '.asyncapi'
```

derberg:
@Souvikns the thing is that we should have no test-related logic in library, this is why I propose that we have all customizable and we just override it in the `test` script, and we create context config in `test` dir during testing, and make sure the file is in gitignore

Souvikns:
> @Souvikns the thing is that we should have no test-related logic in library, this is why I propose that we have all customizable and we just override it in the `test` script, and we create context config in `test` dir during testing, and make sure the file is in gitignore

Oh now I get it, yeah that would be better

aayushmau5:
hmm :confused: , so the problem still persists. I guess we can try [this](https://github.com/asyncapi/cli/pull/135#issuecomment-998770505). @derberg Should I make the changes in this PR or make a new one?

derberg:
as this is super small change, you can do it in this PR. Cleaner would of course be separate.

aayushmau5:
So, I'm putting `CONTEXT_FILENAME` to `.test.asyncpi`, what should I put into `CONTEXT_FILE_PATH`?

aayushmau5:
so, the problem still remains :confused:

aayushmau5:
I wonder if this bug is occurring due to the tests I added in this PR? Since the other PRs doesn't seem to have this problem.

aayushmau5:
> I wonder if this bug is occurring due to the tests I added in this PR? Since the other PRs doesn't seem to have this problem.

commented out the test, and the test for windows still fails.

Souvikns:
@aayushmau5 the tests are working now, check once.

aayushmau5:
oh, neat! Looks like triggering the workflow again solves the problem.

Souvikns (approved):
Looks good to me 🚀

src/commands/diff.ts
```diff
@@ -0,0 +1,118 @@
...
+      description: 'type of the output',
+      default: 'all',
+      options: ['breaking', 'non-breaking', 'unclassified', 'all'],
+    }),
+  };
+
+  static args = [
+    {
+      name: 'old',
+      description: 'old spec path or context-name',
```
derberg, line 32:
```suggestion
      description: 'old spec path, URL or context-name',
```

src/commands/diff.ts
```diff
@@ -0,0 +1,118 @@
...
+      required: true,
+    },
+    {
+      name: 'new',
+      description: 'new spec path or context-name',
```
derberg, line 37:
```suggestion
      description: 'new spec path, URL or context-name',
```

derberg (commented):
some docs related improvements

from functionality point of view, we do not have option that enables user to provide a link to a custom standard. Can you add it?

aayushmau5:
@derberg since AsyncAPI spec loader already supports loading with URL, I should just make the description changes, right?

derberg:
also I don't think that you need a dedicated test of `url` really. You are testing generic CLI functionality and not the ones added by `diff` command. Isn't it kinda redundant?

aayushmau5:
Right! I will remove them.

derberg:
@aayushmau5 I closed and reopened hoping that this will force CI to run on a different worker. Rerun of job did not help 😞

derberg:
I opened https://github.com/actions/virtual-environments/issues/4902

aayushmau5:
Great! :crossed_fingers:

derberg:
@aayushmau5 there is a workaround proposed, please have a look https://github.com/actions/virtual-environments/issues/4902#issuecomment-1019845960 it kinda makes sense

aayushmau5:
The NodeJS doc says, "Existing hidden files can be opened for writing with the 'r+' flag."

But I don't think the `test.asyncpi` context file already exists. Adding the `r+` flag locally results in failed tests :(

derberg:
yeah, and in case of tests, context is actually not a hidden file but regular `test.asyncapi` file

derberg:
@aayushmau5 some more hints -> https://github.com/actions/virtual-environments/issues/4902#issuecomment-1020051944

test/testHelper.ts
```diff
@@ -26,7 +26,10 @@ export default class ContextTestingHelper {
   }
 
   createDummyContextFile(): void {
-    writeFileSync(DEFAULT_CONTEXT_FILE_PATH, JSON.stringify(this._context), { encoding: 'utf-8' });
+    writeFileSync(DEFAULT_CONTEXT_FILE_PATH, JSON.stringify(this._context), { encoding: 'utf-8', flag: 'w+' });
+    if (!existsSync(DEFAULT_CONTEXT_FILE_PATH)) {
+      throw new Error('Context file doesn\'t exist');
+    }
```
aayushmau5, line 32: @derberg I added this check, and it seems like the file exists. The action doesn't throw this error. Let me know if I should add the check anywhere else.

aayushmau5:
@derberg How about I close this PR, and make another one? Would that do anything :thinking:

derberg:
The problem is I saw the same on another PR :(

How did checks work, so are we sure the context file actually exists?

derberg:
@aayushmau5 how is it? you mind if I push some code into this PR on my own to play a bit? it is the only way i can test on windows :)

aayushmau5:
Sure! Go ahead.

derberg:
@aayushmau5 I think I fixed it. The problem was related to wrongly written tests. Like not tests themselves, just wrong usage of mocha. The thing is that the before and after hooks were in the wrong location, they need to be combined with `it`. So it was failing randomly as sometimes they run properly and sometimes hooks were not fired properly, and `write` to the disk was failing as the file was there already, not removed because the hook that was supposed to delete it was too late.

can you have a final look and cleanup stuff that are actually not needed, this access checking and special options passed to `writeFileSync`. Also please adjust your `diff` tests accordingly

aayushmau5:
Awesome! Thanks, @derberg 

> can you have a final look and cleanup stuff that are actually not needed, this access checking and special options passed to writeFileSync. Also please adjust your diff tests accordingly

Yeah! I will refactor some code.

scripts/fetch-asyncapi-example.js
```diff
@@ -6,7 +6,7 @@ const unzipper = require('unzipper');
 const path = require('path');
 const parser = require('@asyncapi/parser');
 
-const SPEC_EXAMPLES_ZIP_URL = 'https://github.com/asyncapi/spec/archive/refs/heads/master.zip';
+const SPEC_EXAMPLES_ZIP_URL = 'https://github.com/asyncapi/spec/archive/refs/tags/v2.2.0.zip';
```
derberg, line 9:
```suggestion
const SPEC_EXAMPLES_ZIP_URL = 'https://github.com/asyncapi/spec/archive/refs/tags/v2.3.0.zip';
```

we just released 2.3 yesterday, and long term solution for this hardcoded link will be solved with https://github.com/asyncapi/cli/issues/205

test/commands/diff.test.ts
```diff
@@ -0,0 +1,114 @@
+/* eslint-disable sonarjs/no-duplicate-string */
+import { expect, test } from '@oclif/test';
+
+describe('diff', () => {
+  describe('with file paths, and there are no difference between the files', () => {
```
derberg, line 5: test should be inside `it`

why not changing this:
```
  describe('with file paths, and there are no difference between the files', () => {
    test
      .stderr()
      .stdout()
      .command(['diff', './test/specification.yml', './test/specification.yml'])
      .it('works when file path is passed', (ctx, done) => {
        expect(JSON.stringify(ctx.stdout)).to.equal(
          '"{\\n  \\"changes\\": []\\n}\\n"'
        );
        expect(ctx.stderr).to.equal('');
        done();
      });
  });
```

into this

```
  it('with file paths, and there are no difference between the files', () => {
    test
      .stderr()
      .stdout()
      .command(['diff', './test/specification.yml', './test/specification.yml'])
      
      expect(JSON.stringify(ctx.stdout)).to.equal(
        '"{\\n  \\"changes\\": []\\n}\\n"'
      );
      expect(ctx.stderr).to.equal('');
      done();
      
  });
```

and same for the rest?

derberg (changes requested):
left just 2 comments

can you remind me as I do not remember, we fought with this one quite some time 😅 
what did we agree on with another flag that is needed, to enable user to pass a path to custom standard?

scripts/fetch-asyncapi-example.js
aayushmau5: I'm getting `Version 2.3.0 is not supported.` error at the moment.

derberg: ah, yes, the automated version bump failed, can you bump parser 1.14?

derberg:
@aayushmau5 your test still runs outside `it` function, have a look at the context tests and how I changed them

aayushmau5:
>  what did we agree on with another flag that is needed, to enable user to pass a path to custom standard?

Oops, I don't think we have discussed this :thinking: . I completely forgot about passing custom standard. 

What flag should we go with? Also, I'm assuming this file will be a json file(or we should treat it as such)?

derberg:
yeah, just enable what API of the diff library allows you to do. So definitely a json file, and I think `--overrides` is good, just provide good help description

src/commands/diff.ts
```diff
@@ -0,0 +1,139 @@
...
+    } else {
+      this.log(`The output type ${outputType} is not supported at the moment.`);
+    }
+  }
+}
+
+async function readOverrideFile(path: string): Promise<diff.OverrideObject> {
+  const overrideStringData = await readFile(path, { encoding: 'utf8' });
+  return JSON.parse(overrideStringData);
+}
```
aayushmau5, line 139: Not sure where to put this function.

derberg: imho it is fine if it just stays here, just add js doc

most important is to have proper error handling, awaits should be wrapped in try/catch and we need test, what error goes out when user passed path to not existing file and what error goes out when user passed path to not json file and `JSON.parse` fails

aayushmau5:
@derberg It seems adding tests like

```
it(..., () => {
	it(..., () => {})
})
```

results in the tests passing no more whether it really fails or not. 

But 
```
describe(..., () => {
	it(..., () => {})
})
```

seems to work fine.

derberg:
@aayushmau5 yes, go ahead with 

```
describe(..., () => {
	it(..., () => {})
})
```

src/errors/diff-error.ts
```diff
@@ -0,0 +1,15 @@
...
+    this.name = 'DiffOverrideFileError';
+    this.message = 'Override file not found';
+  }
+}
+
+export class DiffOverrideJSONError extends Error {
+  constructor() {
+    super();
+    this.name = 'DiffOverrideJSONError';
+    this.message = 'Error converting override file to JSON';
```
derberg, line 13:
```suggestion
    this.message = 'Provided override file is not a valid JSON file';
```

test/commands/diff.test.ts
```diff
@@ -0,0 +1,171 @@
...
+      .command([
+        'diff',
+        './test/fixtures/specification_v1.yml',
+        './test/fixtures/specification_v2.yml',
+        '--overrides=./test/fixtures/invalid-overrides.json',
+      ])
+      .it((ctx, done) => {
+        expect(ctx.stdout).to.equal('');
+        expect(ctx.stderr).to.equal(
+          'DiffOverrideJSONError: Error converting override file to JSON\n'
```
derberg, line 166:
```suggestion
          'DiffOverrideJSONError: Provided override file is not a valid JSON file\n'
```

derberg:
@aayushmau5 lgtm 🚀 I guess I can merge, right?

derberg (approved):

aayushmau5:
Yes! 🚀

Issue #47, opened 2021-08-12
Integrate `cupid` in the CLI


arjungarg07:
We are going to integrate [app-relations-discovery](https://github.com/asyncapi/app-relations-discovery) functionalities into asyncapi cli which would allow users to access all the features of the library and get more work done from the terminal.

derberg:
@arjungarg07 you definitely need to provide a description mate 😆  

I know you mean `asyncapi relations` but others need to know more and also the capabilities we need to take into account

arjungarg07:
@derberg My proposition for commands usage:

```
asyncapi relations --files {FILES_PATH} --plantUml  // For plantUML classDiagram syntax
asyncapi relations --files {FILES_PATH} --mermaid  // For mermaid flowchart syntax
asyncapi relations --files {FILES_PATH} --reactFlow  // For reactFlow nodes syntax
```

derberg:
I think we can start with `relations` as we can refactor name easily. you might want to share your thoughts here https://github.com/asyncapi/cli/issues/37 . Just keep in mind we need to support `--files` and `--contexts` too. As for the types of outputs we need `--type` imho, so you have `asyncapi relations --files {FILES_PATH} --type plantUML`, and when you do `asyncapi relations --files {FILES_PATH}` you get default


Thoughts?

arjungarg07:
> Just keep in mind we need to support `--files` and `--contexts` too.

What role will the `--contexts` play here, I think `relations` command would be standalone and would not be affected by contexts 🤔 

> As for the types of outputs we need `--type` imho, so you have `asyncapi relations --files {FILES_PATH} --type plantUML`, and when you do `asyncapi relations --files {FILES_PATH}` you get default

Yeah, we can move forward with this 👍🏽

derberg:
context is like a file reference, so you do not need to remember the file path

so you can create a number of contexts, called file1, file2, file3 so you can do `asyncapi validate --context file1`, so you should also be able to do `asyncapi relations --contexts file1, file2, file3`

imagine you are working on this project https://github.com/amadeus4dev/amadeus-async-flight-status. You work with it every day, you have 2 different AsyncAPI files there, would you like to pass filepath every single time?

arjungarg07:
Got it! Super cool, so we have to support `--files` and `--contexts` too.

derberg:
[...]

Issue #129, opened 2021-11-03
Integrate modelina to generate models


jonaslagoni:
#### Reason/Context

Sometimes you just want to generate the data models for the payloads in your AsyncAPI file, therefore I would like to integrate Modelina with the CLI, so we can offer an easy way for users to generate them.

#### Description
As a start, I would like to integrate it with no customization, at least for now, so you can only select the output language and nothing more.

But which format of commands is best fitting?

I think something like the following should be fine: `asyncapi generate models --language=ts`

boyney123:
@jonaslagoni like the idea!

How will it know which file to use?

jonaslagoni:
Without knowing any depth of the implementation from a CLI side of things, I assumed you would use the context?

Or would it need to use `asyncapi generate models --file=./myfile.yaml --language=ts`? 🤔

jonaslagoni:
Also, one point to consider, since Modelina, don't only support AsyncAPI documents as inputs, I guess through the CLI it should be possible to provide other input types as well, or?🤔

boyney123:
> Also, one point to consider, since Modelina, don't only support AsyncAPI documents as inputs, I guess through the CLI it should be possible to provide other input types as well, or?

I'm not sure, where do the boundaries start and stop. Given it's the `asyncapi` cli, I think it should just handle the related inputs personally (asyncapi files) or at least to start?

jonaslagoni:
> I'm not sure, where do the boundaries start and stop. Given it's the asyncapi cli, I think it should just handle the related inputs personally (asyncapi files) or at least to start?

Yea, I agree. Lets just focus on AsyncAPI files and see where it goes 👍

derberg:
are you sure about `models` here `asyncapi generate models`? what about generator itself, how would it be called? `asyncapi generate template`? -> I'm afraid user experience depends on how underlying libraries are structured

shouldn't it be `asyncapi generate filepath/contextname model/template --parameters/--options` ?

derberg:
except of `@asyncapi/template` is just confusing 🤔 as it is not a command :( it is an argument that might be different depending on what template you want to use. Also template cannot be a flag as options should not be mandatory 🤔 


Have a look at https://github.com/asyncapi/cli/pull/143 and another use case for `asyncapi generate`

derberg:
how about (especially looking at the PR mentioned above by @fmvilas that relates to @arjungarg07 work):
```
asyncapi generate diagram --type='plantuml'
[...]

PR #143, opened 2021-11-22
feat: integrate `cupid` in cli using the `relation` command

arjungarg07:
**Description**

This PR integrates @asyncapi/cupid in cli using the `relation` command.

Command usage syntax:
`asyncapi relation <context-1/filepath1> <context-2/filepath2> <context-3/filepath3> ...`

It has one flag:
`--type` which determines the output syntax of the relations and if not provided, would output default output syntax.

**Related issue(s)**
#47


fmvilas:
Great stuff, @arjungarg07! My only concern is that the command is not a verb. We decided to follow the form `asyncapi VERB [NOUN]` and `relation` is not a verb. What about `visualize`?

```
asyncapi visualize context1/asyncapi1 context2/asyncapi2 ...
```

arjungarg07:
Yeah @fmvilas , I did thought of using the verb `visualize` but I was skeptical about it since we are not actually visualizing the architecture right away but just getting the relations in desired output format and later pasting it in live editor to visualize.

fmvilas:
🤔 I see what you mean. Maybe it should be something like `asyncapi generate diagram`? It kinda collapses with `asyncapi generate` where we probably want to use Generator. But maybe we can use `asyncapi generate code` and `asyncapi generate docs` for those...

For sure it can't be `relation` as it's not a verb/command, just a noun. Any other ideas?

arjungarg07:
What about `asyncapi generate relations` ?

fmvilas:
🤔 It might sound as you're generating the relations but the relations are already there, you're just "finding" them and generating a diagram, isn't it?

arjungarg07:
I didn't think in that way, in that case `asyncapi generate diagram` would be good.

fmvilas:
Yeah, now what worries me is that clash with `asyncapi generate` command. There's a related conversation going on in #129.

felixjung:
What about `asyncapi relate`?

arjungarg07:
> What about `asyncapi relate`?

Yeah, this could be done.
@derberg what do you think about it?

fmvilas:
Yeah, I still think it's better if we stick to `generate diagram` or similar. Always look at it from the point of view of the user who doesn't know the underlying technology and instead they expect the command to represent what they want to achieve. The tool used for this is irrelevant for them.

derberg:
@arjungarg07 hey is this one ready for review?

- can you remove 2 code smells?
- can you remove manual bump of the cli version from package.json and package-lock.json, as we do it by the bot, and it will constantly cause issues for you on a PR, conflicts you do not want to have

derberg:
@arjungarg07 hey, not sure if you noticed but `asyncapi generate` is now possible, so you can continue with this PR if you want

arjungarg07:
@derberg what would be the final command look like? `asyncapi generate relation` ?

derberg:
@arjungarg07 yo, should we keep it open?

arjungarg07:
@derberg I had some doubts, to be resolved(asked above in the thread). Maybe, would open a new PR.

derberg:
up to you 😃

derberg:
> to be resolved(asked above in the thread)

oh wait, sorry, which doubts are not yet solved?

derberg:
@arjungarg07 I'm guessing we should close it, right?

chinma-yyy:
> @arjungarg07 I'm guessing we should close it, right?

So does it mean that this issue is open to other contributors @derberg ??

test/commands/relation.test.ts
Shurtu-gal, line 1: @arjungarg07 there has been a recent change in the folder structure for tests in #712. This can be moved into https://github.com/asyncapi/cli/tree/master/test/integration.

Shurtu-gal (changes requested):
Also, the example folder can be moved into https://github.com/asyncapi/cli/tree/master/test/fixtures

derberg:
closing for now, I'm in touch with @arjungarg07 to start a process of looking for new maintainers for `cupid`

PR #157, opened 2021-12-07
test: increased async test timeout to 10s from 5s

Souvikns:
**Description**
The tests are failing sometimes, we don't know all the reasons the tests might be failing but recently we figured that some tests are failing because they are taking more time than the async timeout that was set to 5s. This PR aims to resolve these issues, currently, all the tests in the development server are passing so no way to understand which tests cases are taking more time than intended. 

It was running fine in ubuntu but was having issues in mac and windows.

### DIFF
--- a/package.json
+++ b/package.json
@@ -133,7 +133,7 @@
     "pretest:coverage": "npm run build",
     "release": "semantic-release",
     "start": "npm run build && node dist/cli.js",
-    "test": "cross-env TEST=1 nyc --extension .ts mocha --require ts-node/register --reporter spec --timeout 6000 \"test/**/*.test.ts\"",
+    "test": "cross-env TEST=1 nyc --extension .ts mocha --require ts-node/register --reporter spec --timeout 10000 \"test/**/*.test.ts\"",
     "version": "oclif-dev readme && git add README.md",
     "readme": "oclif readme"
   },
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -133,7 +133,7 @@
     "pretest:coverage": "npm run build",
     "release": "semantic-release",
     "start": "npm run build && node dist/cli.js",
-    "test": "cross-env TEST=1 nyc --extension .ts mocha --require ts-node/register --reporter spec --timeout 5000 \"test/**/*.test.ts\"",
+    "test": "cross-env TEST=1 nyc --extension .ts mocha --require ts-node/register --reporter spec --timeout 6000 \"test/**/*.test.ts\"",
```
magicmatatjahu, line 136:
```suggestion
    "test": "cross-env TEST=1 nyc --extension .ts mocha --require ts-node/register --reporter spec --timeout 10000 \"test/**/*.test.ts\"",
```

If you wanna increase timeout, you shouldn't increase it by 1s, but as doubled previous value, so 5s -> 10s. If it's really problem with timeout, 1s probably won't be a fix, better to have some "backup" of timeout :)

Souvikns: 😄 sure let's have some backup timeout.

magicmatatjahu (approved):
LGTM, but please wait for @derberg review :)

derberg (approved):

derberg:
I think conventional commits have prefix for test but don't remember if it is `test` or tests`

Souvikns:
Yeah it is `test` shall I update the pt title ?

derberg:
@Souvikns yes please, or just update it in the commit message when you will squash and merge

PR #173, opened 2022-01-05
feat: integrate `modelina` to generate models

### LINKED ISSUE #129: Integrate modelina to generate models

jonaslagoni:
**Description**

- ...
- ...
- ...

**Related issue(s)**
fixes #129

### DIFF
--- a/src/commands/types.ts
+++ b/src/commands/types.ts
@@ -1,0 +1,78 @@
+import { CSharpFileGenerator, JavaFileGenerator, JavaScriptFileGenerator } from '@asyncapi/modelina';
+import { flags } from '@oclif/command';
+import Command from '../base';
+import { load } from '../models/SpecificationFile';
+import { parse } from '@asyncapi/parser';
+export default class GenerateTypes extends Command {
+  static description = 'Generates typed models';
+
+  static flags = {
+    help: flags.help({ char: 'h' }),
+    file: flags.string({ char: 'f', description: 'path to the AsyncAPI file to generate types for' }),
+    outputDirectory: flags.string({ char: 'o', description: 'output path to write the types to', required: true}),
+    language: flags.enum({ char: 'l', description: 'language output', options: ['typescript', 'csharp', 'golang', 'java', 'javascript'], required: true}),
+    /**
+     * Go specific options
+     */
+    goPackageName: flags.string({description: 'Go specific, define the package to use for the generated types', required: false}),
+    /**
+     * Java specific options
+     */
+    javaPackageName: flags.string({description: 'Java specific, define the package to use for the generated types', required: false}),
+    /**
+     * C# specific options
+     */
+    csharpNamespace: flags.string({description: 'C# specific, define the namespace to use for the generated types', required: false}),
+  }
+
+  static args = []
+
+  async run() {
+    const { flags } = this.parse(GenerateTypes);
+    const outputDirectory = flags.outputDirectory;
+    const file = await load(flags.file) || await load();
+    const parsedInput = await parse(await file.read());
+    const language = flags.language;
+    let fileGenerator;
+    let fileOptions = {};
+    if (language === 'typescript') {
+      //fileGenerator = new TypeScriptFileGenerator();
+    } else if (language === 'javascript') {
+      fileGenerator = new JavaScriptFileGenerator();
+    } else if (language === 'csharp') {
+      if (flags.csharpNamespace === undefined) {
+        throw new Error('Missing namespace option. Add `--csharpNamespace NAMESPACE` to set the desired namespace.');
+      }
+      fileGenerator = new CSharpFileGenerator();
+      const namespace = flags.csharpNamespace;
+      fileOptions = {
+        namespace
+      };
+    } else if (language === 'golang') {
+      if (flags.goPackageName === undefined) {
+        throw new Error('Missing package name option. Add `--goPackageName PACKAGENAME` to set the desired package name.');
+      }
+      const packageName = flags.goPackageName;
+      //fileGenerator = new GoFileGenerator();
+      fileOptions = {
+        packageName
+      };
+    } else if (language === 'java') {
+      if (flags.javaPackageName === undefined) {
+        throw new Error('Missing package name option. Add `--javaPackageName PACKAGENAME` to set the desired package name.');
+      }
+      fileGenerator = new JavaFileGenerator();
+      const packageName = flags.javaPackageName;
+      fileOptions = {
+        packageName
+      };
+    }
+    if (fileGenerator === undefined) {
+      throw new Error(`Could not determine generator for language ${language}`);
+    }
+    await fileGenerator.generateToFiles(
+      parsedInput as any, 
+      outputDirectory, 
+      fileOptions as any);
+  }
+}
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,6 +9,7 @@
       "version": "0.11.1",
       "license": "Apache-2.0",
       "dependencies": {
+        "@asyncapi/modelina": "^0.41.0",
         "@asyncapi/parser": "^1.12.0",
         "@asyncapi/studio": "^0.5.0",
         "@fmvilas/oclif-plugin-spaced-commands": "^1.0.4",
@@ -97,6 +98,35 @@
       },
       "bin": {
         "js-yaml": "bin/js-yaml.js"
+      }
+    },
+    "node_modules/@apidevtools/openapi-schemas": {
+      "version": "2.1.0",
+      "resolved": "https://registry.npmjs.org/@apidevtools/openapi-schemas/-/openapi-schemas-2.1.0.tgz",
+      "integrity": "sha512-Zc1AlqrJlX3SlpupFGpiLi2EbteyP7fXmUOGup6/DnkRgjP9bgMM/ag+n91rsv0U1Gpz0H3VILA/o3bW7Ua6BQ==",
+      "engines": {
+        "node": ">=10"
+      }
+    },
+    "node_modules/@apidevtools/swagger-methods": {
+      "version": "3.0.2",
+      "resolved": "https://registry.npmjs.org/@apidevtools/swagger-methods/-/swagger-methods-3.0.2.tgz",
+      "integrity": "sha512-QAkD5kK2b1WfjDS/UQn/qQkbwF31uqRjPTrsCs5ZG9BQGAkjwvqGFjjPqAuzac/IYzpPtRzjCP1WrTuAIjMrXg=="
+    },
+    "node_modules/@apidevtools/swagger-parser": {
+      "version": "10.0.3",
+      "resolved": "https://registry.npmjs.org/@apidevtools/swagger-parser/-/swagger-parser-10.0.3.tgz",
+      "integrity": "sha512-sNiLY51vZOmSPFZA5TF35KZ2HbgYklQnTSDnkghamzLb3EkNtcQnrBQEj5AOCxHpTtXpqMCRM1CrmV2rG6nw4g==",
+      "dependencies": {
+        "@apidevtools/json-schema-ref-parser": "^9.0.6",
+        "@apidevtools/openapi-schemas": "^2.0.4",
+        "@apidevtools/swagger-methods": "^3.0.2",
+        "@jsdevtools/ono": "^7.1.3",
+        "call-me-maybe": "^1.0.1",
+        "z-schema": "^5.0.1"
+      },
+      "peerDependencies": {
+        "openapi-types": ">=7"
       }
     },
     "node_modules/@asyncapi/avro-schema-parser": {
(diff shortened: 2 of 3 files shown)


src/commands/types.ts
```diff
@@ -0,0 +1,78 @@
...
+
+  async run() {
+    const { flags } = this.parse(GenerateTypes);
+    const outputDirectory = flags.outputDirectory;
+    const file = await load(flags.file) || await load();
+    const parsedInput = await parse(await file.read());
+    const language = flags.language;
+    let fileGenerator;
+    let fileOptions = {};
+    if (language === 'typescript') {
```
magicmatatjahu, line 38: I know that this is a draft PR, but maybe you should use `switch case` with separate function for every language? That small suggestion 😄

jonaslagoni: Changed 🙂

magicmatatjahu: Where? 😄

jonaslagoni: In my local commit-tree 😆

jonaslagoni: Finishing the PR before pushing the changes.

jonaslagoni:
PR is ready to be reviewed 🙂

src/commands/generate/types.ts
```diff
@@ -0,0 +1,88 @@
...
+import Command from '../../base';
+import { load } from '../../models/SpecificationFile';
+import { parse } from '@asyncapi/parser';
+export default class Types extends Command {
+  static description = 'Generates typed models';
+
+  static flags = {
+    help: flags.help({ char: 'h' }),
+    file: flags.string({ char: 'f', description: 'path to the AsyncAPI file to generate types for' }),
+    outputDirectory: flags.string({ char: 'o', description: 'output path to write the types to', required: true}),
```
derberg, line 12: why not `output` as it is already in generator cli? nobody complained on it

derberg:
@magicmatatjahu @fmvilas @Souvikns hey folks, can you remind me as I can find it anywhere but my brain is telling me that we had a discussion about this principle that `something should be a flag only if it can be optional and if something cannot be optional, and no default is possible then it should be part of the command`. Referring to `language` flag

@Souvikns I remember that at some point in time we were discussing that commands should be stored in separate files so they are then reused in source and tests and not duplicated. Was it done and lost during refactoring or? can't remembers, sorry :(

Souvikns:
@derberg is https://github.com/asyncapi/cli/pull/59#issuecomment-915152405 what you are looking for

derberg:
@Souvikns not, not this one, this is about helping out if you do `asyncapi generte` and error ask if you meant `asyncapi generate`

fmvilas:
Yeah I can't find where we discussed that either. I think --file should not be a flag but an argument instead. Something like `asyncapi generate types <asyncapi-file>`. Same applies to language, unless we want to provide a default value.

magicmatatjahu:
Yeah I would also be in favor of making the asyncapi file and language optional, so the call itself should look like this:

```
asyncapi generate types ./asyncapi.json typescript -o ./output
```

I also wonder if that long `goPackageName`, `javaPackageName` or `csharpNamespace` flags are needed. I know this is because of the different types of modules in the languages, but maybe `packageName` and `namespace` is enough? Another languages have `namespace` or module (packages) system so maybe we should go only with that two flags and reuse them with appropriate language 🤔

derberg:
yeah, actually should not there be support for `file/context/fallback` commands anyway?

src/commands/generate/types.ts
```diff
@@ -0,0 +1,88 @@
...
+      warn: this.warn,
+      error: this.error,
+    });
+    const models = await fileGenerator.generateToFiles(
+      parsedInput as any, 
+      outputDirectory, 
+      {...fileOptions, } as any);
+    const generatedModelNames = models.map((model) => {return model.modelName;});
+
+    this.log(`We successfully generated the following models ${generatedModelNames.join(', ')}`);
```
boyney123, line 86: Very minor one, but in this context who is `We?`, maybe just

`this.log(`Successfully generated the following models ${generatedModelNames.join(', ')}`);`?

src/commands/generate/types.ts
jonaslagoni: That makes sense 👍

magicmatatjahu:
I don't want this to be too late, but haven't you thought about being compatible with the cli generator, which we will move to this repository? E.g. the parameters for template are given by flags `-p` giving the name and value. Maybe we could go the same way for modelina, so "shape" of `generate types` command should looks like:

```bash
asyncapi generate types <asyncapi-file> <language> -p <name=value>
```

We can also add JSON Schemas (in the modelina repo) for available parameters to the given language and we can validate these parameters by AJV. Some benefits by this solution:
- we will update available options in the modelina repo, and after release we will have that new options in the cli without having to contribute here. We will also avoid unnecessary flags in that command, everything will be inside modelina repo and due to fact that we have automatic bumps dependant packages, we will always up to date with modelin in cli :)
- we can also add available languages list in the modelina so we can also avoid contributions to the cli when we will add new language

What do you think? @jonaslagoni

src/commands/generate/types.ts
jonaslagoni: Changed 👍

jonaslagoni:
There is a large discrepancy between how files are provided to commands, some use `fileName` others `file`. I leave it up to you to create that consistency and remain with the file as a flag to allow context switching.

Finished adapting all the suggestions.

> I also wonder if that long `goPackageName`, `javaPackageName` or `csharpNamespace` flags are needed. I know this is because of the different types of modules in the languages, but maybe `packageName` and `namespace` is enough? Another languages have `namespace` or module (packages) system so maybe we should go only with that two flags and reuse them with appropriate language 🤔

I like it, changed 👍 

> I also wonder if we should make as followup flag to load JS file to load config for given Langauge Generator? As you know user has possibility to change format of model name etc. and of course that config we cannot convert to the CLI flags/parameters. WDYT?

Another feature, that Modelina natively should support at some point, agree (similar to tsconfig.json) 👍

magicmatatjahu:
It's a bit of a shame with these hardcoded things, is there really nothing we can do? Very soon CLI will be not up-to-date with the `modelina`. I know about that problem with JSON Schema and `$id` fields but can we do it with the draft-07 approach?

jonaslagoni:
> It's a bit of a shame with these hardcoded things, is there really nothing we can do? Very soon CLI will be not up-to-date with the `modelina`. I know about that problem with JSON Schema and `$id` fields but can we do it with the draft-07 approach?

This is not something I will rush 😄 We have a discussion here https://github.com/asyncapi/community/issues/249 and in Modelina and we will see what the outcome will be.

If codeowners of CLI do not want this to be merged until that is resolved, then that is perfectly fine. I can close down the PR for now, just let me know 🙂

magicmatatjahu:
> If you do not want this to be merged until that is resolved, then that is perfectly fine. I can close down the PR for now, just let me know 🙂

We can push it further, but we have to remember that there will be a big breaking change in the future 😅

jonaslagoni:
> We can push it further, but we have to remember that there will be a big breaking change in the future (of course when we will decide to go with [asyncapi/community#249](https://github.com/asyncapi/community/issues/249) proposal) 😅

You think it will? Dont you think it would be possible to structure the JSON schema documents in a way to make it a 'refactor` rather than a major version change?

fmvilas:
My two cents: I'd move forward with this. A breaking change in the CLI is not really a big problem and the other proposal is not really mature so I won't make users wait for something that may or may not happen.

jonaslagoni:
Pr has been updated and is ready to be reviewed again.

Souvikns (dismissed):
👍🏼👍🏼

jonaslagoni:
Anything needed to move this forward?

README.md
```diff
@@ -66,14 +64,19 @@ USAGE
...
 COMMANDS
   config    access configs
-  diff      find diff between two asyncapi files
-  new       creates a new asyncapi file
+  diff      find diff between two AsyncAPI files
+  new       creates a new AsyncAPI file
   start     starts a new local instance of Studio
-  validate  validate asyncapi file
+  validate  validate an AsyncAPI file
+  generates         generate all kinds of stuff
```
derberg, line 71:
```suggestion
  generate         generate all kinds of stuff
```

src/commands/generate/types.ts
```diff
@@ -0,0 +1,86 @@
...
+  static description = 'Generates typed models';
+
+  static args = [
+    { name: 'language', description: 'language output', options: ['typescript', 'csharp', 'golang', 'java', 'javascript'], required: true},
+    { name: 'file', description: 'spec path, url, or context-name', required: true },
+  ]
+
+  static flags = {
+    help: Flags.help({ char: 'h' }),
+    output: Flags.string({ char: 'o', description: 'output path to write the types to', required: true}),
```
derberg, line 16: should not be required, if `--output` is not provided, then print result to terminal. In AsyncAPI CLI flags always are optional

derberg (changes requested):
I noticed 2 things only

src/commands/generate/types.ts
magicmatatjahu: But types generation in console (and also templates) can be a really big problem. I understand that flags are optional, but in this case output is very important. Additionally I don't like this solution in cli from generator where for output we have fallback in form of generating in given folder where we run cli - it's not a good UX solution for me.

derberg: what is your suggestion though? output as argument?

fmvilas: I agree that it shouldn't be required and default to printing on `stdout` (console). This way people can use the CLI to pipe the output to another tool:

```bash
asyncapi generate types | sed ... | etc.
```

magicmatatjahu: Yes, but generating types and templates is usually a dozen or so files, how do you want to handle it on stdout? Is it even possible to iterate by array?

fmvilas: Can't you print all of them? Just like `git diff` does:

```
### File: MyClass.java

class MyClass {
...
}

### File: MyOtherClass.java

class MyOtherClass {
...
}
```

magicmatatjahu: Just what value will it have for the user himself? I can understand if someone would operate on a single file and not on a collection of files. I'm afraid it will only cause problems.

magicmatatjahu: Another solution would be to generate (without `output` flag) things in a folder named `generated-{timestamp}` and in the output we would give the name of this folder. Of course that folder will be located in current working directory.

jonaslagoni: > I already explained the value. People can chain commands to manipulate the output. Useful for CI environments for instance, where you may want to replace some values from the output. But I understand your point. I'd not worry about these details in this PR but something to consider long-term. A CLI that can't be chained in the command line is a pain in the ass.

I agree, that we need to support chaining. 

I did not realize that it should be possible to output for stdout 🤔 I dont think that it's useful without it being documented how it is possible to chain it, or at least give some examples, otherwise we risk that it won't be possible at all. For example for the models, how do you split up the models and access the information needed to further your chaining? I.e. accessing the model names and the content of the model and iterating over them in shell scripts? 🤔

src/commands/generate/types.ts
```diff
@@ -0,0 +1,123 @@
...
+      description: 'language output', 
+      options: Object.keys(Languages), 
+      required: true 
+    },
+    { name: 'file', description: 'spec path, url, or context-name', required: true },
+  ]
+
+  static flags = {
+    help: Flags.help({ char: 'h' }),
+    output: Flags.string({ char: 'o', description: 'output path to write the model to' }),
```
fmvilas, line 29: If this is going to be a directory, should we call it `outputDir` instead? Or is this flag representing the name of a single file?

jonaslagoni: It could be `outputDir` yes, that would be more precise as it is the directory where all the models are going to be generated to 🤔

fmvilas: I think it's just a matter of what I found most common in my career. `output` usually refers to a file or stdout. When the output is a whole dir, it should be noted IMHO. I don't think many people will expect the output to be a dir, to be honest. Anyway, not a blocker for me, just my two cents.

jonaslagoni:
I went over all the descriptions and clarified them where I found them necessary 🙂

derberg (dismissed):
LGTM 🚀 

@Souvikns @magicmatatjahu 

Do you also approve? You were involved here so want to make sure you are fine too

/dnm

Souvikns (dismissed):
👍🏼

magicmatatjahu (approved):

magicmatatjahu (commented):
One comment: Should we change `generate types` to the `generate models`?

derberg:
@jonaslagoni some conflicts popped up after the last merge.

> One comment: Should we change generate types to the generate models?

I do not have an opinion, we can support both 😄

jonaslagoni:
> One comment: Should we change `generate types` to the `generate models`?

I like `generate models` better, but types are fine as well 🤷 So no strong opinion.

> @jonaslagoni some conflicts popped up after the last merge.

Fixed.

magicmatatjahu:
So `generate models` 😄

jonaslagoni:
> So `generate models` 😄

@Souvikns @derberg what do you think? Stay or switch?

derberg (commented):
`models` is fine 🤷🏼

jonaslagoni:
Alright, switched to models 👍

Souvikns (approved):

PR #174, opened 2022-01-10
fix: show correct error message when no context is saved

### LINKED ISSUE #148: not getting the right error when calling `asyncapi validate` or `asyncapi start studio` without context and without file in work dir
When I have no context and no asyncapi file in my work dir I get: `ContextError: No context is set as current, please set a current context` but should get:

```
  - You can provide a path to the AsyncAPI file: asyncapi validate path/to/file/asyncapi.yml
  - You can also pass a saved context that points to your AsyncAPI file: asyncapi validate mycontext
  - In case you did not specify a context that you want to use, the CLI checks if there is a default context and uses it. To set default context run: asyncapi context use mycontext
```

I think we need to pull https://github.com/asyncapi/cli/blob/master/src/errors/validation-error.ts#L19 to more generic error and also fix https://github.com/asyncapi/cli/blob/b9cc9388d0aac81caa8d5bd9322b4c430127cb53/src/models/SpecificationFile.ts#L43

more details https://github.com/asyncapi/cli/pull/130#issuecomment-976752976

Souvikns:
**Description**

If there is no input we then we need to read `.asyncapi` file to access the current context. But when there is no current context we need to tell the user how they can save context and use it. 

I updated the Error message to -
```
Specify what AsyncAPI file to be used.
 - You can provide a path to the AsyncAPI file: asyncapi <command> path/to/file/asyncapi
 - In case you did not specify a context that you want to use, the CLI checks if there is a default context and uses it. To set default context run: asyncapi context use mycontext
```

This message is not command specific as before, One issue with command-specific error messages is it needs to know which command is called. Now when we were using ink we were using a hook to figure out which command has been called. But right now when we are loading the `.asyncapi` file we throw an error when `.asyncapi` file is missing now this should not know which command has called it. 

**Related issue(s)**

Fixes #148


derberg:
@Souvikns you have 2 code smells there

src/errors/specification-file.ts
```diff
@@ -22,3 +22,37 @@ export class SpecificationURLNotFound extends SpecificationFileError {
     this.message = `Unable to fetch specification file from url: ${URL}`;
   }
 }
+
+type From = 'file' | 'url' | 'context'
+
+export class ErrorLoadingSpec extends Error {
+  private readonly errorMessages = {
+    default: `Specify what AsyncAPI file to be used.
```
derberg, line 30: can't we reuse `NO_CONTEXTS_SAVED` to not duplicate this?

test/commands/validate.test.ts
```diff
@@ -31,7 +31,7 @@ describe('validate', () => {
       .command(['validate', './test/not-found.yml'])
       .it('should throw error if file path is wrong', (ctx, done) => {
         expect(ctx.stdout).to.equals('');
-        expect(ctx.stderr).to.equals('ValidationError: There is no file or context with name "./test/not-found.yml".\n');
+        expect(ctx.stderr).to.equals('Error Loading Specification From File: ./test/not-found.yml is an invalid file path\n');
```
derberg, line 34: make sure we do not do always upper case in these error messages, it should be like regular sentence

```suggestion
        expect(ctx.stderr).to.equals('Error loading AsyncAPI document from file: ./test/not-found.yml is an invalid file path\n');
```

derberg:
@Souvikns have a look at it, looks like after you apply some fixes we can merge it

src/errors/specification-file.ts
```diff
@@ -22,3 +23,31 @@ export class SpecificationURLNotFound extends SpecificationFileError {
...
+type From = 'file' | 'url' | 'context'
+
+export class ErrorLoadingSpec extends Error {
+  private readonly errorMessages = {
+    default: NO_CONTEXTS_SAVED
+  }
+  constructor(from?: From, param?: string) {
+    super();
+    if (from === 'file') {
+      this.name = 'error loading specification from file';
```
derberg, line 36: instead of `loading specificaton` always write `loading AsyncAPI document` in all error messages

src/commands/validate.ts
```diff
@@ -3,36 +3,31 @@ import * as parser from '@asyncapi/parser';
...
 import { load } from '../models/SpecificationFile';
-import { SpecificationFileNotFound } from '../errors/specification-file';
 
 export default class Validate extends Command {
   static description = 'validate asyncapi file';
 
   static flags = {
-    help: flags.help({ char: 'h' })
+    help: flags.help({ char: 'h' }),
+    file: flags.boolean({ char: 'f', description: 'Input will be treated as a file path' }),
```
derberg, line 12: why do we have file, context and user as flags now?

derberg (changes requested):
some comments

src/commands/validate.ts
Souvikns: I was playing with something, so it was like `asyncapi validate filepath` and then we figure out if that is file or context or URL now I was just adding a flag that specifically tells what that input should be treated at so if you do `asyncapi validate <some-input> -f` then it will take the input as a file

Souvikns: I did this to help other commands which are using flags to take file input. Basically `load` function reads the specification from file, context or URL. So even if a command is using the `-f` flag to load the specification file it still does all the checks, which it shouldn't as we are telling it that it is a file. 

I thought that we can use this to solve the issue but I think this should not be in this PR. Actually, I forgot to remove it.

derberg: yeah, it just causes confusion, just remove them from this pr

derberg (approved):
🚀 :shipit:

Issue #41, opened 2021-07-29
Add `convert` command


magicmatatjahu:
Our [comverter-js](https://github.com/asyncapi/converter-js) has builded custom CLI, but we should move it to the main CLI. 

AC: 
- add new `convert` command to the CLI
- add tests for new command
- update docs (Readme.md) with new command
- add `--help` flag
- to discuss: should we remove `--id` flag from https://github.com/asyncapi/converter-js CLi? It's only needed between 1.X.X -> 2.X.X.

Usage (my proposition):

```bash
asyncapi convert --file {FILE} --to 2.1.0 // convert to the 2.1.0 version
asyncapi convert --file {FILE} --to latest // convert to the latest version of spec
asyncapi convert --file {FILE} // convert to the latest version of spec (omit the `--to` flag)
asyncapi convert --file {FILE} --output {OUTPUT_FILE} // convert and save in the `output`
```

Maybe we should treat `--to` as standalone argument?

```bash
asyncapi convert --file {FILE} 2.1.0 // convert to the 2.1.0 version
```

I have also problem with `--output` flag, because I don't know if we should follow with custom CLI for converter:

![image](https://user-images.githubusercontent.com/20404945/127479569-adac626c-f428-4eea-9109-c284fa29f9ab.png)

Any suggestions?

Also we should consider saving the converted spec to the existing context, it means that user will operate on some context and action like:

```bash
asyncapi convert --context {CONTEXT} --to 2.1.0
```

will save the output to the `{CONTEXT}`.

Some my questions about the future of the command:
- should we only focus on converting the AsyncAPI, or maybe extend it with converting OpenAPI to the AsyncAPI? and another specs 😏 
- should we also add option to convert the RAML, AVRO etc formats to our Schema Object format - similar to our custom parsers for formats, but as standalone command in our CLI?

@derberg @Souvikns What do you think?

Souvikns:
Maybe we have a `--output` flag that takes a path as an input. Converts the old version to the new version and outputs in that path, and if the `--output` flag is missing we can just update the original file.

derberg:
> Our comverter-js has builded custom CLI, but we should move it to the main CLI.

definitely, same with generator. This is the ultimate goal of this CLI. One CLI to rule them all 😄 

[...]

PR #178, opened 2022-01-12
fix: improve `validation-error.ts` file

peter-rr:
**Description**

- This fix removes the 'if' condition evaluating 'errorHasTitle' (line 42) and also removes that variable (line 37) since the 'title' property from the corresponding error must be mandatory as specified [here](https://github.com/asyncapi/parser-js/blob/master/API.md#asyncapiparserparsererror--error). Therefore the evaluation of that condition is not needed. In fact, the 'title' property is already used in the previous 'if' condition (line 40).

**Related issue(s)**
- Discovered while working on #41


derberg:
@peter-rr It is true we have `title` in `ParserError` under `validationErrors` but we do not really make sure it is used. For example look at https://github.com/asyncapi/parser-js/blob/master/lib/parser.js#L85. Can you check your change by running `asyncapi validate` against document that starts with `asyncapi: 1.2.0`? I think we still need it and just need to fix and not remove how we validate it

peter-rr:
@derberg I'm getting the following error when running `asyncapi validate` against a spec file with `asyncapi: 1.2.0`:
```
ValidationError: Version 1.2.0 is not supported.

undefined
```
Does it mean the `title` is not used as you mentioned on your comment? And, if that's the case, should we need to refactor this `validation-error.ts`file?

derberg:
yes, exactly, `undefined` goes through because of lack of validation.

- so we still need `const errorHasTitle = !!e.title;`
- and we just need to check `errorHasTitle` when building 
    ```
    errorsInfo.push(`${e.title} ${e.location.startLine}:${e.location.startColumn}`);
    ```
    we need one condition where `errorHasTitle` and `errorHasLocation` is true and build 
   ```
   `errorsInfo.push(`${e.title} ${e.location.startLine}:${e.location.startColumn}`);`
   ```
    but also in case only `errorHasLocation` is available, we need to construct info without title


I'm just thinking now, if we could add some test 🤔 might be that in future someone else gets into this file and tries to change things as you suggested 🤔 or maybe you could add just extensive comments to the code?

peter-rr:
Hey @derberg 
Based on your suggestions, what if we add the following conditions to cover all the cases you mentioned:

```
        if (errorHasTitle && errorHasLocation) {
          errorsInfo.push(`${e.title} ${e.location.startLine}:${e.location.startColumn}`);
        } 
        if (errorHasTitle && !errorHasLocation) {
          errorsInfo.push(`${e.title}`);
        }
        if (!errorHasTitle && errorHasLocation) {
          errorsInfo.push(`${e.location.startLine}:${e.location.startColumn}`);
        }
```
🤔 I think we'd be adding just one more condition to the current code and it seems self-explanatory to avoid adding extensive comments. What do you think?

derberg:
sounds good to me, the conditions

> self-explanatory

yeah, I suggest not going into the trap of `self-explanatory`. One small comment that explains why different conditions are needed for parser errors will not harm 😄

peter-rr:
> One small comment that explains why different conditions are needed for parser errors will not harm 😄

Yeah, you're right! Ok, then I'll add a brief comment above those conditions 👌

peter-rr:
@derberg Ready to be reviewed ^^

derberg (dismissed):
Thanks 🚀

derberg:
@peter-rr have a look at failing linter

peter-rr:
@derberg No more code smells now 💪

derberg:
@peter-rr sorry for being kinda picky, but the comment that you added describe what is already visible in the `if` statement. What I meant by simple comment was to add short comment that this conditions are done like this because validation errors come from another library, Parser JS that does not assure that always all the fields are there in the error, so there might be cases that even title is not provided. You know what I mean? so basically add a comment that explains why conditions are needed, and not what they are doing, makes sense?

peter-rr:
@derberg Sure, it does make sense! No worries, actually I appreciate so much this kind of corrections 🙏 They are really helpful to me 💪📚 I already pushed the correction suggested, so tell me if anything wrong or not clear enough yet :)

derberg (approved):
Well done 👏🏼

derberg:
@all-contributors please add @peter-rr for code

PR #187, opened 2022-01-18
feat: add release steps and instructions to use CLI on MacOS

### LINKED ISSUE #168: Enable users to use CLI without installing NodeJS
#### Reason/Context

AsyncAPI CLI is supposed to be a tool that exposes all asyncapi tools to the community and makes it easier for the community to work with AsyncAPI files. This means it must be easy to access for folks that are not NodeJS fans and we should not force users to install NodeJS. 

### Brew for MacOS
Get us in brew and then automate release of upgrades. I already opened a PR https://github.com/Homebrew/homebrew-core/pull/92115 and will drive it forward and then come up with GH workflow to automate release (upgrade PR) of new versions.

### Executable for Linux, Mac and Windows 

Start using https://github.com/vercel/pkg and enrich our release pipeline (super easy with semantic-releases, I can help) to produce an executable for the CLI + have quick instructions how one can download it and use it

### Chocolatey for Windows

Get us in chocolatey and then automate release of upgrades. We need windows users that can lead it.

derberg:
Still draft as I'm waiting for first brew formula release, and then I will add release steps here to get it updated automatically

Resolves #168

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,9 +23,22 @@
 
 <!-- toc -->
 
+- [Installation](#installation)
+  * [Using NPM and Node](#using-npm-and-node)
+  * [MacOS](#macos)
+    + [Using brew](#using-brew)
+    + [Using pkg](#using-pkg)
+- [Usage](#usage)
+- [Contributing](#contributing)
+  * [Set up development environment](#set-up-development-environment)
+  * [Command Structure and Patterns](#command-structure-and-patterns)
+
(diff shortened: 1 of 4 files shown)


Souvikns (approved):
LGTM 👍🏼

magicmatatjahu (approved):
I tested locally `.pkg` and it works 🚀

PR #188, opened 2022-01-20
Add `convert` command

### LINKED ISSUE #41: Add `convert` command

peter-rr:
**Description**

* Implementation of 'asyncapi convert' command.

**Related issue(s)**

* Resolves #41

### DIFF
--- a/src/commands/convert.ts
+++ b/src/commands/convert.ts
@@ -24,7 +24,7 @@ export default class Convert extends Command {
     const { args } = this.parse(Convert);
     const filePath = args['spec-file'];
     let specFile;
-    
+
     // LOAD FILE
     try {
       specFile = await load(filePath);
@@ -54,21 +54,25 @@ export default class Convert extends Command {
         err: error
       });
     }
-    
+
     // CONVERSION
     try {
-      /* if (specFile.getFilePath()) {
+      /* 
+      if (specFile.getFilePath()) {
         await parser.parse(specFile.text());
         this.log(`File ${specFile.getFilePath()} successfully validated!`);
       } else if (specFile.getFileURL()) {
         await parser.parse(specFile.text());
         this.log(`URL ${specFile.getFileURL()} successfully validated`);
-      } */
+        }
+      */
     } catch (error) {
-/*       throw new ValidationError({
+      /*
+      throw new ValidationError({
         type: 'parser-error',
         err: error
       });
- */ }
+      */
+    }
   }
 }
--- a/src/commands/convert.ts
+++ b/src/commands/convert.ts
@@ -4,13 +4,15 @@ import Command from '../base';
 import { ValidationError } from '../errors/validation-error';
 import { load } from '../models/SpecificationFile';
 import { SpecificationFileNotFound } from '../errors/specification-file';
+import { convert } from '@asyncapi/converter';
+import { Specification } from '../models/SpecificationFile';
 
 export default class Convert extends Command {
   static description = 'convert asyncapi documents older to newer versions';
 
   static flags = {
     help: flags.help({ char: 'h' }),
-    'file-name': flags.string({ char: 'n', description: 'name of the file to convert' }),
+    file: flags.string({ char: 'f', description: 'path to the file to convert' }),
     output: flags.string({ char: 'o', description: 'name of the file where the result is saved' }),
     'target-version': flags.string({ char: 't', description: 'asyncapi version to convert to' }),
     id: flags.string({ char: 'i', description: 'application id, if needed' })
@@ -24,10 +26,28 @@ export default class Convert extends Command {
     const { args } = this.parse(Convert);
     const filePath = args['spec-file'];
     let specFile;
+    let convertedFile;
 
-    // LOAD FILE
     try {
+      // LOAD FILE
       specFile = await load(filePath);
+
+      // VALIDATION
+      if (specFile.getFilePath()) {
+        await parser.parse(specFile.text());
+        this.log(`File ${specFile.getFilePath()} successfully validated!`);
+      } else if (specFile.getFileURL()) {
+        await parser.parse(specFile.text());
+        this.log(`URL ${specFile.getFileURL()} successfully validated!`);
+      }
+
+      // CONVERSION
+      convertedFile = await convert(specFile.text(), Convert.flags['target-version'], Convert.flags.id);
+      if (convertedFile.getFilePath()) {
+        this.log(`File ${convertedFile.getFilePath()} successfully converted!`);
+      } else if (convertedFile.getFileURL()) {
+        this.log(`URL ${convertedFile.getFileURL()} successfully converted!`);
+      }
     } catch (err) {
       if (err instanceof SpecificationFileNotFound) {
         this.error(new ValidationError({
@@ -37,42 +57,6 @@ export default class Convert extends Command {
       } else {
         this.error(err as Error);
       }
-    }
-
-    // VALIDATION
-    try {
-      if (specFile.getFilePath()) {
-        await parser.parse(specFile.text());
-        this.log(`File ${specFile.getFilePath()} successfully validated!`);
-      } else if (specFile.getFileURL()) {
-        await parser.parse(specFile.text());
-        this.log(`URL ${specFile.getFileURL()} successfully validated`);
-      }
-    } catch (error) {
-      throw new ValidationError({
-        type: 'parser-error',
-        err: error
-      });
-    }
-
-    // CONVERSION
-    try {
-      /* 
-      if (specFile.getFilePath()) {
-        await parser.parse(specFile.text());
-        this.log(`File ${specFile.getFilePath()} successfully validated!`);
-      } else if (specFile.getFileURL()) {
-        await parser.parse(specFile.text());
-        this.log(`URL ${specFile.getFileURL()} successfully validated`);
-        }
-      */
-    } catch (error) {
-      /*
-      throw new ValidationError({
-        type: 'parser-error',
-        err: error
-      });
-      */
     }
   }
 }
--- a/src/commands/convert.ts
+++ b/src/commands/convert.ts
@@ -1,0 +1,80 @@
+import { flags } from '@oclif/command';
+import * as parser from '@asyncapi/parser';
+import Command from '../base';
+import { ValidationError } from '../errors/validation-error';
+import { load } from '../models/SpecificationFile';
+import { SpecificationFileNotFound } from '../errors/specification-file';
+
+export default class Convert extends Command {
+  static description = 'convert asyncapi documents older to newer versions';
+
+  static flags = {
+    help: flags.help({ char: 'h' }),
+    'file-name': flags.string({ char: 'f', description: 'name of the file to convert' }),
+    output: flags.string({ char: 'o', description: 'name of the file where the result is saved' }),
+    'target-version': flags.string({ char: 't', description: 'asyncapi version to convert to' }),
+    id: flags.string({ char: 'i', description: 'application id, if needed' })
+  }
+
+  static args = [
+    { name: 'spec-file', description: 'spec path, url, or context-name', required: false },
+  ]
+
+  async run() {
+    const { args } = this.parse(Convert);
+    const filePath = args['spec-file'];
+    let specFile;
+
+    try {
+      specFile = await load(filePath);
+    } catch (err) {
+      if (err instanceof SpecificationFileNotFound) {
+        this.error(new ValidationError({
+          type: 'invalid-file',
+          filepath: filePath
+        }));
+      } else {
+        this.error(err as Error);
+      }
+    }
+    try {
+      if (specFile.getFilePath()) {
+        await parser.parse(specFile.text());
+        this.log(`File ${specFile.getFilePath()} successfully validated!`);
+      } else if (specFile.getFileURL()) {
+        await parser.parse(specFile.text());
+        this.log(`URL ${specFile.getFileURL()} successfully validated`);
+      }
+    } catch (error) {
+      throw new ValidationError({
+        type: 'parser-error',
+        err: error
+      });
+    }
+
+    /*     
+    try {
+      if (specFile.getFilePath()) {
+        await parser.parse(specFile.text());
+        this.log(`File ${specFile.getFilePath()} successfully validated!`);
+      } else if (specFile.getFileURL()) {
+        await parser.parse(specFile.text());
+        this.log(`URL ${specFile.getFileURL()} successfully validated`);
+      }
+    } catch (error) {
+      throw new ValidationError({
+        type: 'parser-error',
+        err: error
+      });
+    } */
+    
+    // Conversion
+    try {
+      
+
+    } catch (error) {
+      
+      };
+    }
+
+  }
--- a/src/commands/convert.ts
+++ b/src/commands/convert.ts
@@ -10,7 +10,7 @@ export default class Convert extends Command {
 
   static flags = {
     help: flags.help({ char: 'h' }),
-    'file-name': flags.string({ char: 'f', description: 'name of the file to convert' }),
+    'file-name': flags.string({ char: 'n', description: 'name of the file to convert' }),
     output: flags.string({ char: 'o', description: 'name of the file where the result is saved' }),
     'target-version': flags.string({ char: 't', description: 'asyncapi version to convert to' }),
     id: flags.string({ char: 'i', description: 'application id, if needed' })
@@ -24,7 +24,8 @@ export default class Convert extends Command {
     const { args } = this.parse(Convert);
     const filePath = args['spec-file'];
     let specFile;
-
+    
+    // LOAD FILE
     try {
       specFile = await load(filePath);
     } catch (err) {
@@ -37,6 +38,8 @@ export default class Convert extends Command {
         this.error(err as Error);
       }
     }
+
+    // VALIDATION
     try {
       if (specFile.getFilePath()) {
         await parser.parse(specFile.text());
@@ -51,30 +54,21 @@ export default class Convert extends Command {
         err: error
       });
     }
-
-    /*     
-    try {
-      if (specFile.getFilePath()) {
-        await parser.parse(specFile.text());
-        this.log(`File ${specFile.getFilePath()} successfully validated!`);
-      } else if (specFile.getFileURL()) {
-        await parser.parse(specFile.text());
-        this.log(`URL ${specFile.getFileURL()} successfully validated`);
-      }
-    } catch (error) {
-      throw new ValidationError({
-        type: 'parser-error',
-        err: error
-      });
-    } */
     
-    // Conversion
+    // CONVERSION
     try {
-      
-
+      /* if (specFile.getFilePath()) {
+        await parser.parse(specFile.text());
+        this.log(`File ${specFile.getFilePath()} successfully validated!`);
+      } else if (specFile.getFileURL()) {
+        await parser.parse(specFile.text());
+        this.log(`URL ${specFile.getFileURL()} successfully validated`);
+      } */
     } catch (error) {
-      
-      };
-    }
-
+/*       throw new ValidationError({
+        type: 'parser-error',
+        err: error
+      });
+ */ }
   }
+}
--- a/src/commands/convert.ts
+++ b/src/commands/convert.ts
@@ -1,3 +1,4 @@
+import { promises as fPromises } from 'fs';
 import { flags } from '@oclif/command';
 import * as parser from '@asyncapi/parser';
 import Command from '../base';
@@ -5,7 +6,6 @@ import { ValidationError } from '../errors/validation-error';
 import { load } from '../models/SpecificationFile';
 import { SpecificationFileNotFound } from '../errors/specification-file';
 import { convert } from '@asyncapi/converter';
-import { Specification } from '../models/SpecificationFile';
 
 export default class Convert extends Command {
   static description = 'convert asyncapi documents older to newer versions';
@@ -13,7 +13,7 @@ export default class Convert extends Command {
   static flags = {
     help: flags.help({ char: 'h' }),
     file: flags.string({ char: 'f', description: 'path to the file to convert' }),
-    output: flags.string({ char: 'o', description: 'name of the file where the result is saved' }),
+    output: flags.string({ char: 'o', description: 'path to the file where the result is saved' }),
     'target-version': flags.string({ char: 't', description: 'asyncapi version to convert to' }),
     id: flags.string({ char: 'i', description: 'application id, if needed' })
   }
@@ -43,10 +43,11 @@ export default class Convert extends Command {
 
       // CONVERSION
       convertedFile = await convert(specFile.text(), Convert.flags['target-version'], Convert.flags.id);
-      if (convertedFile.getFilePath()) {
-        this.log(`File ${convertedFile.getFilePath()} successfully converted!`);
-      } else if (convertedFile.getFileURL()) {
-        this.log(`URL ${convertedFile.getFileURL()} successfully converted!`);
+      this.log(`File ${specFile.getFilePath()} successfully converted!`);
+      if (Convert.flags.output) {
+        await fPromises.writeFile(`${Convert.flags.output}`, convertedFile, { encoding: 'utf8' });
+      } else {
+        this.log(`${convertedFile}`);
       }
     } catch (err) {
       if (err instanceof SpecificationFileNotFound) {


src/commands/convert.ts
```diff
@@ -0,0 +1,63 @@
...
+      if (specFile.getFilePath()) {
+        await parser.parse(specFile.text());
+        this.log(`File ${specFile.getFilePath()} successfully validated!`);
+      } else if (specFile.getFileURL()) {
+        await parser.parse(specFile.text());
+        this.log(`URL ${specFile.getFileURL()} successfully validated!`);
+      }
+
+      // CONVERSION
+      convertedFile = await convert(specFile.text(), Convert.flags['target-version'], Convert.flags.id);
```
fmvilas, line 45:
```suggestion
      convertedFile = await convert(specFile.text(), flags['target-version'], {
        id: flags.id,
      });
```

src/commands/convert.ts
```diff
@@ -0,0 +1,63 @@
...
+    'target-version': flags.string({ char: 't', description: 'asyncapi version to convert to' }),
+    id: flags.string({ char: 'i', description: 'application id, if needed' })
+  }
+
+  static args = [
+    { name: 'spec-file', description: 'spec path, url, or context-name', required: false },
+  ]
+
+  async run() {
+    const { args } = this.parse(Convert);
```
fmvilas, line 26:
```suggestion
    const { args, flags } = this.parse(Convert);
```

src/commands/convert.ts
fmvilas: The `convert` method expects an object at the end. See https://github.com/asyncapi/converter-js/#as-a-package

src/commands/convert.ts
fmvilas: If you do this, you don't have to do `Convert.flags.` in the rest of the file. Also, not sure `Convert.flags.whatever` would give you the value of the flag. I think it would give you the definition of the flag above: https://github.com/asyncapi/cli/pull/188/files#diff-ab22990f7615d8bd1a90b1e9c264b48bd488123b5c36a2095c205e6b21086e72R13.

src/commands/convert.ts
```diff
@@ -0,0 +1,63 @@
...
+        this.log(`File ${specFile.getFilePath()} successfully validated!`);
+      } else if (specFile.getFileURL()) {
+        await parser.parse(specFile.text());
+        this.log(`URL ${specFile.getFileURL()} successfully validated!`);
+      }
+
+      // CONVERSION
+      convertedFile = await convert(specFile.text(), Convert.flags['target-version'], Convert.flags.id);
+      this.log(`File ${specFile.getFilePath()} successfully converted!`);
+      if (Convert.flags.output) {
```
fmvilas, line 47:
```suggestion
      if (flags.output) {
```

src/commands/convert.ts
```diff
@@ -0,0 +1,63 @@
...
+        await fPromises.writeFile(`${Convert.flags.output}`, convertedFile, { encoding: 'utf8' });
```
fmvilas, line 48:
```suggestion
        await fPromises.writeFile(`${flags.output}`, convertedFile, { encoding: 'utf8' });
```

src/commands/convert.ts
```diff
@@ -0,0 +1,63 @@
...
+      } else {
+        this.log(`${convertedFile}`);
```
fmvilas, line 50:
```suggestion
        this.log(convertedFile);
```

You can log the variable directly :)

fmvilas (changes requested):
This is looking great, Peter! 🙌 

Left a few comments where you can improve the code. If I'm not mistaken, this should already work if you try in your machine 🎉

peter-rr:
@fmvilas All requested changes applied and seem to be working fine 👍 I think we're almost ready to start testing the `convert` command with real asyncapi files 🚀

peter-rr:
@Souvikns I wonder if you have any clue about an error I'm getting when passing the `-h` flag. The info displayed by the command when the `-h` flag is passed is correct, but an unexpected `Error: EEXIT: 0`  is shown at the bottom:

image

However, when the `--help` flag is passed, that error is not shown and everything is displayed correctly.

In order to solve this issue I've been trying to import the flags from `@oclif/core` as you mentioned in [this fix](https://github.com/asyncapi/cli/pull/201), but then I'm getting this error plugin when executing the command:

image

Do you know if there is something missing or that I'm not taking into consideration? Thank you!! :)

peter-rr:
@magicmatatjahu @fmvilas I've been testing the `--id `flag and I didn't get any successful result for none of the different spec file versions:

- For 2.x versions: The conversion is successful but the `--id` flag applied doesn't have any effect on the converted file.
- For 1.x versions: The conversion failed since the following error is thrown (I guess this is not related to the flag itself): `Error: Version 1.2.0 is not supported`.

So, my questions about this flag:
- Am I missing something at this flag's implementation? Or am I not considering the right use cases?
- Does it make sense to keep using this flag? Or should we consider to remove it from `convert` command.

What do you think, guys?

Souvikns:
@peter-rr can you confirm where you are seeing this error i.e. in your development environment or production.
I checked your branch and it is working as expected. 

Now there is an issue in production where we are getting a warning, the CLI works fine but still throws a warning. You can see https://github.com/asyncapi/cli/issues/192 for full details. We thought #201 would fix the issue but that is not the case currently I have been working on https://github.com/asyncapi/cli/pull/203 to migrate oclif to the latest version.

peter-rr:
@Souvikns Yeap, I see both errors in my development environment updated with the latest changes from production. Maybe I'm missing some package or plugin needed, not sure 🤔 

> I checked your branch and it is working as expected.

Do you mean you didn't get any of those errors running the command on my branch?

> We thought #201 would fix the issue but that is not the case currently I have been working on #203 to migrate oclif to the latest version.

Then I think I should wait until #203 is merged so I can test my branch with the latest oclif version and check if those errors still persist. I'll let you know :)

Souvikns:
> Do you mean you didn't get any of those errors when running the command on my branch?

Yeah, I opened your PR in my local setup. Do an `npm install` and see if it fixes the issue. Do let me know if it doesn't

src/commands/convert.ts
```diff
@@ -0,0 +1,69 @@
...
+      specFile = await load(filePath);
+
+      // VALIDATION
+      if (specFile.getFilePath()) {
+        await parser.parse(specFile.text());
+        this.log(`File ${specFile.getFilePath()} successfully validated!`);
+      } else if (specFile.getFileURL()) {
+        await parser.parse(specFile.text());
+        this.log(`URL ${specFile.getFileURL()} successfully validated!`);
+      }
```
magicmatatjahu, line 42: Here you have problem with that `Error: Version 1.2.0 is not supported` error :) ParserJS doesn't support `1.X.X` versions (including `2.0.0-rcX`), so you should first check if given version is supported and then make the validation process, if not you should only convert file :)

> Does it make sense to keep using this flag? Or should we consider to remove it from convert command.

I think that we should rid of that flag. Using this flag in the converter itself is quite "problematic", because you can always add the `id` field to the spec manually after the conversion, so I would get rid of it. cc @fmvilas

> For 2.x versions: The conversion is successful but the --id flag applied doesn't have any effect on the converted file.

Because that flag adds the `id` field to the converted file (from `1.X.X`), we don't change `id` between `2.X.X` -> `2.X.X` conversion.

magicmatatjahu (commented):
Awesome, but I have some insight :)

peter-rr:
> Do an `npm install` and see if it fixes the issue. Do let me know if it doesn't

I run it and now I'm getting this warning:
```
npm WARN deprecated mkdirp@0.5.1: Legacy versions of mkdirp are no longer supported. Please update to mkdirp 1.x. (Note that the API surface has changed to use Promises in 1.x.)

added 283 packages, removed 235 packages, changed 48 packages, and audited 2211 packages in 38s

125 packages are looking for funding
  run `npm fund` for details

47 vulnerabilities (21 moderate, 26 high)

To address issues that do not require attention, run:
  npm audit fix

To address all issues possible (including breaking changes), run:
  npm audit fix --force

Some issues need review, and may require choosing
a different dependency.

Run `npm audit` for details.
```
Should I update to mkdirp 1.x. to solve the issue? And how can I do it in that case? 🤔

src/commands/convert.ts
peter-rr: @magicmatatjahu @fmvilas 
Should we skip then the validation process for every version to avoid complexity and just convert the file? Do you agree?

magicmatatjahu: @peter-rr Atm converting `2.X.X` -> `2.X.X` only bumps the version of AsyncAPI, so we can say that even invalid spec, we will perform converting without problems. Of course we cannot validate the `1.X.X` versions. So... please don't validate any spec, only convert, hah 😄

peter-rr: Validation process removed  as well as `--id` flag. Everything seems to be working properly 👍

src/commands/convert.ts
```diff
@@ -0,0 +1,56 @@
...
+import { convert } from '@asyncapi/converter';
+
+export default class Convert extends Command {
+  static description = 'convert asyncapi documents older to newer versions';
+
+  static flags = {
+    help: flags.help({ char: 'h' }),
+    file: flags.string({ char: 'f', description: 'path to the file to convert' }),
+    output: flags.string({ char: 'o', description: 'path to the file where the result is saved' }),
+    'target-version': flags.string({ char: 't', description: 'asyncapi version to convert to', default: '2.3.0' })
```
magicmatatjahu, line 16: I think that we can have problems with hardcoded `2.3.0` version. We can forget with the 2.4.0 version release to bump this. I think that the best solution will be installing `@asyncapi/specs` package, where we have all JSON Schema for AsyncAPI spec and then make this:

```ts
import specs from '@asyncapi/specs';

const latestVersion = Object.keys(specs).pop() as string;
```

and we will be always up to date :)

src/commands/convert.ts
```diff
@@ -0,0 +1,56 @@
...
+    try {
+      // LOAD FILE
+      specFile = await load(filePath);
+
+      // CONVERSION
+      convertedFile = await convert(specFile.text(), flags['target-version'], {});
+      if (convertedFile) {
+        this.log(`File ${specFile.getFilePath()} successfully converted!`);
+      } else {
+        return;
```
magicmatatjahu, line 38: Please add here some log, why we stop functionality.

magicmatatjahu (changes requested):
Awesome! I added some comments, we need "fix" some parts :) Please also add some tests. If you will have problems, please ping us :)

src/commands/convert.ts
```diff
@@ -0,0 +1,56 @@
+import { promises as fPromises } from 'fs';
+import { flags } from '@oclif/command';
+import Command from '../base';
+import { ValidationError } from '../errors/validation-error';
+import { load } from '../models/SpecificationFile';
+import { SpecificationFileNotFound } from '../errors/specification-file';
+import { convert } from '@asyncapi/converter';
```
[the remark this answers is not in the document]
peter-rr: After installing and also cloning the repo (maybe not necessary) I've tried with `npm i --save-dev @types/asyncapi__converter` as suggested by the warning, but the result was unsuccessful:
```
npm ERR! code E404
npm ERR! 404 Not Found - GET https://registry.npmjs.org/@types%2fasyncapi__converter - Not found
npm ERR! 404  '@types/asyncapi__converter@*' is not in this registry.
```
So I tried the another option: just created a `global.d.ts` at `cli/src` directory with `declare module '@asyncapi/converter';` included on it, and it seems to be working fine. Is that a valid way to install the package? Or am I missing a more suitable way to do it? 🤔

src/commands/convert.ts
```diff
@@ -0,0 +1,56 @@
...
+
+export default class Convert extends Command {
+  static description = 'convert asyncapi documents older to newer versions';
+
+  static flags = {
+    help: flags.help({ char: 'h' }),
+    file: flags.string({ char: 'f', description: 'path to the file to convert' }),
```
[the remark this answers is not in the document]
peter-rr: Great 👌 Actually I tested it and it's working fine, but I guess it's not adding any value since we can use the 'spec-file' arg instead, right?

src/commands/convert.ts
peter-rr: I added `declare module '@asyncapi/specs';` to the `global.d.ts` file created before and everything works fine after adding the lines you suggested as well and changing the hardcoded version for `default: latestVersion` 👌

src/commands/convert.ts
peter-rr:
```suggestion
        this.log(`File ${specFile.getFilePath()} could not be converted.`);
        return;
```

peter-rr:
Now working on adding some tests 💪 🚀

src/commands/convert.ts
magicmatatjahu: Yeah, but awesome would be if we will describe why we cannot convert, you know, not supported version etc.

src/commands/convert.ts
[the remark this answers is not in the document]
magicmatatjahu: Yes, it doesn't add any value :)

src/commands/convert.ts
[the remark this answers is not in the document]
magicmatatjahu: > Is that a valid way to install the package? Or am I missing a more suitable way to do it? 🤔

We don't have any types (packages with types) for `@asyncapi/cli` so your approach is good.

magicmatatjahu:
@peter-rr If you think that this PR is ready to review, please open it (change from draft to normal PR) :)

src/commands/convert.ts
magicmatatjahu: Awesome! :)

src/commands/convert.ts
peter-rr: Taking into account the converter already throws a message when the conversion is not supported, which one among these options do you think should be more suitable? 🤔 

1. A generic message like:
`Conversion to target-version provided is not supported.`

2. A more specific message like:
`Conversions to lower, equal or non-existing target-version are not supported. `

These are the converter's messages I'm referring to:

image

magicmatatjahu: @peter-rr 

Could you wait for https://github.com/asyncapi/converter-js/pull/99 PR? When we will merge it, we will be able to print the message from error like:

```ts
this.log(e.message);
```

magicmatatjahu: And then (of course) we should wrap code with `try catch`:

```ts
try {
  convertedFile = await convert(specFile.text(), flags['target-version'], {});
} catch(e) {
  this.log(e.message);
}
```

peter-rr: Sure, I'll wait for it! Now I understand why the error cases from the converter were not handled by the catch block 😅 Thanks!

magicmatatjahu: @peter-rr Ok you can use `0.8.0` version of converter where we throw errors. Please update dep and propagate my suggestion about `try {} catch {}` and I guess that we can open PR to the final review, what do you think? :)

peter-rr: Yeah, sounds great 👌 Changes suggested already applied and opening PR for the final review 💪 
Tests are still missing though.

src/commands/convert.ts
```diff
@@ -0,0 +1,56 @@
...
+const latestVersion = Object.keys(specs).pop() as string;
+
+export default class Convert extends Command {
+  static description = 'convert asyncapi documents older to newer versions';
+
+  static flags = {
+    help: flags.help({ char: 'h' }),
+    output: flags.string({ char: 'o', description: 'path to the file where the result is saved' }),
+    'target-version': flags.string({ char: 't', description: 'asyncapi version to convert to', default: latestVersion })
+  }
```
magicmatatjahu, line 19: Please add `--watch, -w` flag like here https://github.com/asyncapi/cli/blob/master/src/commands/validate.ts#L14 and update `run` function like here https://github.com/asyncapi/cli/blob/master/src/commands/validate.ts#L28 to support watching of file :)

magicmatatjahu (changes requested):
I see that you have errors on the PR https://github.com/asyncapi/cli/runs/5274001453?check_suite_focus=true we need fix that problem with missed types for `@asyncapi/converter` and `@asyncapi/specs`. I see that `global.d.ts` doesn't dix that, so we have to handle that in different way.

Also please fix that problem https://sonarcloud.io/project/issues?resolved=false&types=CODE_SMELL&pullRequest=188&id=asyncapi_cli&open=AX6TFUzR8z_kgRrGAYC0 like here https://github.com/asyncapi/cli/blob/master/src/commands/validate.ts#L22 :)

And of course, we need some simple tests. Please let me know if you will have problems with that :) And please don't rush, we can wait for that command.

src/commands/convert.ts
fmvilas: 🤔 Do you think it makes sense to have `--watch` on the convert command? I can't imagine anyone editing an old AsyncAPI file while converting. Am I missing something?

magicmatatjahu: Right, I don't remember what I was thinking when I wrote that comment that the flag `watch` here would be needed 😅 

@peter-rr Please don't integrate `--watch` flag. Here it's unnecessary.

peter-rr:
Hi @Souvikns :)
I got all the 'ModuleLoad' errors below when trying to execute the `convert` command. I've installed the '@oclif/core' package, updated the package.json with the downloaded version (1.3.0). Should I update the oclif's dependencies for all the files listed or am I missing any other detail? 🤔 
```
Peter-Mac:cli pedro.ramos$ ./bin/run convert test/specification.yml
(node:4994) [MODULE_NOT_FOUND] ModuleLoadError Plugin: @asyncapi/cli: [MODULE_NOT_FOUND] require failed to load /Users/pedro.ramos/cli/lib/commands/config
module: @oclif/core@1.3.0
task: toCached
plugin: @asyncapi/cli
root: /Users/pedro.ramos/cli
See more details with DEBUG=*
(node:4994) [MODULE_NOT_FOUND] ModuleLoadError Plugin: @asyncapi/cli: [MODULE_NOT_FOUND] require failed to load /Users/pedro.ramos/cli/lib/commands/convert.js
module: @oclif/core@1.3.0
task: toCached
plugin: @asyncapi/cli
root: /Users/pedro.ramos/cli
See more details with DEBUG=*
(node:4994) [MODULE_NOT_FOUND] ModuleLoadError Plugin: @asyncapi/cli: [MODULE_NOT_FOUND] require failed to load /Users/pedro.ramos/cli/lib/commands/diff.js
module: @oclif/core@1.3.0
task: toCached
plugin: @asyncapi/cli
root: /Users/pedro.ramos/cli
See more details with DEBUG=*
(node:4994) [MODULE_NOT_FOUND] ModuleLoadError Plugin: @asyncapi/cli: [MODULE_NOT_FOUND] require failed to load /Users/pedro.ramos/cli/lib/commands/new.js
module: @oclif/core@1.3.0
task: toCached
plugin: @asyncapi/cli
root: /Users/pedro.ramos/cli
See more details with DEBUG=*
(node:4994) [MODULE_NOT_FOUND] ModuleLoadError Plugin: @asyncapi/cli: [MODULE_NOT_FOUND] require failed to load /Users/pedro.ramos/cli/lib/commands/start
module: @oclif/core@1.3.0
task: toCached
plugin: @asyncapi/cli
root: /Users/pedro.ramos/cli
See more details with DEBUG=*
(node:4994) [MODULE_NOT_FOUND] ModuleLoadError Plugin: @asyncapi/cli: [MODULE_NOT_FOUND] require failed to load /Users/pedro.ramos/cli/lib/commands/validate.js
module: @oclif/core@1.3.0
task: toCached
plugin: @asyncapi/cli
root: /Users/pedro.ramos/cli
See more details with DEBUG=*
(node:4994) [MODULE_NOT_FOUND] ModuleLoadError Plugin: @asyncapi/cli: [MODULE_NOT_FOUND] require failed to load /Users/pedro.ramos/cli/lib/commands/config/context
module: @oclif/core@1.3.0
task: toCached
plugin: @asyncapi/cli
root: /Users/pedro.ramos/cli
See more details with DEBUG=*
(node:4994) [MODULE_NOT_FOUND] ModuleLoadError Plugin: @asyncapi/cli: [MODULE_NOT_FOUND] require failed to load /Users/pedro.ramos/cli/lib/commands/start/studio.js
module: @oclif/core@1.3.0
task: toCached
plugin: @asyncapi/cli
root: /Users/pedro.ramos/cli
See more details with DEBUG=*
(node:4994) [MODULE_NOT_FOUND] ModuleLoadError Plugin: @asyncapi/cli: [MODULE_NOT_FOUND] require failed to load /Users/pedro.ramos/cli/lib/commands/config/context/add.js
module: @oclif/core@1.3.0
task: toCached
plugin: @asyncapi/cli
root: /Users/pedro.ramos/cli
See more details with DEBUG=*
(node:4994) [MODULE_NOT_FOUND] ModuleLoadError Plugin: @asyncapi/cli: [MODULE_NOT_FOUND] require failed to load /Users/pedro.ramos/cli/lib/commands/config/context/current.js
module: @oclif/core@1.3.0
task: toCached
plugin: @asyncapi/cli
root: /Users/pedro.ramos/cli
See more details with DEBUG=*
(node:4994) [MODULE_NOT_FOUND] ModuleLoadError Plugin: @asyncapi/cli: [MODULE_NOT_FOUND] require failed to load /Users/pedro.ramos/cli/lib/commands/config/context/list.js
module: @oclif/core@1.3.0
task: toCached
plugin: @asyncapi/cli
root: /Users/pedro.ramos/cli
See more details with DEBUG=*
(node:4994) [MODULE_NOT_FOUND] ModuleLoadError Plugin: @asyncapi/cli: [MODULE_NOT_FOUND] require failed to load /Users/pedro.ramos/cli/lib/commands/config/context/remove.js
module: @oclif/core@1.3.0
task: toCached
plugin: @asyncapi/cli
root: /Users/pedro.ramos/cli
See more details with DEBUG=*
(node:4994) [MODULE_NOT_FOUND] ModuleLoadError Plugin: @asyncapi/cli: [MODULE_NOT_FOUND] require failed to load /Users/pedro.ramos/cli/lib/commands/config/context/use.js
module: @oclif/core@1.3.0
task: toCached
plugin: @asyncapi/cli
root: /Users/pedro.ramos/cli
See more details with DEBUG=*
 ›   Error: command convert:test/specification.yml not found
```

Souvikns:
Hey @peter-rr  I don't think you need to update `@oclif/core`, somethings changed from before since we updated to the latest oclif version, now you have to build every time you want to run the command localy so just delete the `lib` folder and run `npm run build` and it should work. 

Do let me know if the problem is not solved.

peter-rr:
> Hey @peter-rr I don't think you need to update `@oclif/core`, somethings changed from before since we updated to the latest oclif version, now you have to build every time you want to run the command localy so just delete the `lib` folder and run `npm run build` and it should work.
> 
> Do let me know if the problem is not solved.

Yay! Now it's working 💪 Thank you @Souvikns  :)

peter-rr:
@Souvikns By the way, I wonder how to solve these errors related to missed types for `@asyncapi/converter` and `@asyncapi/specs` : https://github.com/asyncapi/cli/runs/5424612848?check_suite_focus=true

Do you have any clue about them? 🤔

Souvikns:
Typescript is throwing this error because `spec` and `converter` are js projects and they lack types so there are basically two ways to go around this, 
1. You create a `.d.ts` file and write the types as per your requirements (this is more preferred)
2. you can ts ignore the line

Now I see you have already created a `.d.ts` file but you have to change the `tsconfig.json` file. Just add your `global.d.ts` file inside `include` array   

https://github.com/asyncapi/cli/blob/20004196500bbbb8011143d15dfcdbb337bc6dad/tsconfig.json#L12

peter-rr:
> Now I see you have already created a `.d.ts` file but you have to change the `tsconfig.json` file. Just add your `global.d.ts` file inside `include` array

Still getting the same errors after applying that change. I guess there is something I'm doing wrong or not taking into account 🤔

src/global.d.ts
```diff
@@ -0,0 +1,2 @@
+declare module '@asyncapi/converter';
```
Souvikns, line 1:
```suggestion
declare module '@asyncapi/converter' {
  export async function convert(spec: string, targetVersion: string, options: any): Promise<any>
}
```

@peter-rr try writing the module as per your needs, I am just giving an example for `convert` do this for `@asyncapi/spec` as well. 

You can take some inspiration from https://github.com/asyncapi/server-api/blob/master/src/server-api.d.ts

src/commands/convert.ts
```diff
@@ -0,0 +1,59 @@
+import { promises as fPromises } from 'fs';
+import { Command, Flags } from '@oclif/core';
+import { ValidationError } from '../errors/validation-error';
+import { load } from '../models/SpecificationFile';
+import { SpecificationFileNotFound } from '../errors/specification-file';
+import { convert } from '@asyncapi/converter';
+import specs from '@asyncapi/specs';
```
magicmatatjahu, line 7:
```suggestion
// @ts-ignore
import specs from '@asyncapi/specs';
```

peter-rr:
> Please propagate my changes. There's problem with `ts-node` and how exactly it reads the custom typings from external modules.

Great! Problems related to external modules are solved now and I'm trying to fix all the current failing tests. They are only 4 at this moment 💪

src/commands/convert.ts
```diff
@@ -0,0 +1,62 @@
+/* eslint-disable @typescript-eslint/ban-ts-comment */
+import { promises as fPromises } from 'fs';
+import { Command, Flags } from '@oclif/core';
```
Souvikns, line 3:
```suggestion
import { Flags } from '@oclif/core';
import Command from '../base';
```

Use the custom base class we have created for error handling.

peter-rr:
@Souvikns @magicmatatjahu
All tests passing now 🎉 Let me know if anything missing or could be improved :)

Souvikns (approved):
Wow 🤩 looks good, wait for @magicmatatjahu's approval then we can merge this.

fmvilas (approved):
LGTM 🚀

magicmatatjahu (approved):
Thanks @peter-rr for contribution! LGTM! 🚀

Issue #192, opened 2022-01-23
CLI is breaking in production!


Souvikns:
#### Describe the bug
CLI is working fine in the development environment, but in production, it is breaking, no command is usable and most probably this is happening because of some plugin issue. 

![image](https://user-images.githubusercontent.com/41781438/150667224-19b7c363-913f-4853-88a5-efa0d4857c05.png)

```
(node:11624) [MODULE_NOT_FOUND] Error Plugin: @asyncapi/cli: Cannot find module '@oclif/plugin-help/lib/command'
Require stack:
- C:\Users\XRIG\AppData\Roaming\npm\node_modules\@asyncapi\cli\node_modules\@fmvilas\oclif-plugin-spaced-commands\lib\hooks\init.js
- C:\Users\XRIG\AppData\Roaming\npm\node_modules\@asyncapi\cli\node_modules\@oclif\config\lib\config.js
- C:\Users\XRIG\AppData\Roaming\npm\node_modules\@asyncapi\cli\node_modules\@oclif\config\lib\index.js
- C:\Users\XRIG\AppData\Roaming\npm\node_modules\@asyncapi\cli\node_modules\@oclif\command\lib\command.js
- C:\Users\XRIG\AppData\Roaming\npm\node_modules\@asyncapi\cli\node_modules\@oclif\command\lib\index.js
- C:\Users\XRIG\AppData\Roaming\npm\node_modules\@asyncapi\cli\bin\run
module: @oclif/config@1.18.2
task: runHook init
plugin: @asyncapi/cli
root: C:\Users\XRIG\AppData\Roaming\npm\node_modules\@asyncapi\cli
See more details with DEBUG=*
(Use `node --trace-warnings ...` to show where the warning was created)
    Error: Unable to load configured help class "./lib/help", failed with message:
    Cannot find module '@oclif/plugin-help/lib/command'
    Require stack:
    - C:\Users\XRIG\AppData\Roaming\npm\node_modules\@asyncapi\cli\lib\help\CommandHelper.js
    - C:\Users\XRIG\AppData\Roaming\npm\node_modules\@asyncapi\cli\lib\help\index.js
    -
    C:\Users\XRIG\AppData\Roaming\npm\node_modules\@asyncapi\cli\node_modules\@oclif\help\lib\util.js
    - C:\Users\XRIG\AppData\Roaming\npm\node_modules\@asyncapi\cli\node_modules\@oclif\help\lib\comman
    d.js
    -
[...]

PR #201, opened 2022-01-26
fix: fix cli breaking in production

Souvikns:
**Description**
Removing custom help for now, as it is causing the CLI to break in production. oclif has updated and `@oclif/plugin-help` is now under `@oclif/core` so I think we could just update the whole CLI and remove `@oclif/command` and follow as per their documentation. 

**Related issue(s)**

See also #192


derberg (dismissed):

magicmatatjahu:
As I see we have still `"@oclif/plugin-help": "^3.2.3",` in deps in this PR, I think we should remove that plugin, if we don't use it.

Souvikns:
I have removed the `@oclif/plugin-help` and it is still working as intended, I was under the impression that it needs that plugin for the help command to work.

magicmatatjahu (approved):

derberg (approved):
:shipit:

Souvikns:
No `oclif/plugin-help` is a plugin that handled the help command functionalities, but now it is already in core so we do not need this plugin anymore.

imabp:
Thank you @Souvikns  :) for the clarification

Souvikns:
Will merge this once @fmvilas approves. 

> Also, this is something that I found today https://github.com/asyncapi/cli/issues/192#issuecomment-1023041282

derberg:
No need to wait. You have it approved. It is a bug that makes CLI not work so yeah, merge it 🙂

PR #221, opened 2022-02-09
feat: add generate fromTemplate command

### LINKED ISSUE #9: generator features
CLI will be porting all the functionalities from the generator CLI, so as of now the supported options and commands by the generator CLI are 

#### arguments
`<asyncapi> <template>`
`<asyncapi>` - path to the asyncapi file 
`<template>` - template name 

#### Options
- `-d, --disable-hook [hooks...]` disable a specific hook type or hooks from a given hook type
- `--debug` to enable specific errors in the console 
- `-i, --install` installs the template and its dependencies (defaults to false)
- `-n, --no-overwrite <glob>` glob or path of the file(s) to skip when regenerating
- `-o, --output <outputDir>` directory where to put the generated files (defaults to current directory)
- `-p, --param <name=value>` additional param to pass to templates
- `--force-write` force writing of the generated files to given directory even if it is a git repo with unstaged files or not empty dir (defaults to false)
- `--watch-template`  watches the template directory and the AsyncAPI document, and re-generate the files when changes occur. Ignores the output directory. 
- `--map-base-url <url:folder>`  maps all schema references from base url to local folder

Souvikns:
**Description**
This PR aims to implement generate command that will replace [Generator CLI](https://github.com/asyncapi/generator#cli-usage)

**Related issue(s)**

Resolves #9


Souvikns:
@magicmatatjahu should I create local types for https://github.com/asyncapi/generator or is there a plan for adding typescript support?

magicmatatjahu:
@Souvikns 

> should I create local types for https://github.com/asyncapi/generator or is there a plan for adding typescript support?

We have plan to add types https://github.com/asyncapi/generator/issues/468 but as I see you added types in this project. For me it's not a problem, we can always remove it and switch to the official typing.

src/commands/generate.ts
```diff
@@ -0,0 +1,119 @@
...
+      }),
+      output: flags.string({
+        char: 'o',
+        description: 'directory where to put the generated files (defaults to current directory)'
+      }),
+      'force-write': flags.boolean({
+        default: false,
+        description: 'force writing of the generated files to given directory even if it is a git repo with unstaged files or not empty dir (defaults to false)'
+      }),
+      'watch-tempalte': flags.boolean({
```
magicmatatjahu, line 41: @Souvikns Some tip: you can use https://github.com/asyncapi/cli/blob/master/src/flags.ts#L3 that flag and that https://github.com/asyncapi/cli/blob/master/src/globals.ts#L28 helper to enable watcher :)

Souvikns:
@magicmatatjahu I need your help with this 

For some reason, due to generate command the CLI is throwing ts error all over the project.

Currently, I am getting these warnings when I run CLI. 

<details>

<summary>Error Logs</summary>

```
PS D:\Programs\AsyncAPI\cli> ./bin/run 
(node:6664) TSError Plugin: @asyncapi/cli: ⨯ Unable to compile TypeScript:
src/commands/config/context.ts:2:5 - error TS7022: '__importDefault' implicitly has type 'any' because it does not have a type annotation and is referenced directly or indirectly in its own initializer.

2 var __importDefault = (this && this.__importDefault) || function (mod) {
      ~~~~~~~~~~~~~~~
src/commands/config/context.ts:2:67 - error TS7006: Parameter 'mod' implicitly has an 'any' type.

2 var __importDefault = (this && this.__importDefault) || function (mod) {
                                                                    ~~~

module: @oclif/config@1.17.0
task: toCached
plugin: @asyncapi/cli
root: D:\Programs\AsyncAPI\cli
See more details with DEBUG=*
(node:6664) TSError Plugin: @asyncapi/cli: ⨯ Unable to compile TypeScript:
src/commands/start/studio.ts:2:67 - error TS7006: Parameter 'mod' implicitly has an 'any' type.

2 var __importDefault = (this && this.__importDefault) || function (mod) {
                                                                    ~~~
src/commands/start/studio.ts:6:7 - error TS2451: Cannot redeclare block-scoped variable 'command_1'.

6 const command_1 = require("@oclif/command");
        ~~~~~~~~~

  src/commands/config/context.ts:6:7
    6 const command_1 = require("@oclif/command");
            ~~~~~~~~~
    'command_1' was also declared here.
src/commands/start/studio.ts:7:7 - error TS2451: Cannot redeclare block-scoped variable 'base_1'.

7 const base_1 = __importDefault(require("../../base"));
        ~~~~~~

  src/commands/config/context.ts:7:7
    7 const base_1 = __importDefault(require("../../base"));
            ~~~~~~
    'base_1' was also declared here.

module: @oclif/config@1.17.0
task: toCached
plugin: @asyncapi/cli
root: D:\Programs\AsyncAPI\cli
See more details with DEBUG=*
(node:6664) TSError Plugin: @asyncapi/cli: ⨯ Unable to compile TypeScript:
src/commands/config/context/add.ts:2:67 - error TS7006: Parameter 'mod' implicitly has an 'any' type.

2 var __importDefault = (this && this.__importDefault) || function (mod) {
                                                                    ~~~
src/commands/config/context/add.ts:6:7 - error TS2451: Cannot redeclare block-scoped variable 'command_1'.

6 const command_1 = require("@oclif/command");
        ~~~~~~~~~

  src/commands/config/context.ts:6:7
    6 const command_1 = require("@oclif/command");
            ~~~~~~~~~
    'command_1' was also declared here.
src/commands/config/context/add.ts:7:7 - error TS2451: Cannot redeclare block-scoped variable 'base_1'.

7 const base_1 = __importDefault(require("../../../base"));
        ~~~~~~

  src/commands/config/context.ts:7:7
    7 const base_1 = __importDefault(require("../../base"));
            ~~~~~~
    'base_1' was also declared here.

module: @oclif/config@1.17.0
task: toCached
plugin: @asyncapi/cli
root: D:\Programs\AsyncAPI\cli
See more details with DEBUG=*
(node:6664) TSError Plugin: @asyncapi/cli: ⨯ Unable to compile TypeScript:
src/commands/config/context/current.ts:2:67 - error TS7006: Parameter 'mod' implicitly has an 'any' type.

2 var __importDefault = (this && this.__importDefault) || function (mod) {
                                                                    ~~~
src/commands/config/context/current.ts:6:7 - error TS2451: Cannot redeclare block-scoped variable 'command_1'.

6 const command_1 = require("@oclif/command");
        ~~~~~~~~~

  src/commands/config/context.ts:6:7
    6 const command_1 = require("@oclif/command");
            ~~~~~~~~~
    'command_1' was also declared here.
src/commands/config/context/current.ts:7:7 - error TS2451: Cannot redeclare block-scoped variable 'base_1'.

7 const base_1 = __importDefault(require("../../../base"));
        ~~~~~~

  src/commands/config/context.ts:7:7
    7 const base_1 = __importDefault(require("../../base"));
            ~~~~~~
    'base_1' was also declared here.
src/commands/config/context/current.ts:8:7 - error TS2451: Cannot redeclare block-scoped variable 'Context_1'.

8 const Context_1 = require("../../../models/Context");
        ~~~~~~~~~

  src/commands/config/context/add.ts:8:7
    8 const Context_1 = require("../../../models/Context");
            ~~~~~~~~~
    'Context_1' was also declared here.

module: @oclif/config@1.17.0
task: toCached
plugin: @asyncapi/cli
root: D:\Programs\AsyncAPI\cli
See more details with DEBUG=*
(node:6664) TSError Plugin: @asyncapi/cli: ⨯ Unable to compile TypeScript:
src/commands/config/context/list.ts:2:67 - error TS7006: Parameter 'mod' implicitly has an 'any' type.

2 var __importDefault = (this && this.__importDefault) || function (mod) {
                                                                    ~~~
src/commands/config/context/list.ts:6:7 - error TS2451: Cannot redeclare block-scoped variable 'command_1'.

6 const command_1 = require("@oclif/command");
        ~~~~~~~~~

  src/commands/config/context.ts:6:7
    6 const command_1 = require("@oclif/command");
            ~~~~~~~~~
    'command_1' was also declared here.
src/commands/config/context/list.ts:7:7 - error TS2451: Cannot redeclare block-scoped variable 'base_1'.

7 const base_1 = __importDefault(require("../../../base"));
        ~~~~~~

  src/commands/config/context.ts:7:7
    7 const base_1 = __importDefault(require("../../base"));
            ~~~~~~
    'base_1' was also declared here.
src/commands/config/context/list.ts:8:7 - error TS2451: Cannot redeclare block-scoped variable 'Context_1'.

8 const Context_1 = require("../../../models/Context");
        ~~~~~~~~~

  src/commands/config/context/add.ts:8:7
    8 const Context_1 = require("../../../models/Context");
            ~~~~~~~~~
    'Context_1' was also declared here.

module: @oclif/config@1.17.0
task: toCached
plugin: @asyncapi/cli
root: D:\Programs\AsyncAPI\cli
See more details with DEBUG=*
(node:6664) TSError Plugin: @asyncapi/cli: ⨯ Unable to compile TypeScript:
src/commands/config/context/remove.ts:2:67 - error TS7006: Parameter 'mod' implicitly has an 'any' type.

2 var __importDefault = (this && this.__importDefault) || function (mod) {
                                                                    ~~~
src/commands/config/context/remove.ts:6:7 - error TS2451: Cannot redeclare block-scoped variable 'command_1'.

6 const command_1 = require("@oclif/command");
        ~~~~~~~~~

  src/commands/config/context.ts:6:7
    6 const command_1 = require("@oclif/command");
            ~~~~~~~~~
    'command_1' was also declared here.
src/commands/config/context/remove.ts:7:7 - error TS2451: Cannot redeclare block-scoped variable 'base_1'.

7 const base_1 = __importDefault(require("../../../base"));
        ~~~~~~

  src/commands/config/context.ts:7:7
    7 const base_1 = __importDefault(require("../../base"));
            ~~~~~~
    'base_1' was also declared here.
src/commands/config/context/remove.ts:8:7 - error TS2451: Cannot redeclare block-scoped variable 'Context_1'.

8 const Context_1 = require("../../../models/Context");
        ~~~~~~~~~

  src/commands/config/context/add.ts:8:7
    8 const Context_1 = require("../../../models/Context");
            ~~~~~~~~~
    'Context_1' was also declared here.

module: @oclif/config@1.17.0
task: toCached
plugin: @asyncapi/cli
root: D:\Programs\AsyncAPI\cli
See more details with DEBUG=*
(node:6664) TSError Plugin: @asyncapi/cli: ⨯ Unable to compile TypeScript:
src/commands/config/context/use.ts:2:67 - error TS7006: Parameter 'mod' implicitly has an 'any' type.

2 var __importDefault = (this && this.__importDefault) || function (mod) {
                                                                    ~~~
src/commands/config/context/use.ts:6:7 - error TS2451: Cannot redeclare block-scoped variable 'command_1'.

6 const command_1 = require("@oclif/command");
        ~~~~~~~~~

  src/commands/config/context.ts:6:7
    6 const command_1 = require("@oclif/command");
            ~~~~~~~~~
    'command_1' was also declared here.
src/commands/config/context/use.ts:7:7 - error TS2451: Cannot redeclare block-scoped variable 'base_1'.

7 const base_1 = __importDefault(require("../../../base"));
        ~~~~~~

  src/commands/config/context.ts:7:7
    7 const base_1 = __importDefault(require("../../base"));
            ~~~~~~
    'base_1' was also declared here.
src/commands/config/context/use.ts:8:7 - error TS2451: Cannot redeclare block-scoped variable 'Context_1'.

8 const Context_1 = require("../../../models/Context");
        ~~~~~~~~~

  src/commands/config/context/add.ts:8:7
    8 const Context_1 = require("../../../models/Context");
            ~~~~~~~~~
    'Context_1' was also declared here.

module: @oclif/config@1.17.0
task: toCached
plugin: @asyncapi/cli
root: D:\Programs\AsyncAPI\cli
See more details with DEBUG=*
All in one CLI for all AsyncAPI tools

VERSION
  @asyncapi/cli/0.14.1 win32-x64 node-v16.13.0

USAGE
  $ asyncapi [COMMAND]

COMMANDS
  config    access configs
  diff      find diff between two asyncapi files
  generate  Generator is a tool that you can use to generate whatever you want basing on the AsyncAPI specification file as an input.
  new       creates a new asyncapi file
  validate  validate asyncapi file

PS D:\Programs\AsyncAPI\cli> 
```

</details>

These warnings are coming from here 
```ts
const generator = new Generator(template, flags.output || path.resolve(os.tmpdir(), 'asyncapi-generator'), {
  forceWrite: flags['force-write'],
  install: flags.install,
  debug: flags.debug,
  templateParams: params,
  noOverwriteGlobs: flags['no-overwrite'],
  mapBaseUrlToFolder: flags['map-base-url'],
  disableHooks,
  mapBaseURLToFolder
});
await generator.generateFomString(asyncapi.text());
```
To be precise it breaks in when I create a new `Generator` object. I am trying to follow [this](https://github.com/asyncapi/generator#example-using-the-module), but maybe I am missing something.

magicmatatjahu:
@Souvikns I will check that but as I remember we had similar issue in the `server-api` project and I changed the imports to fix them. It's a problem with TS/TS-node, not with generator package.

Souvikns:
https://github.com/asyncapi/generator/blob/1359f15a99a72f58380249b489bded4c29d0e978/package.json#L76-L77

@magicmatatjahu why does `generator` has `ts-node` and `typescript` as dependencies also why does it need typescript when it is a js project? 

> Found https://github.com/TypeStrong/ts-node/issues/883 is this relevant in our case.

Souvikns:
@magicmatatjahu I tried ignoring the import but I am getting the same errors.

Souvikns:
The tests are running in my local setup 😞 
image

Souvikns:
@magicmatatjahu, @derberg generator command is ready for review. I have added a spinner as per https://github.com/asyncapi/cli/issues/9#issuecomment-1081950527, I want some opinion on that.

src/commands/generate.ts
```diff
@@ -0,0 +1,161 @@
...
+import Command from '../base';
+// eslint-disable-next-line
+// @ts-ignore
+import AsyncAPIGenerator from '@asyncapi/generator';
+import path from 'path';
+import os from 'os';
+import { load } from '../models/SpecificationFile';
+import { Example } from '@oclif/core/lib/interfaces';
+
+export class GenerateFlagParser {
```
magicmatatjahu, line 11: Please avoid creating helper classes, that you can "merge" to the command class. Please move `parse` method to the `Generator` class and call it as `parseFlags` and return from it object as:

```ts
interface ParsedFlags {
  params: Record<string, string>;
  disableHooks: Record<string, string>;
  mapBaseUrlToFolder: {
    url: string;
    folder: string;
  }
}
```

src/commands/generate.ts
```diff
@@ -0,0 +1,161 @@
...
+      char: 'o',
+      description: 'directory where to put the generated files (defaults to current directory)'
+    }),
+    'force-write': Flags.boolean({
+      default: false,
+      description: 'force writing of the generated files to given directory even if it is a git repo with unstaged files or not empty dir (defaults to false)'
+    }),
+    'watch-tempalte': Flags.boolean({
+      description: 'watches the template directory and the AsyncAPI document, and re-generate the files when changes occur. Ignores the output directory.'
+    }),
```
magicmatatjahu, line 121: we should use `watch` flag from it https://github.com/asyncapi/cli/blob/master/src/flags.ts#L3

src/commands/generate.ts
```diff
@@ -0,0 +1,161 @@
...
+      multiple: true
+    }),
+    'map-base-url': Flags.string({
+      description: ' maps all schema references from base url to local folder'
+    })
+  }
+
+  static args = [
+    { name: 'asyncapi', description: 'Local path or url pointing to AsyncAPI specification file', required: true },
+    { name: 'template', description: 'Name of the generator template like for example @asyncapi/html-template or https://github.com/asyncapi/html-template' }
```
magicmatatjahu, line 134: It's also required.

src/commands/generate.ts
```diff
@@ -0,0 +1,161 @@
...
+    const asyncapi = await load(args['asyncapi']);
+    const template = args['template'];
+
+    const flagParser = new GenerateFlagParser(flags['disable-hook'], flags['param'], flags['map-base-url'] as string);
+
+    const params = flagParser.params();
+    const disableHooks = flagParser.disableHooks();
+    const mapBaseURLToFolder = flagParser.mapBaseUrlToFolder();
+
+    const generator = new AsyncAPIGenerator(template, flags.output || path.resolve(os.tmpdir(), 'asyncapi-generator'), {
```
magicmatatjahu, line 148: I know that we use `path.resolve(os.tmpdir(), 'asyncapi-generator')` as fallback in current CLI, but I think that we should annotate `output` flag as required.

magicmatatjahu (changes requested):
I leave some important comments :) Also please try to implement `watch` mode as in current CLI in generator repo (https://github.com/asyncapi/generator/blob/master/cli.js#L108). Please ping me if you will have problems.

src/commands/generate.ts
Souvikns: We have a custom description in the watch flag for the generator command, I think we can convert [`watchFlag`](https://github.com/asyncapi/cli/blob/8fe89e2d122517775905d80531de3a0edc8f6244/src/flags.ts#L3) into a function that would return the watch flag then we can have custom description.

derberg:
I'm not gonna do code review as I see @magicmatatjahu already did it. Just let me know when is the right time to do user testing on local. Happy to help and check DX

src/commands/generate.ts
magicmatatjahu: I suggested you to remove that class and move all methods (as private) to the `Generate` class :)

src/commands/generate.ts
magicmatatjahu: If you want, you can convert `watchFlag` to the function, but if you don't please only change name of flag to the `--watch` and add fallback `-w`

src/commands/generate.ts
magicmatatjahu: As output is required then `path.resolve(os.tmpdir(), 'asyncapi-generator')` is not needed.

src/flags.ts
```diff
@@ -1,6 +1,11 @@
...
-  description: 'Enable watch mode'
-});
+export const watchFlag = (description?: string) => {
+  if (!description) {
+    description = 'Enable watch node';
+  }
+  return Flags.boolean({
+    char: 'w',
+    description
+  });
```
magicmatatjahu, line 10:
```suggestion
  return Flags.boolean({
    char: 'w',
    description: description || 'Enable watch node',
  });
```

magicmatatjahu (changes requested):

Souvikns:
I am getting this message when running `asyncapi --help` 

```
=============================================
simple-git has supported promises / async await since version 2.6.0.
 Importing from 'simple-git/promise' has been deprecated and will be
 removed by July 2022.

To upgrade, change all 'simple-git/promise' imports to just 'simple-git'
=============================================

```
CLI is not using `simple-git`, but maybe a tool we depend on is using it.

derberg:
@Souvikns generator has `simple-git` as dependency

magicmatatjahu:
@derberg We know, in the PR we used old version of generator, but after bumping the issue should be resolved :)

Souvikns:
Yeah, I have bumped the generator version, pr is ready for review now.

src/commands/generate.ts
```diff
@@ -0,0 +1,149 @@
...
+    output: Flags.string({
+      char: 'o',
+      description: 'directory where to put the generated files (defaults to current directory)',
+      required: true
+    }),
+    'force-write': Flags.boolean({
+      default: false,
+      description: 'force writing of the generated files to given directory even if it is a git repo with unstaged files or not empty dir (defaults to false)'
+    }),
+    'watch-tempalte': watchFlag(
```
magicmatatjahu, line 56:
```suggestion
    'watch-template': watchFlag(
```

src/commands/generate.ts
```diff
@@ -0,0 +1,149 @@
...
+      description: 'disable a specific hook type or hooks from a given hook type',
+      multiple: true
+    }),
+    install: Flags.boolean({
+      char: 'i',
+      default: false,
+      description: 'installs the template and its dependencies (defaults to false)'
+    }),
+    debug: Flags.boolean({
+      description: 'to enable specific errors in the console'
```
magicmatatjahu, line 40:
```suggestion
      description: 'enable more specific errors in the console'
```

magicmatatjahu (changes requested):
For me implementation is good. Only thing that I don't see is the lack of implementation for the `watch` flag. https://github.com/asyncapi/generator/blob/master/cli.js#L108

@derberg Do you want to test PR from UX perspective? In previous comments you wrote you want to 😄

derberg:
I definitely want to do DX manual testing, gimme few days 🙏🏼

magicmatatjahu:
@derberg No problem, however remember that now we don't have implemented `--watch-template` functionality in this PR.

@Souvikns If you want, I can help you with it :)

Souvikns:
Oh sorry, I was kinda confused about how the CLI's watcher is working https://github.com/asyncapi/cli/blob/b0cd740a6b405c04c092345c1db02097ee53c47c/src/globals.ts#L28, I will give another try, give me some time on this. 


> @Souvikns If you want, I can help you with it :)

Thanks @magicmatatjahu I will ask you if get stuck.

Souvikns:
The watcher is not working correctly. 

Getting this error 

```
➜ ./bin/run generate https://bit.ly/asyncapi @asyncapi/html-template --output ./docs -w           
Error: Cannot find module '/Users/postman/Documents/GitHub/cli/@asyncapi/html-template/package.json'
Require stack:
- /Users/postman/Documents/GitHub/cli/lib/commands/generate.js
- /Users/postman/Documents/GitHub/cli/node_modules/@oclif/core/lib/module-loader.js
- /Users/postman/Documents/GitHub/cli/node_modules/@oclif/core/lib/config/plugin.js
- /Users/postman/Documents/GitHub/cli/node_modules/@oclif/core/lib/config/config.js
- /Users/postman/Documents/GitHub/cli/node_modules/@oclif/core/lib/config/index.js
- /Users/postman/Documents/GitHub/cli/node_modules/@oclif/core/lib/command.js
- /Users/postman/Documents/GitHub/cli/node_modules/@oclif/core/lib/index.js
- /Users/postman/Documents/GitHub/cli/bin/run

```

@magicmatatjahu need your help here.

Souvikns:
Thanks, to @magicmatatjahu implementation, is ready, @derberg I think you can do manual tests now.

src/commands/generate.ts
```diff
@@ -0,0 +1,276 @@
...
+  params: Record<string, string>,
+  disableHooks: Record<string, string>,
+  mapBaseUrlToFolder: {
+    url: string,
+    folder: string
+  }
+}
+
+export default class Generate extends Command {
+  static description = 'Generator is a tool that you can use to generate whatever you want basing on the AsyncAPI specification file as an input.';
```
derberg, line 29:
```suggestion
  static description = 'generate whatever you want using templates compatible with AsyncAPI Generator';
```

src/commands/generate.ts
```diff
@@ -0,0 +1,276 @@
...
+      multiple: true
+    }),
+    'map-base-url': Flags.string({
+      description: ' maps all schema references from base url to local folder'
+    })
+  }
+
+  static args = [
+    { name: 'asyncapi', description: 'Local path or url pointing to AsyncAPI specification file', required: true },
+    { name: 'template', description: 'Name of the generator template like for example @asyncapi/html-template or https://github.com/asyncapi/html-template', required: true }
```
derberg, line 79:
```suggestion
    { name: 'asyncapi', description: '- Local path or url pointing to AsyncAPI file', required: true },
    { name: 'template', description: '- Name of the generator template like for example @asyncapi/html-template or https://github.com/asyncapi/html-template', required: true }
```

this way it reads better in the terminal

derberg (changes requested):
I started review, added come comments to change help messages but then 2 major questions poped up.....I think I'm gonna be a party pooper here:
- I think you folks saw https://github.com/asyncapi/cli/pull/173 and are aware of the direction that we were going there with `generate` command. With `generate types`. Shouldn't we merge the other PR first and add this PR with `generate template` ?
- in code I see custom handling of `file vs url`, and in help message no mention of context which makes me feel we are not reusing global handlers for handling file/url/context, which is not good

Have a look, thanks 🙇🏼

magicmatatjahu:
@derberg 

> I think you folks saw https://github.com/asyncapi/cli/pull/173 and are aware of the direction that we were going there with generate command. With generate types. Shouldn't we merge the other PR first and add this PR with generate template ?

We can even add `generate template` in this PR and then handle mentioned PR. What do you think?

> in code I see custom handling of file vs url, and in help message no mention of context which makes me feel we are not reusing global handlers for handling file/url/context, which is not good

Where do you see that?

Souvikns:
😆 sorry, we are using the `load` function [here](https://github.com/Souvikns/cli/blob/426c3611e9c739aa819ef8408f75c00e559655a0/src/commands/generate.ts#L197). Then @derberg shall we just move the command under `generate template`?

derberg:
> We can even add generate template in this PR and then handle mentioned PR. What do you think?
> Then @derberg shall we just move the command under generate template?

to be fair with contributor, we should first merge modelina, and then this 😉 unless he is ok with your suggestion.

> sorry, we are using the load function [here](https://github.com/Souvikns/cli/blob/426c3611e9c739aa819ef8408f75c00e559655a0/src/commands/generate.ts#L197)

oh sorry, I noticed https://github.com/asyncapi/cli/pull/221/files#diff-8a64c17c2ac6b0f75d2c565e975631f3c831a8eb5b0aa22eaa4cfda960b08920R8-R10 but did not check if you are using it 😞 sorry, please remove this helper

also please update help to make it clear that `asyncapi` argument can also be context

Souvikns:
> to be fair with contributor, we should first merge modelina, and then this 😉 unless he is ok with your suggestion.

Sure @derberg I have no problem waiting, let's merge this after we have merged modelina.

derberg:
can you please fix some DX with your PR in the new `asyncapi generate` command?

```
$ asyncapi generate
 ›   Warning: help is not a asyncapi command.
Did you mean new? [y/n]: n
```

☝🏼 should actually work the way that I out of the box get a display of `help` message with list of commands available in `generate` topic

```
$ asyncapi generate
Generates typed models

USAGE
  $ asyncapi generate COMMAND

COMMANDS
  generate models  Generates typed models
```

Souvikns:
Yeah, it does not support out of the box for topic, I forgot to add the workaround for the `generate` command. 

```
cli on  generator-command is 📦 v0.21.0 via ⬢ v16.14.0 
➜ ./bin/run generate 
Generate typed models or templates

USAGE
  $ asyncapi generate

DESCRIPTION
  Generate typed models or templates

COMMANDS
  generate models    Generates typed models
  generate template  generate whatever you want using templates compatible with AsyncAPI Generator


cli on  generator-command is 📦 v0.21.0 via ⬢ v16.14.0 took 4s 
➜ 

```

derberg:
@Souvikns you have some conflicts with `package-lock.json`. Lemme know when we can do the review

Souvikns:
@derberg it is ready for review 👍🏼

src/commands/generate/template.ts
```diff
@@ -0,0 +1,276 @@
...
+        new ValidationError({
+          type: 'invalid-file',
+          filepath: asyncapi,
+        }),
+        { exit: 1 },
+      );
+    }
+    const generator = new AsyncAPIGenerator(template, output, options);
+
+    CliUx.ux.action.start('Generating template');
```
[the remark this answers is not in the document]
derberg: btw, love the spinner indicator!!!

src/commands/generate/template.ts
```diff
@@ -0,0 +1,276 @@
...
+    output: Flags.string({
+      char: 'o',
+      description: 'directory where to put the generated files (defaults to current directory)',
+      required: true
+    }),
+    'force-write': Flags.boolean({
+      default: false,
+      description: 'force writing of the generated files to given directory even if it is a git repo with unstaged files or not empty dir (defaults to false)'
+    }),
+    'watch-template': watchFlag(
```
derberg, line 64: shouldn't it be watch as it is globally in this CLI?

derberg (changes requested):
lgtm so far 🚀  great progress and initiative 💪🏼 

- we are missing tests, and we should test every flag I think
-  `-o` flag is mandatory, and like in case of other generate commands, it should not, yes, I understand it is not in the generator. It should be contributed as now the DX is not cool, that I get error about missing arguments when running `asyncapi generate template`, and then when I add missing arguments I get error about missing flags, and flags should not be mandatory
- we should not propagate logs from Generator in my opinion, as they have a different style, and colors, so info about `Check out your shiny new generated files at dupa.` or error that for example I should use `--force-write` flag, should come from this CLI
- please update readme 😉 

I also left few comments.

src/commands/generate/template.ts
Souvikns: We are using the global flag https://github.com/asyncapi/cli/blob/master/src/flags.ts, do you mean the message should we use `Enable Watch Mode`.

derberg: I mean the name of the flag, I think it should be `--watch`

derberg:
@Souvikns is there any other way we can test them? We need to increase 🤷🏼

src/commands/generate/template.ts
derberg: it is ok to change flag names as we are migrating and basically creating kinda new CLI, so IMHO better to be consistent with AsyncAPI CLI

derberg:
I'm thinking if `asyncapi generate template` is not confusing:
- `asyncapi generate models` actually generates models
- `asyncapi generate relations` (or whatever name will be when integrating `cupid`) actually generates a relations diagram
- `asyncapi generate template` is actually not generating a template 😄 so maybe `asyncapi generate fromTemplate`?

Souvikns:
> @Souvikns is there any other way we can test them? We need to increase 🤷🏼

Increasing the timeout works couldn't find anything else, maybe we try to reduce the time it takes for generation 🤷🏼‍♂️. Will add more tests if increasing the timeout is alright. 

> asyncapi generate template is actually not generating a template 😄 so maybe asyncapi generate fromTemplate? 

Yeah makes sense we can have `fromTemplate` what do you think @magicmatatjahu 




> -o flag is mandatory, and like in case of other generate commands, it should not, yes, I understand it is not in the generator. It should be contributed as now the DX is not cool, that I get error about missing arguments when running asyncapi generate template, and then when I add missing arguments I get error about missing flags, and flags should not be mandatory 

This was a change requested by @magicmatatjahu https://github.com/asyncapi/cli/pull/221#discussion_r865802887 so we should ask him. Should we revert `-o` to not be required flag. Also, we can make `asyncapi` file path not required as we can use `load` function to pick up from working directory or current context, but `template` arg should be required. 


cc: @magicmatatjahu @derberg

Souvikns:
@derberg can you help me, tests are failing in the windows even though they were running on other platforms, and also sonarcloud is picking up code smells from `minimaltemplate` I added the `sonar-project.properties` but it is still not excluding the folder.

derberg:
@magicmatatjahu if we commit `__transpiled` for test template, will it speed up tests run? I think you told me recently that no.

@Souvikns regarding exclusions from sonar, try removing spaces and using wildcards like in this example -> https://github.com/asyncapi/generator/blob/master/.sonarcloud.properties#L2

magicmatatjahu:
@derberg No, generator will try transpile it again.

derberg:
@Souvikns so yeah, add transpiled dir to git ignore

Souvikns:
I ran the the tests in my windows computer and there `generate template` command is missing, which is wierd 

```
(node:16620) [MODULE_NOT_FOUND] ModuleLoadError Plugin: @asyncapi/cli: [MODULE_NOT_FOUND] require failed 
to load D:\Programs\Asyncapi\cli\lib\commands\generate\template.js: Cannot find module 'D:\Programs\Asyncapi\cli\node_modules\@asyncapi\generator\node_modules\fs.extra\fs.extra'. Please verify that the package.json has a valid "main" entry
module: @oclif/core@1.11.0
task: toCached
plugin: @asyncapi/cli
root: D:\Programs\Asyncapi\cli
See more details with DEBUG=*
 »   Error: Command generate:template not found.
```

Also something related to `fs-extra` @magicmatatjahu do I need to udpate the `@asyncapi/generator`

derberg:
@Souvikns yo, I'm back, you need help? I noticed you reverted `minimaltemplate` from using it in test

Souvikns:
Yeah, tests were failing in windows, after some testing on windows, I found out that the generator command does not work in windows when we are using local template and it throws this error 

```
Error: EISDIR: illegal operation on a directory, symlink 'D:\Programs\Asyncapi\cli\test\minimaltemplate' 
-> 'D:\Programs\Asyncapi\cli\node_modules\@asyncapi\generator\node_modules\undefined
```

derberg:
hmmm, maybe we actually need a `name` in `package.json`? this `undefined` made me think about it. Give it a try

Souvikns:
Even after adding the `name` parameter I still get the same error,After https://superuser.com/questions/1031213/windows-10-symbolic-links-problems I tried to run in adminitrator command prompt and then it ran, So it is clearly some issue with rights, 
image

derberg:
@Souvikns just to be 100% sure, windows tests are failing on both your local and GH Actions? and it is fixable only if on local you run as Admin?

Souvikns:
> @Souvikns just to be 100% sure, windows tests are failing on both your local and GH Actions? and it is fixable only if on local you run as Admin?

Yeah

derberg:
There are always issues with windows... 😞 

I'm confused because when you do `npm install` of a package that is on a disk, symlinks are created only if the folder is outside the project where the `npm install` is ran. In other words, `cli` is a folder where `@asyncapi/generator` is installed under `node_modules` and therefore dependencies of `minimaltemplate` should be located there as well.

> If <folder> sits inside the root of your project, its dependencies will be installed and may be hoisted to the top-level node_modules as they would for other types of dependencies. If <folder> sits outside the root of your project, npm will not install the package dependencies in the directory <folder>, but it will create a symlink to <folder>

So symlink, in this case, should not even be created by npm 😞 or I'm missing something 🤷🏼

Souvikns:
> So symlink, in this case, should not even be created by npm 😞 or I'm missing something 🤷🏼

Since it is not in our package.json I think npm is not creating any symlink. 

I think https://www.npmjs.com/package/@npmcli/arborist is installing `simpletemplate` as a dev dependency to `@asyncapi/generator` but windows is not allowing it to do that.

derberg:
@Souvikns maybe we can work around it by adding a name to the template like `@asyncapi/minimaltemplate` and in `cli` add it as `devDependency`, and point to `./test/minimaltemplate` - this way it is 💯 installed in `node_modules` 🤔 can you give it a try?

Souvikns:
> @Souvikns maybe we can work around it by adding a name to the template like @asyncapi/minimaltemplate and in cli add it as devDependency, and point to ./test/minimaltemplate - this way it is 💯 installed in node_modules 🤔 can you give it a try?


Yeah now it works in my computer
image

Souvikns:
Yeah you can review 👍🏼

derberg:
I'm reviewing after the holidays, so my memory might be wrong 😅  sorry 🙏🏼 

- as far as I remember we agreed `template` to be `fromTemplate`
- I think I left a comment about error handling, that we should not propagate errors from the generator, but wrap ours around. We should not mix CLI and Generator errors
    ```bash
    $ bin/run generate template test/asyncapi.yml @asyncapi/html-template
    Generation in progress. Keep calm and wait a bit... done
    Something went wrong:
    Error: "/Users/wookiee/sources/cli" is in a git repository with unstaged changes. Please commit your changes before proceeding or add proper directory to .gitignore file. You can also use the --force-write flag to skip this rule (not recommended).
        at Generator.verifyTargetDir (/Users/wookiee/sources/cli/node_modules/@asyncapi/generator/lib/generator.js:846:71)
        at processTicksAndRejections (internal/process/task_queues.js:95:5)
        at Generator.generate (/Users/wookiee/sources/cli/node_modules/@asyncapi/generator/lib/generator.js:173:29)
        at Template.generate (/Users/wookiee/sources/cli/lib/commands/generate/template.js:122:13)
        at Template.run (/Users/wookiee/sources/cli/lib/commands/generate/template.js:37:13)
        at Template._run (/Users/wookiee/sources/cli/node_modules/@oclif/core/lib/command.js:67:22)
        at Config.runCommand (/Users/wookiee/sources/cli/node_modules/@oclif/core/lib/config/config.js:268:25)
        at Object.run (/Users/wookiee/sources/cli/node_modules/@oclif/core/lib/main.js:78:5)
    Error: EEXIT: 1
    ```
- Now we have `output` optional, but we are generating files to workdir. We do not put content in terminal. I'm ok to accept such behavior as this is current generator CLI behavior and as thanks to `force-write` nobody can harm themselves by overwriting their work with generation. I'm aware that Generator does not support generation in the way that we could use it to render results to the terminal. The only possible solution would be to use `tmpdir()` and by default, generate files there and then read them to the terminal, but yeah, I'm not so sure. Can you just create a follow-up issue in the Generator, to enable what we need?
- In Help in terminal, all help description should start with upper case
- We should have a test for every flag, and also test bad flag usage and ensuring that proper error is rendered
- In regards to `disable-hook`:
  - we do not have a test
  - I tried to check how it works and I failed, looks like it never works
- In regards to `params`
  - we need test with 2 params provided
  - it is also confusing that current test uses flag `-p=customParam=\'1.0.0\'` and it works, but `-p customParam=\'1.0.0\'` works to. I tried with nodejs-template `-p server=production asyncapiFileDir=my` and was great, and could not figure how to do it if it starts with `-p=`. That was not intuitive. We need to be consistent, support one way or two, as long as help is consistent. Now even help says something different if you compare it with this error message https://github.com/asyncapi/cli/pull/221/files#diff-1d6320fe12a9862589159673bb610b4cd3e8d5cfaf471e29c3a7fc6fce5cf9d8R127
- In regards to `map-base-url` I did not check it as I assume problems I described above are related to all flags that require custom input parser
- We need some examples when running `asyncapi generate fromTempate --help`, complex examples with multiple flags used, and when explaining each flag, example of a flag.
- I was thinking about how we can ensure better in tests that the given flag was applied and works. Now in tests, you only evaluate that a successful message got into `stdout`. Maybe we are missing unit tests for flags parsers (look at my issues with flags like `params`) and unit tests that make sure the given function was invoked with proper values 🤔 . Or we need integration tests: so basically evaluate also if generated file/dir contains (or not) content, based on the selected flag. I'm not sure what is the best approach here. Thoughts?

Sorry 😅 
I'm happy to hop on a call if you need some help

Souvikns:
I am trying to check if the flag parsing is working and I think it is working as expected 

```
./bin/run generate fromTemplate s s -d generate:before=foo,bar generate:after --map-base-url https://schema.example.com/crm/:./test/docs/ -p version=1.0.0 name=souvik
```

```ts
{
  params: { version: '1.0.0', name: 'souvik' },
  disableHooks: { 'generate:before': [ 'foo', 'bar' ], 'generate:after': true },
  mapBaseUrlToFolder: {
    url: 'https://schema.example.com/crm',
    folder: '/Users/postman/Documents/GitHub/cli/test/docs'
  }
}

```

@derberg can you share how you are testing `disable-hook` flag

derberg:
> I found something about map-base-url flag, for some reason [generateFromString](https://github.com/asyncapi/generator/blob/30a9df9c003933b4d2f047b538f49aaac3d31ac4/lib/generator.js#L258) command does not support map-base-url and which is something we are using because we are using CLI's load function for loading the spec file, so currently map-base-url flag does not work at all. What should we do, not use our CLI's load function?

Ah yes, cause we are not using `generateFromURL` as there is no point in doing it as we have our own load. 
I think we should be consistent and use our own load function.  

There are 2 options:
- you just take https://github.com/asyncapi/generator/blob/master/lib/generator.js#L326-L328 and move it to `generateFromString`. I don't see a problem with it. In the end, somebody consciously passes `mapBaseUrlToFolder` to the constructor, so they know what they're doing 🤷🏼 and now the current solution limits/forces them to use `generateFromURL` 🤔 
- you can just make CLI responsible for `getMapBaseUrlToFolderResolver`, not rely on generator 🤔 

> I am struggling to solve this, generator does not have any errors that I can extend, one thing we can do is match output string and create our own errors and use them, or should we wait for https://github.com/asyncapi/community/issues/266
@derberg what do your think?

yeah it will take some time until it is actually used in the generator. `one thing we can do is match output string and create our own errors and use them` is a good option, we just need to have some meaningful fallback, in case error strings change 😄 

@magicmatatjahu @boyney123 any thoughts on above challenges?

README.md
```diff
@@ -99,6 +99,7 @@ COMMANDS
       java          generate the models for Java
       javascript    generate the models for JavaScript
       dart          generate the models for Dart
+    fromTemplate     generate whatever you want using templates compatible with AsyncAPI Generator
```
M3lkior, line 102: Maybe you can remove the doc part 

```
As of now, the @asyncapi/cli only supports validation of the specification file. (This is still under development for more features.)
```
Because now the CLI support other features

M3lkior (changes requested):

Souvikns:
> you can just make CLI responsible for getMapBaseUrlToFolderResolver, not rely on generator 🤔

@derberg I went with this and also removed the stack from the output so now the error messages look something like this 

Screenshot 2022-09-09 at 2 15 12 PM


The messages coming from the generator commands are self explanatory so I didn't changed anything, can you review and let me know if any thing needs to be changed.

README.md
M3lkior: It seems that this line of docs is not removed :)

src/commands/generate/fromTemplate.ts
```diff
@@ -0,0 +1,294 @@
...
+      noOverwriteGlobs: flags['no-overwrite'],
+      mapBaseUrlToFolder: parsedFlags.mapBaseUrlToFolder,
+      disabledHooks: parsedFlags.disableHooks,
+    };
+    const watchTemplate = flags['watch'];
+    const genOption: any = {};
+
+    if (flags['map-base-url']) {
+      genOption.resolve = {resolve: this.getMapBaseUrlToFolderResolver(parsedFlags.mapBaseUrlToFolder)};
+    }
```
Souvikns, line 107: > you can just make CLI responsible for getMapBaseUrlToFolderResolver, not rely on generator 🤔

Not relying on generator for `mapBaseURL` for now.

src/commands/generate/fromTemplate.ts
```diff
@@ -0,0 +1,294 @@
...
+    const generator = new AsyncAPIGenerator(template, output || path.resolve(os.tmpdir(), 'asyncapi-generator'), options);
+
+    CliUx.ux.action.start('Generation in progress. Keep calm and wait a bit');
+    try {
+      await generator.generateFromString(specification.text(), genOption);
+      CliUx.ux.action.stop();
+    } catch (err: any) {
+      CliUx.ux.action.stop('done\n');
+      throw new GeneratorError(err);
```
Souvikns, line 193: Created a new error for generator errors, and removed the error stack only printing the error message 

For Example: 
```
Generator Error: "docs" is in a git repository with unstaged changes. Please commit your changes before proceeding or add proper directory to .gitignore file. You can also use the --force-write flag to skip this rule (not recommended).
```

test/commands/generate/fromTemplate.test.ts
```diff
@@ -0,0 +1,156 @@
...
+
+const cleanup = async (filepath: string) => {
+  rimraf.sync(filepath);
+};
+
+describe('template', () => {
+  test
+    .stdout()
+    .command([...generalOptions, '--output=./test/docs', '--force-write'])
+    .it('should generate html tempalte', (ctx, done) => {
```
derberg, line 22:
```suggestion
    .it('should generate minimal template', (ctx, done) => {
```

derberg (commented):
@Souvikns 3 small suggestions, 

in the meantime I'll do some manual testing

package.json
```diff
@@ -97,7 +100,12 @@
     "topicSeparator": " ",
     "topics": {
       "config:context": {},
-      "config": {}
+      "config": {
+        "description": "CLI config settings"
+      },
+      "generate": {
```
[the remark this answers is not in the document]
Souvikns: These are for the topic help messages https://oclif.io/docs/topics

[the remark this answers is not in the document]
derberg: ah, makes sense, thanks!

test/commands/generate/fromTemplate.test.ts
```diff
@@ -0,0 +1,156 @@
...
+      });
+  });
+
+  describe('install', () => {
+    test
+      .stdout()
+      .command([
+        'generate:fromTemplate',
+        './test/specification.yml',
+        '@asyncapi/html-template',
```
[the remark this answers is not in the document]
Souvikns: the install flag looks into the npm registry for installing templates 

I get this error when I am using `minimaltemplate` 
```
Generator Error: 404 Not Found - GET https://registry.npmjs.org/@asyncapi%2fminimaltemplate - Not found
```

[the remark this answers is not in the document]
derberg: so just change it from `@asyncapi/html-template` to `./test/minimaltemplate`. I checked it on local and works fine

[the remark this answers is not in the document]
Souvikns: Oh yeah it does, I was using `@asyncapi/minimaltemplate` 😄. It still needs the extra timeout to run though.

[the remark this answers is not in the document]
derberg: at least smaller 😄

magicmatatjahu (approved):
LGTM! 🚀

PR #230, opened 2022-02-11
docs: fix missing toc

Souvikns:
**Description**

For some reason, the TOC is not generated in README.md. Just manually creating it in this PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,6 +23,14 @@
 
 <!-- toc -->
 
+- [Requirements](#requirements)
+  * [Installation](#installation)
+- [Usage](#usage)
+- [Contributing](#contributing)
+  * [Set up development environment](#set-up-development-environment)
+  * [Command Structure and Patterns](#command-structure-and-patterns)
+- [Contributors](#contributors)
+
 <!-- tocstop -->
 
 ## Requirements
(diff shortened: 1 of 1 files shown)


magicmatatjahu:
@Souvikns Did you run manually `generate:markdown:toc`? I think that we have some problem on the workflow side and we should try to make it by merging PR.

Souvikns:
> @Souvikns Did you run manually `generate:markdown:toc`? I think that we have some problem on the workflow side and we should try to make it by merging PR.

Yeah, I just ran the command and pushed I figured something was with the workflow because we have a workflow command for this, but was not able to find it out.

magicmatatjahu:
@Souvikns Yeah we have that here https://github.com/asyncapi/cli/blob/master/.github/workflows/if-nodejs-version-bump.yml#L31 I also wonder why it didn't run on the post release workflow 🤔  I can accept that changes because I don't think so that we should wait for that, but please rerun again the markdown-toc generation because maybe there are some changes in the master and ping me, ok? :)

Souvikns:
@magicmatatjahu I just rerun and there aren't any new changes and the branch is up to date with upstream/master

magicmatatjahu (approved):

PR #240, opened 2022-02-21
feat: integrate YAML output in `diff` command

aayushmau5:
**Description**

- integrate YAML output in the `diff` command
Ex.
```
./bin/run diff first.yml second.yml --format=yaml
```


src/commands/diff.ts
```diff
@@ -26,7 +26,7 @@ export default class Diff extends Command {
       char: 'f',
       description: 'format of the output',
       default: 'json',
```
Souvikns, line 28: Should we have `yaml` as default?

aayushmau5: I'm not sure. We can use yaml instead, but not sure what benefit we are getting for it.

Souvikns: Just telling because I see most people use yaml format so keeping it as default could be one less flag as input. I think we should see what most people prefer.

aayushmau5: Good idea. Perhaps I should drop in a message in slack asking which format they prefer :thinking:

Souvikns: Yeah, maybe a small poll about AsyncAPI tools would be benefit us. I think we should have surveys about AsyncAPI, do we have anything like that?

aayushmau5: I don't think we have any survey platform :thinking:

magicmatatjahu: YAML is more preferable if you write AsyncAPI document for humans (as documentation), but if you perform some action then JSON is more preferable. We can make survey on the slack but in my opinion `yaml` should be treated as default. We can also make it in this way that first check input format (json or yaml) and then perform action based on that format.

magicmatatjahu: However it's a output from tool to describe changes, so `json` is good for me as default option 👍🏼

Souvikns (changes requested):
Awesome this looks good, you have two minor code smells, we can merge after you fix them.

derberg (commented):
I think better would be to have `yaml` as default really. Easier to read and more common. JSON for me is considered to be purely technical, while YAML is there to make it easy to read for humans

also, how does it look like after printing, nicely formatted ?

aayushmau5:
> also, how does it look like after printing, nicely formatted ?

It looks kinda like this
image

aayushmau5:
So, should I change the default to YAML? 

Been thinking about it, and I changed my mind. I would prefer YAML instead of JSON now, since the diff output should be readable as compared to being more "technical" JSON.

Souvikns:
> I would prefer YAML instead of JSON now, since the diff output should be readable as compared to being more "technical" JSON.

Same 🤚🏻

aayushmau5:
Alright! Changed the default output format to yaml.

Souvikns (approved):
Looks good to me 👍🏻

PR #245, opened 2022-02-25
fix: patch invalid path issue in studio command

### LINKED ISSUE #128: `asyncapi start studio --file=random-file-that-does-not-exisit` launches studio but then displays a message that the file cannot be watched
#### Describe the bug

If you run the command `asyncapi start studio --file=random-file-that-does-not-exist` (for a file that does not exist) the studio will continue to open (with default spec) and the terminal will give an error.

```
[Error: ENOENT: no such file or directory, open 'exampleasdadsads'] {
  errno: -2,
  code: 'ENOENT',
  syscall: 'open',
  path: 'random-file-that-does-not-exist'
}
```

I think it might be a better developer experience if we just failed and not launch the studio?

ritik307:
**Description**

- Added checker for file path existence in `src/models/Studio.ts`

**Related issue(s)**

Fixes #128


ritik307:
@Souvikns @boyney123 Kindly review the PR 😊

Souvikns:
@ritik307 you have a minor code smell and lint issues, for lint you can just run `npm run lint:fix`

ritik307:
@Souvikns Done! Kindly review 😊

src/models/Studio.ts
```diff
@@ -13,7 +13,15 @@ const messageQueue: string[] = [];
...
 export const DEFAULT_PORT = 3210;
 
+function isValidFilePath(filePath: string): boolean {
+  return existsSync(filePath);
+}
+
 export function start(filePath: string, port: number = DEFAULT_PORT): void {
+  if (!isValidFilePath(filePath)) {
+    console.log('Invalid file path: ', filePath);
+    return;
```
Souvikns, line 23:
```suggestion
    throw new SpecificationFileNotFound(filePath);
```
I think it would be better to throw an error. Since we already have a error for this scenario we should reuse it. 
You can look for it here https://github.com/asyncapi/cli/blob/283de0f36c8bf934b9fd3337bf52ce44932a5e0d/src/errors/specification-file.ts#L9

I think it would be better to throw an error instead

Souvikns (changes requested):

Souvikns:
@ritik307 the test does not stop and it keeps on running, I canceled it because it was running for 41 minutes, I think best would be to close the ws server once you have successfully ran the test.

ritik307:
@Souvikns I wrote the test to run the project locally it doesn't belong to the issue you have raised. I accidentally pushed the test. 😅

Souvikns (approved):
@ritik307 looks good to me 👍🏻 .

boyney123:
Sorry it's taken so long, yeah the code looks OK to me 👍, my only thought would be you could remove the `isValidFilePath` function and just replace it with `existsSync`, but it's not big deal 👍

boyney123 (approved):

src/models/Studio.ts
```diff
@@ -1,4 +1,5 @@
-import { promises as fPromises } from 'fs';
+import { existsSync ,promises as fPromises } from 'fs';
+import {SpecificationFileNotFound} from '../errors/specification-file';
```
derberg, line 2:
```suggestion
import { existsSync, promises as fPromises } from 'fs';
import { SpecificationFileNotFound } from '../errors/specification-file';
```

super not relevant formatting improvement, rest is good

derberg (approved):
LGTM

anyone else?

Souvikns (approved):
👍🏻

Souvikns:
Merging this?

Issue #123, opened 2021-11-03
readme is missing documentation for the new commands 


Souvikns:
Oclif has a way to generate command help documentation if we are going with that we just need to set it up.

fmvilas:
Yeah, just watch out that it might generate documentation of commands separated by colons (`:`) instead of spaces.

BOLT04:
Hi @Souvikns 👋, automating this would be really interesting so that when we add new commands, the docs on the `README` are automatically updated. I had a look on the `package.json` and there is already a script for the readme, are we already using it on GitHub actions?

I'd like to help with this issue if no one is working on it 🙂. I messed around with `oclif readme` command to see the output, and I just want to clarify if the generated docs are correct:
# Usage


```sh-session
$ npm install -g @asyncapi/cli
$ asyncapi COMMAND
running command...
$ asyncapi (-v|--version|version)
@asyncapi/cli/0.12.7 linux-x64 node-v16.13.0
$ asyncapi --help [COMMAND]
USAGE
  $ asyncapi COMMAND
...
```


# Commands


* [`asyncapi config`](#asyncapi-config)
* [`asyncapi config:context`](#asyncapi-configcontext)
* [`asyncapi config:context:add CONTEXT-NAME SPEC-FILE-PATH`](#asyncapi-configcontextadd-context-name-spec-file-path)
* [`asyncapi config:context:current`](#asyncapi-configcontextcurrent)
* [`asyncapi config:context:list`](#asyncapi-configcontextlist)
* [`asyncapi config:context:remove CONTEXT-NAME`](#asyncapi-configcontextremove-context-name)
* [`asyncapi config:context:use CONTEXT-NAME`](#asyncapi-configcontextuse-context-name)
* [`asyncapi new`](#asyncapi-new)
* [`asyncapi start`](#asyncapi-start)
* [`asyncapi start:studio`](#asyncapi-startstudio)
* [`asyncapi validate [SPEC-FILE]`](#asyncapi-validate-spec-file)

## `asyncapi config`

access configs

```
USAGE
  $ asyncapi config

OPTIONS
  -h, --help  show CLI help
```

_See code: [src/commands/config.ts](https://github.com/asyncapi/cli/blob/v0.12.7/src/commands/config.ts)_

## `asyncapi config:context`

```
USAGE
  $ asyncapi config context

OPTIONS
  -h, --help  show CLI help
```

[...]

PR #252, opened 2022-03-09
fix: issue with prepack command  that blocks release

Souvikns:
**Description**

Trying to fix the issue with `npm prepack` command holding of the gh action and not able to publish. 

Still now clear what is causing the issue, since not much log is present, as of now two action run has failed for this 

- https://github.com/asyncapi/cli/runs/5474535742?check_suite_focus=true 
- https://github.com/asyncapi/cli/runs/5391166652?check_suite_focus=true

I tried running the command separately on my fork and it ran not sure why it is having issues here - https://github.com/Souvikns/cli/runs/5469720664?check_suite_focus=true

### DIFF
--- a/package.json
+++ b/package.json
@@ -135,7 +135,7 @@
     "pretest": "npm run build",
     "pretest:coverage": "npm run build",
     "release": "semantic-release",
-    "test": "cross-env TEST=1 CONTEXT_FILENAME=\"./test.asyncapi\" CONTEXT_FILE_PATH=\"./\" nyc --extension .ts mocha --require ts-node/register --require test/helpers/init.js --reporter spec --timeout 10000 \"test/**/*.test.ts\"",
+    "test": "cross-env TEST=1 CONTEXT_FILENAME=\"./test.asyncapi\" CONTEXT_FILE_PATH=\"./\" nyc --extension .ts mocha --require ts-node/register --require test/helpers/init.js --reporter spec --timeout 10000 \"test/**/*.test.ts\""
   },
   "types": "lib/index.d.ts"
 }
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -131,7 +131,7 @@
     "lint": "eslint --max-warnings 0 --config .eslintrc .",
     "lint:fix": "eslint --max-warnings 5 --config .eslintrc . --fix",
     "postpack": "rimraf oclif.manifest.json",
-    "prepack": "rimraf lib && tsc -b && oclif manifest && oclif readme",
+    "prepack": "npm run build && oclif manifest && oclif readme",
     "prepublishOnly": "npm run build",
```
derberg, line 135: do we need this if we build it in `prepack` stage?

Souvikns: Yeah since we are building in `prepublish` stage which is run before prepack we don't need this. Though we are also running `oclif readme` in `version` so we might now need that too 

https://github.com/asyncapi/cli/blob/9fecc14b1f47ae4df238d9ef9cc4778a869d9e5c/package.json#L140

also I think this is why the TOC is not working

derberg: imho:

- we should only have `prepublishOnly` that runs: `npm run build && oclif manifest`
- `version` script is not needed, can be removed, same `prepack`. Same with `readme` script I guess
- `oclif readme` should be added to `generate:assets`. Although I do not really know what `oclif readme` is actually doing 😆 But if it is doing some important update to readme, it should be in `generate:assets` script. [This workflow](https://github.com/asyncapi/cli/blob/master/.github/workflows/if-nodejs-version-bump.yml#L31) will make sure readme is updated in GH repo. We can also run `generate:assets` in `prepublishOnly` script if we need also some magic updates in readme to show up in NPM package readme view

makes sense?

Souvikns: I have removed `oclif readme`. It creates documentation which is handled here https://github.com/asyncapi/cli/issues/123

derberg (approved):

Souvikns:
I have deleted the `0.16.0` tag and will merge this PR now, hope this fixes the issue 🤞🏻

PR #254, opened 2022-03-10
fix: move over to forked oclif

Souvikns:
**Description**
Since `oclif manifest` command didn't have any way to get logs, this PR is installing a fork version of oclif and adding logs so that we can understand what is going on with the command. 

Forked repo - https://github.com/Souvikns/oclif
adding logs - https://github.com/Souvikns/oclif/blob/main/src/commands/manifest.ts

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,12 +59,12 @@
     "markdown-toc": "^1.2.0",
     "mocha": "^9",
     "nyc": "^14.1.1",
-    "oclif": "github:Souvikns/oclif#dd404d08bcbb4862adbc1fd4d296c7a83f5cea50",
     "rimraf": "^3.0.2",
     "semantic-release": "^17.4.3",
     "ts-node": "^10.4.0",
     "tslib": "^2.3.1",
-    "typescript": "^4.4.3"
+    "typescript": "^4.4.3",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -59,12 +59,12 @@
     "markdown-toc": "^1.2.0",
     "mocha": "^9",
     "nyc": "^14.1.1",
-    "oclif": "^2.5.0",
+    "oclif": "github:Souvikns/oclif#main",
     "rimraf": "^3.0.2",
```
Souvikns, line 63: it says https://github.com/asyncapi/cli/runs/5496164139?check_suite_focus=true#step:9:12 but we have `rimraf`

derberg:
@Souvikns you definitely do not want to commit node_modules 😅 as it is just for testing really, can you just publish fork to your private npm? in linked issue folks say it helped to workaround the issue

Souvikns:
@derberg I released a new npm library with all the logs, once we are done with this I will that library.

derberg (approved):
:shipit: 

btw, in a few days NPM will not allow you to delete the library. This is to prevent bad behavior in dependency manager. But yeah, not a problem to have it there. Might be you will need it in the future 😄

Souvikns:
Now I will delete the tag and merge this PR.

Issue #219, opened 2022-02-08
Add `asyncapi bundle` command


derberg:
> This issue is for https://www.openforce.tech/ participants. If you do not participate in this even please pick up other issue from this repo, there is plenty of help needed 😄 

#### Reason/Context

We need to enable the community to use all AsyncAPI tools in CLI if only possible -> https://github.com/asyncapi/bundler

#### Description

- add new command `asyncapi bundle` that uses https://github.com/asyncapi/bundler and enable users to bundle multiple AsyncAPI files into one
- user should be able to pass multiple files to the bundler `asyncapi bundler <context-1/filepath1/url1> <context-2/filepath2/url2> <context-3/filepath3/url>3` (getting file from context or filepath or URL is already provided in  CLI, you just use `load` function)

> This issue is for https://www.openforce.tech/ participants. If you do not participate in this even please pick up other issue from this repo, there is plenty of help needed 😄 

@Souvikns pinging you so you are aware 😄

ravikrshn:
I would like to work on this issue.

AMANKANOJIYA:
I would like to work on this issue as part of Openforce 2022

sudoshreyansh:
I want to work on this issue as part of OpenForce 2022

amanbedi1:
HI @derberg I am Amanpreet  a contributor to OpenForce 2022. I would like to work on this issue. I would be making a PR as soon as I am done with resolving the issue. Thank you.

amanbedi1:
I just want to ask in this command the first file passed to command should be  base file or not.

Souvikns:
> I just want to ask in this command the first file passed to command should be base file or not. 

`bundler` takes an array of files but it does not read the files, so you have to read and pass in the content. And `base` is an option that you can pass. `base` is for when you want to preserve some `schema` objects while you are merging multiple separate specifications. 



> and there should also be an option like --output that user could use to specify where to output bundled doc as a file. Thoughts? 

Yeah actually when `bundler` had a CLI it had a `--output` but since it is library now it just returns a document object and in the CLI interface we can implement the `--output`. My thought behind this was since `server-api` will also integrate `bundler` it would be easier to just bundler and then send the final document over REST.

amanbedi1:
Ok so I think I just have to output bundled file in terminal or ```--output```  flag  specify where to output bundled doc as a file.
I will try to resolve it.

derberg:
[...]

PR #260, opened 2022-03-18
feat: bundle ASYNCAPI Command added

amanbedi1:
**Description**
  

- ```bundle``` asyncapi cli command added using ```@asyncapi/bundle``` package.
- New ```--output``` flag also added.

- This commands take multiple files and bundled to provide output to terminal or redirect to file using  ```--output```  flag.
- Command structure:  
    ```bundle <file1.yml> <file2.yml> <file3.yml>```  

    ```bundle <file1.yml> <file2.yml> <file3.yml> --output=output.yml```  

    ```bundle <file1.yml> <file2.yml> <file3.yml> -o output.yml```

**Related issue(s)**
<a href="https://github.com/asyncapi/cli/issues/219">#219</a>

### DIFF
--- a/src/commands/bundle.ts
+++ b/src/commands/bundle.ts
@@ -1,0 +1,71 @@
+import Command from '../base';
+import { Flags } from '@oclif/core';
+import { load } from '../models/SpecificationFile';
+import { outputFlag } from '../flags';
+import path  from 'path'; 
+import fs from 'fs';
+// @ts-ignore
+import yaml from 'js-yaml';
+// @ts-ignore
+import bundler from '@asyncapi/bundler';
+
+export default class Bundle extends Command{
+    static description = 'It enables users to bundle/merge specification into one';
+
+    static flags={
+        help: Flags.help({char : 'h'}),
+        output: outputFlag,
+    };
+    
+    // To input multiple files by disabling argument validation
+    static strict = false;
+
+    async run(){
+        const {argv,flags} = await this.parse(Bundle);
+        const specFiles = await loadSpecFilesfromFilePaths(argv);
+        const documents = await bundler(specFiles);
+        
+        // checking if output flag is passed or not as output flag is treated as object
+        if(Object.keys(flags).length === 0 && flags.constructor === Object){
+            this.log(documents.yml());
+        } else{
+            const outputFileLocation = flags['output'];
+            const outputFileExtension = path.extname(outputFileLocation!);
+    
+            if(outputFileExtension ==='.yaml' || outputFileExtension ==='.yml'){
+                fs.writeFile(outputFileLocation!, yaml.dump(documents.yml()), (err) => {
+                    if (err) {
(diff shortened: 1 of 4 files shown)


Souvikns:
@amanbedi1 you have lint issues, you can run `npm run lint:fix` to fix all the auto fixable lint errors.

amanbedi1:
@Souvikns actually there is linting error of `@ts-ignore` so how to resolve this.

src/commands/bundle.ts
```diff
@@ -0,0 +1,67 @@
+import Command from '../base';
+import { Flags } from '@oclif/core';
+import { load } from '../models/SpecificationFile';
+import { outputFlag } from '../flags';
+import path from 'path'; 
+import fs from 'fs';
+// @ts-ignore
+import yaml from 'js-yaml';
+// @ts-ignore
```
Souvikns, line 9:
```suggestion
// eslint-disable-next-line @typescript-eslint/ban-ts-comment
// @ts-ignore
```
One thing you can do is this ⏫ for now.

src/commands/bundle.ts
```diff
@@ -0,0 +1,67 @@
...
+}
+
+// utility function to return an array of loaded files from their file paths.
+async function loadSpecFilesfromFilePaths(filePaths:string[]) {
+  const specFilesPromises = filePaths.map(async filePath => {
+    const specFile = await load(filePath); 
+    return specFile.text();
+  }); 
+  return await Promise.all(specFilesPromises);
+}
```
Souvikns, line 67: We have a function for this that you can reuse - https://github.com/asyncapi/cli/blob/ab06d7a57349a382be66cd594f9f22152803440c/src/models/SpecificationFile.ts#L88 

Also you don't have to read input from argv, instead, you can use turn off strict mode and use args you can read this here https://oclif.io/docs/args by doing `static strict = false`. This way you still generate meaningful documentation in help. 

```
It enables users to bundle/merge specification into one

USAGE
  $ asyncapi bundle [-h] [-o <value>]

FLAGS
  -h, --help            Show CLI help.
  -o, --output=<value>  Path to the output file

DESCRIPTION
  It enables users to bundle/merge specification into one
```

Right now it is a bit confusing on where to pass in the file paths.

Souvikns (changes requested):
@amanbedi1 good job 👍🏻 left some small comments 

The rest of the lint errors are sonarjs errors I would suggest you can just refactor your code as per the rules.

src/commands/bundle.ts
amanbedi1: Ok I will do required changes.

derberg:
@amanbedi1 do you plan to continue working on it? Open Force ended but since you started during the event, it is ok if you complete work later

derberg:
I think @amanbedi1 decided to discontinue, let me close, we can always reopen or new PR can be created

PR #272, opened 2022-04-26
docs: use the generic CONTRIBUTING.md

KhudaDad414:
### Description
Since we are using an exact copy of the CONTRIBUTING.md file from the `.github` repo, I guess it makes sense to use that guide instead (like other tooling repos).

as a bonus point, we won't have [broken links](https://github.com/asyncapi/cli/runs/6143298395?check_suite_focus=true) in this repo after removing this file. 😀


Souvikns (approved):
😄 we forgot to update this, thanks @KhudaDad414

PR #288, opened 2022-05-22
fix: wrong exit code for validation error

### LINKED ISSUE #285: validate command returns 0 exit code when validation error occurs
#### Describe the bug
It was expected that when a validation error occurs when running the validate command, a non-zero exit code would be used. Instead, it returns 0.

#### How to Reproduce
```
> touch asyncapi.yml ; npx @asyncapi/cli validate asyncapi.yml ; echo $? ; rm asyncapi.yml
ValidationError: Document can't be null or falsey.
0
```

#### Expected behavior
A non-zero error code would be used.

peter-rr:
**Description**

- Proposal to fix the wrong exit code when validation error occurs.

**Related issue(s)**

- Fixes #285


derberg (commented):
@Souvikns hey man, need your help as I don't remember what was the purpose of this `base.ts`. Is it a default behavior in case of failing command, that we always catch the error and log it? if I'm right, why do we `console.error` instead of `throw new Error`?

Souvikns:
`base.ts` was like a base class that only implements the error handling. We kind of followed this, https://oclif.io/docs/error_handling#error-handling-in-the-catch-method, and if we throw the error oclif has a global error handler which would handle it, but it is for stuff like `command not found` etc. So we are doing `console.error` just to print the error messages.

We can have a custom error handler in the `bin/run` file as per https://oclif.io/docs/error_handling#binrun-catch-handler. Don't know what would be better though.

src/base.ts
```diff
@@ -3,5 +3,6 @@ import {Command} from '@oclif/core';
 export default abstract class extends Command {
   async catch(e: any) {
     console.error(`${e.name}: ${e.message}`);
+    process.exitCode = 1;
```
Souvikns, line 6: Yeah, this solves it 👍🏼 , but wait for @derberg's review.

derberg: >  and if we throw the error oclif has a global error handler which would handle it, but it is for stuff like command not found

if this is only for missing commands, then I don't think it will do the trick. Related issue talks about exit code from `asyncapi validate` when validation fails. Shouldn't it be done in `validate.ts` then?

magicmatatjahu: I think that https://oclif.io/docs/commands#thiserrormessage-string--error-options-code-string-exit-number-ref-string-suggestions-string will be better :)

peter-rr: @Souvikns Yeah, that solutions works fine 👌  I also tried to apply what @magicmatatjahu proposed and it works as well. The output is not exactly the same though. I mean, when the line below is added to the `catch`:

```
this.error(`${e.name}: ${e.message}`, {exit: 1});
```
Then we get the output as follows:
```
 ›   Error: ValidationError: Document can't be null or falsey.
1
```
Whereas by applying the first solution proposed, the output is:
```
ValidationError: Document can't be null or falsey.
1
```

I guess we should opt for the solution already included in this PR, since the output displayed by the alternative solution seems to be redundant . What do you think, guys? 🤔

magicmatatjahu: @peter-rr @magicmatatjahu 

I would actually prefer the solution that is described in the ocliff documentation (the first output), the `Error...` display is not a problem for me.

Souvikns: I also think we should use the solution described in oclif, and we can even change the error message and not have the `Error...` display. 

I think following https://oclif.io/docs/error_handling#binrun-catch-handler we can just write our own custom error handler function in the `bin/run` file.

magicmatatjahu: I opt for that :) @peter-rr Could you propagate suggestion? :)

peter-rr: Sure! Let me work on it 💪

peter-rr: After applying the solution as described in oclif and also adding the error handling to the `bin/run` file, the output remains the same with the `Error...` display.

I needed to install the `@oclif/error` module since it wasn't included yet in package.json, and finally found out this libray has been replaced by `@oclif/core` 👉  https://github.com/oclif/errors. So I guess there must be any conflict with those libraries and I'm missing something when editing the `bin/run` file 🤔 Any clue about it? @Souvikns @magicmatatjahu

Souvikns: I don't think you need to install `oclif/error` anymore. From newer versions everything is under `oclif/core`. 😄 I think they did not update their documentation, If you see the latest changes in our branch you will see 

```js
#!/usr/bin/env node

const oclif = require('@oclif/core')

oclif.run().then(require('@oclif/core/flush')).catch(require('@oclif/core/handle'))
```

see there is `@oclif/core/handle`, you don't have to install anything else. 

You can just write a custom handler function like 

```js
#!/usr/bin/env node

const oclif = require('@oclif/core')

oclif.run().then(require('@oclif/core/flush')).catch(err => {
  console.error(err.message)
})
```

peter-rr: After applying the latest changes proposed, I'm getting the following output:
```
ValidationError: Document can't be null or falsey.

1
```
However, 10 tests from several commands are failing due to the `Error: ` title is now added to every error and that's something not expected by the tests. Should I update those tests to make them pass? Or is it possible to remove or ignore somehow that `Error: ` title to avoid the tests fail? 🤔

Souvikns: I tested and I think this is coming from `dev.cmd` file in `bin` folder which is where I guess the tests are running. If you do `./bin/run validate wrong/path` then you get `error loading AsyncAPI document from file: ./something.yaml is an invalid file path` but when we do `./bin/dev validate wrong/path` we get 

```
Error: error loading AsyncAPI document from file: ./something.yaml is an invalid file path
    at Object.error (/Users/postman/Documents/GitHub/cli/node_modules/@oclif/core/lib/errors/index.js:29:15)
    at Validate.error (/Users/postman/Documents/GitHub/cli/node_modules/@oclif/core/lib/command.js:90:23)
    at Validate.catch (/Users/postman/Documents/GitHub/cli/src/base.ts:5:10)
    at Validate._run (/Users/postman/Documents/GitHub/cli/node_modules/@oclif/core/lib/command.js:71:29)
    at async Config.runCommand (/Users/postman/Documents/GitHub/cli/node_modules/@oclif/core/lib/config/config.js:226:25)
    at async Object.run (/Users/postman/Documents/GitHub/cli/node_modules/@oclif/core/lib/main.js:73:5)
```

Souvikns:
```ts
  .catch((err) => {
    const oclifHandler = require('@oclif/core/handle');
    console.error(err.message);
    return oclifHandler(err.message);
  });
```

This code is redundant, the control is not reaching here since `this.error` is a logging command as per https://oclif.io/docs/commands#command-methods. 

The control reaches here when we throw the error in the `catch` function. 

this `this.error` function modifies the log quite a lot, This is how it looks

```
 ›   Error: error loading AsyncAPI document from file: ./s.yml is an invalid file path
```

Now if `console.error` and `this.error` is kind of same we should be able to have something like this - 

```ts
export default abstract class extends Command {
  async catch(e: any) {
    console.error(`${e.name}: ${e.message}`);

    this.exit(1);
  }
}
```

This kind of works in terminal you run run `./bin/dev validate ./wrong.yaml`, it gives 

```
error loading AsyncAPI document from file: ./s.yml is an invalid file path
```

but when you run `npm run test` we get 

```
10) validate
       with no arguments
         throws error message if no context file exists:
     Error: EEXIT: 1
      at Object.exit (node_modules/@oclif/core/lib/errors/index.js:22:11)
      at Validate.exit (node_modules/@oclif/core/lib/command.js:82:23)
      at Validate.catch (src/base.ts:1:1345)
      at Validate._run (node_modules/@oclif/core/lib/command.js:71:29)
      at async Config.runCommand (node_modules/@oclif/core/lib/config/config.js:226:25)
      at async Object.run (node_modules/@oclif/test/lib/command.js:21:13)
      at async Context.run (node_modules/fancy-test/lib/base.js:68:25)

```

---
> I also found this, https://github.com/oclif/oclif/issues/229 There is a open issue in `oclif` and there is no fix for it yet.

peter-rr:
Hey @Souvikns, I wonder if we could opt again for the first solution you proposed. I mean, the following one for the `base.ts` file:
```
export default abstract class extends Command {
  async catch(e: any) {
    console.error(`${e.name}: ${e.message}`);
    process.exitCode = 1;
    }
  }
```
The output we get would be as follows:
```
ValidationError: Document can't be null or falsey.
1 
```
And there are only 4 failing tests when this solution is applied, all of them related to missing context, which I guess it might be easier to solve: 
image
image

What do you think about this? Still a possible solution? 🤔

peter-rr:
@Souvikns Also I'd like to solve the conflicts in this branch, related to `package-lock.json` file, so I can fetch the latest changes from master, but I don't have write access to resolve those conflicts. Do you know how I should proceed in this case? Thanks, man :)

Souvikns:
Yeah Everything is working, we are getting the error code as expected just need the conflicts resolved. 

Screenshot 2022-09-22 at 6 37 29 PM

derberg:
@Souvikns I fixed package-lock and also increased timeout in tests, as release is still failing. Please approve

Souvikns (approved):

PR #292, opened 2022-05-27
feat: extend error handling in validate command

peter-rr:
**Description**

- Error handling added when loading a spec file for `validate` command, since this process was not handled before in case of error.

**Related issue(s)**

- See also #288

### DIFF
--- a/src/commands/validate.ts
+++ b/src/commands/validate.ts
@@ -5,7 +5,6 @@ import { ValidationError } from '../errors/validation-error';
 import { load } from '../models/SpecificationFile';
 import { specWatcher } from '../globals';
 import { watchFlag } from '../flags';
-import { SpecificationFileNotFound } from '../errors/specification-file';
 
 export default class Validate extends Command {
   static description = 'validate asyncapi file';
@@ -22,19 +21,14 @@ export default class Validate extends Command {
   async run() {
     const { args, flags } = await this.parse(Validate); //NOSONAR
     const filePath = args['spec-file'];
-
     const watchMode = flags['watch'];
 
     const specFile = await load(filePath);
     if (watchMode) {
-      specWatcher({spec: specFile, handler: this, handlerName: 'validate'});
+      specWatcher({ spec: specFile, handler: this, handlerName: 'validate' });
     }
-    try {
-      const specFile = await load(filePath);
-      if (watchMode) {
-        specWatcher({ spec: specFile, handler: this, handlerName: 'validate' });
-      }
-
+    
+    try {  
       if (specFile.getFilePath()) {
         await parser.parse(specFile.text());
         this.log(`File ${specFile.getFilePath()} successfully validated!`);
@@ -42,15 +36,11 @@ export default class Validate extends Command {
         await parser.parse(specFile.text());
         this.log(`URL ${specFile.getFileURL()} successfully validated`);
       }
-    } catch (e) {
-      if (e instanceof SpecificationFileNotFound) {
-        this.error(e);
-      } else {
-        throw new ValidationError({
-          type: 'parser-error',
-          err: e
-        });
-      }
+    } catch (error) {
+      throw new ValidationError({
+        type: 'parser-error',
+        err: error
+      });
     }
   }
 }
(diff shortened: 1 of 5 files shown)


peter-rr:
After applying some tweaks to solve the failing tests, I've found out that the errors concerning the `load(specFile)` function were already handled in the original code by accessing the `ErrorLoadingSpec` class in `specification-file.ts` file: https://github.com/asyncapi/cli/blob/master/src/errors/specification-file.ts#L36

Then correct me if I'm wrong but I think the improvement proposed is not actually an improvement, and therefore may be discarded 🤔

Souvikns:
I don't fully understand you, we have custom errors already but we don't have support for exit codes. You can just use all the previous error messages defined and add `exit` code in `this.error` since oclif supports 
```
this.error(message: string | Error, options?: {code?: string, exit?: number, ref?: string; suggestions?: string[];})
```

peter-rr:
@Souvikns Yeah, I guess I didn't express myself clearly 😅 What I meant from the beginning with this PR was to handle the errors coming from `load(specFile)` due to I thought those errors were not handled since that function is outside the _try-catch_ block. Then after trying to solve the failing tests I found out the errors happening when loading the spec are already handled by the _try-catch_ block, even though the `load(specFile)` function is outside. Is it clear now? :)

Souvikns:
Oh got it 👍🏼. Then I think this PR is not needed, you can close it if you want.

fmvilas:
I think this PR still makes sense. The error was handled but incorrectly. It was passing an undefined value to the parser and therefore the error we were getting was that the document was invalid but that's not true. The document is not invalid because the document doesn't even exist 😅

Souvikns:
When the file does not exist, we get 
```
error loading AsyncAPI document from file: ./spec.json is an invalid file path
```

I think this error is handled by the `load` function. I think we need to find a better way to handle errors right now it is a bit confusing.

src/commands/validate.ts
```diff
@@ -21,26 +22,34 @@ export default class Validate extends Command {
...
         this.log(`URL ${specFile.getFileURL()} successfully validated`);
       }
-    } catch (error) {
-      throw new ValidationError({
-        type: 'parser-error',
-        err: error
-      });
+    } catch (e) {
+      if (e instanceof SpecificationFileNotFound) {
+        this.error(new ValidationError({
```
fmvilas, line 43: I think you should not trigger a `ValidationError` but instead a [SpecificationFileNotFoundError](https://github.com/asyncapi/cli/blob/master/src/errors/specification-file.ts#L9). This way, we'll get the right message when the file doesn't exist.

src/commands/validate.ts
```diff
@@ -29,18 +30,30 @@ export default class Validate extends Command {
...
+          type: 'invalid-file',
+          filepath: filePath
+        }));
```
fmvilas, line 50: This is what I meant in my previous comment:

```suggestion
        this.error(new SpecificationFileNotFound(filePath));
```

But since `e` is already a `SpecificationFileNotFound` error, you can simply forward the error as is:

```suggestion
        this.error(e);
```

fmvilas (changes requested):

src/errors/specification-file.ts
```diff
@@ -34,15 +34,15 @@ export class ErrorLoadingSpec extends Error {
...
     if (from === 'file') {
       this.name = 'error loading AsyncAPI document from file';
-      this.message = `${param} is an invalid file path`;
+      this.message = `${param} file does not exist.`;
     }
     if (from === 'url') {
-      this.name = 'error loading AsyncAPI docuement from url';
-      this.message = `${param} is an invalid url`;
+      this.name = 'error loading AsyncAPI document from url';
+      this.message = `${param} url does not exist.`;
```
fmvilas, line 41: We can't conclude that the URL does not exist but that we can't access it. Maybe something like this:

```suggestion
      this.message = `Failed to download ${param}.`;
```

fmvilas (changes requested):
It's cool you removed the extra try/catch on the src/commands/validate.ts file. It looks like in the end it was easier 🙌

fmvilas (approved):

Souvikns (approved):

PR #295, opened 2022-06-21
feat: add support for diff markdown output

aayushmau5:
**Description**

- This PR adds support for Markdown output by diff that has been added recently in the diff library.
- Currently, the markdown output is printed to stdout


aayushmau5:
Created a draft PR for now. Have to make the tests better.

src/commands/diff.ts
```diff
@@ -26,14 +26,20 @@ export default class Diff extends Command {
       char: 'f',
       description: 'format of the output',
       default: 'yaml',
-      options: ['json', 'yaml', 'yml'],
+      options: ['json', 'yaml', 'yml', 'markdown', 'md'],
```
Souvikns, line 29:
```suggestion
      options: ['json', 'yaml', 'yml', 'md'],
```

We should have just the `md` option.

Souvikns (changes requested):

Souvikns (approved):

src/commands/diff.ts
```diff
@@ -159,17 +171,31 @@ export default class Diff extends Command {
...
+}
+
+/**
+ * A generic output function for diff output
+ * @param diffOutput The diff output data
+ * @param outputType The output format requested by the user
+ * @returns The output(if the format exists) or a message indicating the format doesn't exist
+ */
+function genericOutput(diffOutput: AsyncAPIDiff, outputType: string) {
+  if (outputType === 'breaking') {
```
boyney123, line 189: Hey,

Just a minor thing, but do you think it's worth adding a switch statement vs the if conditions? Would make extending easier in the future?

Souvikns (approved):

magicmatatjahu (changes requested):

src/commands/diff.ts
```diff
@@ -26,14 +26,20 @@ export default class Diff extends Command {
...
       description: 'type of the output',
       default: 'all',
       options: ['breaking', 'non-breaking', 'unclassified', 'all'],
     }),
+    markdownSubtype: Flags.string({
+      char: 's',
+      description: 'output type of changes in markdown format',
+      default: 'json',
+      options: ['json', 'yaml', 'yml']
+    }),
```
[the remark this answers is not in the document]
derberg: actually I'm not 100% sure what `markdownSubtype` is, why it is not enough to just add additional `md` `format` as a supported one?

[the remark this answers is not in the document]
aayushmau5: @derberg `markdownSubtype` refers to the format of changes you want in markdown code block. 

Let's say a user removed some part of their AsyncAPI document, and uses diff to find the changes. In markdown format, we show the removed changes in a code block. This code block can either show the removed snippet as JSON or YAML.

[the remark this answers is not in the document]
aayushmau5: @magicmatatjahu Removed the `s` short flag.

@derberg `why it is not enough to just add additional md format as a supported one?`, I'm not sure what you mean. Can you clarify?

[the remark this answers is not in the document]
derberg: oh, sorry for being a turtle 🐢
sorry but I'm still not getting it :( 

`the format of changes made to AsyncAPI document, when diff is generated as markdown`

I just do not get what is the difference between ☝🏼 and output format

can you share some example command that I can try locally and actually see it, so then I get it 😄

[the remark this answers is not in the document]
aayushmau5: So, output format refers to different ways of showing the diff data to users. Such as in `json`, `yaml` or `markdown` form.

Now, `markdownSubtype` is a case for `markdown` **output format**. For markdown output, the `changes` data is show in code blocks.

Example:
### Non-breaking changes

```json
{
	"removed": [...],
	"added": [...]
}
```

The `markdownSubtype` key controls this^. We can show the changes in `json` or in `yaml`.

For example,
```
diff --outputFormat=md --markdownSubtype=json
```
Will return:
### Non-breaking changes

```json
{
	"removed": [...],
	"added": [...]
}
```

And

```
diff --outputFormat=md --markdownSubtype=yaml
```

Will return:
### Non-breaking changes

```yaml
removed: [...]
added: [...]
```

@derberg I hope this clears your doubt. Let me know if you need to know something else.

To be honest, I think the way I described `markdownSubtype` in this PR is not that clear. We definitely need to improve upon the wording.

src/commands/diff.ts
```diff
@@ -125,13 +132,16 @@ export default class Diff extends Command {
         {
           override: overrides,
           outputType: outputFormat as diff.OutputType, // NOSONAR
+          markdownSubtype: markdownSubtype as diff.MarkdownSubtype
```
derberg, line 135: I'm not 100% sure if this is correct. I can't figure what is recommended here https://clig.dev/

basically, if I do `asyncapi diff a.yml b.yml --type=json --markdownSubtype=yml` it will work and `markdownSubtype` is just ignored by diff library. 

now, is this right:
- throwing error that `markdownSubtype` is not supported in `json` type is kinda blocking for the user
- ignoring the type is good as diff still works, it's just that flag has no effect

maybe we should just display some kind of `info`?

we have similar case https://github.com/asyncapi/cli/blob/master/src/commands/generate/models.ts but not the same as there it makes sense to "error the user" because critical flag is missing

wdyt?

cc @Souvikns @magicmatatjahu

derberg (changes requested):
@aayushmau5 🙏🏼

src/commands/diff.ts
aayushmau5: I think displaying a"warning"(i.e. a text in yellow color) would be good. Like "Warning: Hey, this option won't work with the given type".

I think throwing an error is bad in this case. Like you said, it will be blocking for users.

derberg: ok, can you make sure we display the warning then?

aayushmau5: @derberg Sorry for the long delay on this PR. Getting back to it now.

> ok, can you make sure we display the warning then?

How does this warning look?

image

derberg: lgtm, maybe add also the value provided by user that will not work

src/commands/diff.ts
```diff
@@ -26,14 +26,19 @@ export default class Diff extends Command {
...
+      options: ['json', 'yaml', 'yml', 'md'],
     }),
     type: Flags.string({
       char: 't',
       description: 'type of the output',
       default: 'all',
       options: ['breaking', 'non-breaking', 'unclassified', 'all'],
     }),
+    markdownSubtype: Flags.string({
+      description: 'the format of changes made to AsyncAPI document. It works only when diff is generated using md type. For example, when you specify subtype as json, then diff information in markdown is dupmed as json structure.',
```
derberg, line 38:
```suggestion
      description: 'the format of changes made to AsyncAPI document. It works only when diff is generated using md type. For example, when you specify subtype as json, then diff information in markdown is dumped as json structure.',
```

derberg:
@aayushmau5 some tests related to diff are failing, don't think it is random stuff

derberg:
@aayushmau5 yo, I suggest resolving conflicts and updating to latest master as lots of random issues in flaky tests were fixed, and all should be good here after update

aayushmau5:
@derberg will do. Was fixing the tests, but some tests were timing out locally. Hopefully the latest master will have this fix.

Souvikns:
@aayushmau5 can you let me know if you are facing the same issues with the tests now?

aayushmau5:
@Souvikns fixed all the tests. No sure what to do with sonar. As you can see, I spent the last 2 hours fighting with it xD

sonar-project.properties
```diff
@@ -1 +1 @@
-sonar.exclusions=test/minimaltemplate/**/*
\ No newline at end of file
+sonar.exclusions=test/minimaltemplate/**/*,test/commands/diff.test.ts
```
derberg, line 1: yeah, I vote for disabling sonar for all tests by default, it is causing confusion only

```suggestion
sonar.exclusions=test/**/*
```

aayushmau5: This didn't disable sonar :(

derberg (commented):
you need to rename sonar config file to `.sonarcloud.properties`, the other is somehow ignored now

derberg (approved):
LGTM 🚀 

@Souvikns please have a look

magicmatatjahu (approved):

PR #300, opened 2022-06-29
fix: make sure release assets are uploaded to latest release

derberg:
So:
- brew automation works https://github.com/Homebrew/homebrew-core/pull/104709
- assets upload (pkg for example) works too https://github.com/asyncapi/cli/releases/tag/0.19.5 but as you can see, it created wrong release instead of using latest. Also wrong bot is used

This PR fixes upload to work on push, and not release because when we have release event, package.json is not yet updated in the repo

### DIFF
--- a/.github/workflows/upload-release-assets.yml
+++ b/.github/workflows/upload-release-assets.yml
@@ -1,13 +1,16 @@
 name: Upload custom assets to GitHub release
 
-on: 
-  release:
-    types: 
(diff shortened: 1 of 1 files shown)


magicmatatjahu (approved):

derberg:
changed to `ci: ` as no need to run another release, we will make sure all works after modelina related PR is merged

Issue #161, opened 2021-12-14
We need a local server for mocking our tests. 


Souvikns:
#### Reason/Context
> I see that @fmvilas already reviews, so from my point of view I only suggest to log a followup issue, something good for the first contribution, to make sure tests do not rely on external resources. Basically, external resources are risky to use in tests and can cause tests to be flaky, like randomly failing just because the external resource is not available. So follow up should describe that tests should actually start independent server hosting asyncapi file, like we do in parser -> https://github.com/asyncapi/parser-js/blob/master/package.json#L20

This issue is related to https://github.com/asyncapi/cli/pull/110#discussion_r768405925

derberg:
we should probably have a look at this one with higher priority, maybe this is the reason for some timeouts

chinma-yyy:
Hey @Souvikns and @derberg , from what i understood this issue is if this remote file cannot be loaded the test may fail and be flaky. Instead we want our own server locally through which we can serve the file through the server but the local file https://github.com/asyncapi/cli/blob/f2633cf9675f4ec135b2ca2e97064ddd1a97e9a7/test/commands/generate/models.test.ts#L12
and for this as @derberg suggested we can use something similar like 
https://github.com/asyncapi/parser-js/blob/481b24e8f6e2216d0ea5c3d9bb95c938f4736eef/test/browser/browser.spec.ts#L17-L26
Where we create a local server and serve the files. Am I right?

derberg:
perfectly understood! we should have a test asyncapi file inside this repo and host it temporarily on local server during tests, to test `parseFromUrl`. No fetching from server we do not control

chinma-yyy:
Cool, I will work on this issue and raise a PR soon :v:

PR #307, opened 2022-06-30
fix: generate models cannot log in certain cases

jonaslagoni:
**Description**
This PR fixes an issue where the oclif logging function does not have access to the `jsonEnabled` function internally.


test/commands/generate/models.test.ts
```diff
@@ -7,6 +7,16 @@ const generalOptions = ['generate:models'];
 const outputDir = './test/commands/generate/models';
 
 describe('models', () => {
+  test
+    .stderr()
+    .stdout()
+    .command([...generalOptions, 'typescript', 'http://bit.ly/asyncapi'])
```
derberg, line 13: oh, another test that can kick us in our buts in the future, but yeah we need to solve https://github.com/asyncapi/cli/issues/161 first

Issue #303, opened 2022-06-30
Help validate Linux installer publishing


derberg:
#### Reason/Context

[This](https://github.com/asyncapi/cli/pull/187) PR introduced automation for publishing of MacOS `pkg` file. It was also designed in the way that it should be easy to plug in generation of installer files for Windows and Linux.

**This issue is for folks that work on Linux machine** and can do some local testing to check if all works as expected

#### Description

* First make sure installer generation works:
  * run `npm run pack:linux`
  * installer should be produced in `dist` directory. Try to install CLI using it
* Then open a PR with the following changes
  * Uncomment this part in the artifact publishing job: https://github.com/asyncapi/cli/blob/master/.github/workflows/upload-release-assets.yml#L18-L21 and modify as much as needed
  * I don't know what will be the name of the generated installer, so you will probably have to refactor existing script for MacOS installer renaming -> https://github.com/asyncapi/cli/blob/master/scripts/macosPackRename.js
  * Update Readme instructions on how to quickly install CLI with new installer -> https://github.com/asyncapi/cli#installation

aayushmau5:
Hey! I use linux. I'll try this out.

aayushmau5:
![image](https://user-images.githubusercontent.com/54525741/176934948-e0a83e21-6b3a-4184-99c5-811713a1803c.png)
Got this error.

derberg:
@aayushmau5 thanks a bunch for looking into this 🙏🏼 

yeah `.deb` is just there in a comment but doesn't have to end up like this. I do not even know what they generate really. IT can be tarball if this is ok, as long as there are good instructions on how to use it on linux

in regards to the error - seem to me like some local permissions issue no?

aayushmau5:
> in regards to the error - seem to me like some local permissions issue no?

Maybe. I'll look into this, and let you know whether it's an error on my end or not.

derberg:
@aayushmau5 hey, any luck?

aayushmau5:
@derberg Ok, I was able to get the package. Had to install some additional packages related to apt package manager.

The `pack:linux` command outputs the following file:
![image](https://user-images.githubusercontent.com/54525741/180138051-0d5820d1-029e-4347-93e0-e7a6f816cf20.png)

Ran `dpkg -i asyncapi_0.21.2.515e58d-1_amd64.deb` to install the deb file. And, we got `asyncapi` command now :D

derberg:
yes yes yes 🚀 
 
[...]

PR #325, opened 2022-07-25
feat: enable release of CLI distribution for Linux

aayushmau5:
**Description**

- This PR adds linux packaging and releasing for the CLI.
- Update README for installation instruction

**Related issue(s)**: https://github.com/asyncapi/cli/issues/303


aayushmau5:
@derberg Some things remain/todo:

- First, we need to verify that the CI actually works. When I tryed installing the package locally, some additional packages were needed. So, we need to verify whether those package are pre-installed in the CI or not.
- Currently, I have only added `deb` releases, we need to support `.tar.gz` as well, i think.

scripts/linuxDebRename.js
```diff
@@ -0,0 +1,21 @@
+/* eslint-disable @typescript-eslint/no-var-requires */
+
+const { rename } = require('fs').promises;
```
derberg, line 3: There is already a lot of code that is in [macosPackRename.js](https://github.com/asyncapi/cli/blob/master/scripts/macosPackRename.js). Maybe merge both, and have one script to rename all distros? and `"pack:macos:rename": "node scripts/macosPackRename.js"` becomes `"pack:rename": "node scripts/releasePackagesRename.js"` ?

derberg (commented):
There is an additional option from oclif to generate `.tar.gz`, you can try it out too -> https://github.com/oclif/oclif#oclif-pack-tarballs

.github/workflows/upload-release-assets.yml
```diff
@@ -15,10 +15,10 @@ jobs:
...
       matrix:
         include:
-          # - os: ubuntu-latest
-          #   npm_script: pack:linux
-          #   dist_folder: deb
-          #   extension: tar.gz #or deb, we do not know yet
+          - os: ubuntu-latest
+            npm_script: pack:linux
+            dist_folder: deb
+            extension: deb #or deb, we do not know yet
```
aayushmau5, line 21: @derberg I'm not sure what this step does? What's the significance of `extension` here? 

Also, I guess I have to add another command to package for `tar.gz`, right?(and create a new step with `extension: tar.gz`)

derberg: it is used here `files: dist/${{ matrix.dist_folder }}/asyncapi.${{ matrix.extension }}` when pointing to binary file, and its extension

> Also, I guess I have to add another command to package for tar.gz, right?(and create a new step with extension: tar.gz)

yup cause I also guess you will not put dist into `deb` but something else

README.md
```diff
@@ -67,6 +67,19 @@ curl -OL https://github.com/asyncapi/cli/releases/latest/download/asyncapi.pkg
...
+
+You can install this CLI using `dpkg`, a package manager for debian.
+
+```bash
+# Download latest release. To download specific release, your link should look similar to https://github.com/asyncapi/cli/releases/download/v0.13.0/asyncapi.deb. All releases are listed in https://github.com/asyncapi/cli/releases
+curl -OL https://github.com/asyncapi/cli/releases/latest/download/asyncapi.deb
+# Install AsyncAPI CLI
+sudo dpkg -i asyncapi.deb
+```
+
```
derberg, line 82: @aayushmau5 can you also add instruction for others, using `tar.gz`?

derberg (commented):
@aayushmau5 perfect job, just left one comment regarding readme

aayushmau5:
@derberg Added installation instruction for tar files. Let me know what you think.

derberg (approved):
@aayushmau5 thanks a 💯

PR #374, opened 2022-10-17
feat: update Modelina to support AsyncAPI 2.5, Python and Rust generator

jonaslagoni:
**Description**
This PR adds support for AsyncAPI 2.5, Python and Rust model generation by upgrading the modelina dependency.


package.json
```diff
@@ -11,7 +11,7 @@
     "@asyncapi/converter": "^1.1.0",
     "@asyncapi/diff": "^0.4.0",
     "@asyncapi/generator": "^1.9.12",
-    "@asyncapi/modelina": "^0.59.8",
+    "@asyncapi/modelina": "^1.0.0-next.22",
```
derberg, line 14: did you adjust modelina release config, so this is not reverted to `0.59.9` with next regular release?

derberg (changes requested):
you have some conflicts with package-lock and failing tests

package.json
jonaslagoni: Can I adapt the release workflow when it's managed by .github? 🤔

jonaslagoni:
Any reason why the tests are broken? 🧐 I don't see any of the changes I did which affected them, do you?

derberg:
@jonaslagoni hard to say really. First time I see such errors. I see you did not make much changes but you did bump modelina to 1.0 release candidate. I dunno how much changed there, but maybe some basic stuff changed that affects the tests? errory types?

jonaslagoni:
@derberg but the problems with the tests are oclif errors?

package.json
derberg: oh crap, you are right, it is possible with react component with https://github.com/asyncapi/asyncapi-react/blob/master/.github/workflows/bump.yml#L34 hack in global workflow but only because HTML template only depends on react component.

but `cli` needs bumps from other packages, not only `modelina` 🤔 

we need something smarter 🤔

derberg: Opting out `modelina` from `bump.yml` updates pushed from `.github` is the only solution I see I think.
It will require some refactor to https://github.com/asyncapi/.github/blob/master/.github/workflows/global-replicator.yml config.

other than that, I could always add some feature to `derberg/npm-dependency-manager-for-your-github-org` but not sure what that could be

derberg:
@jonaslagoni when I checkout your branch and install it, I get the same warning related errors. I think it is like this only because of installation of some crappy patch. You installed dependencies using `"lockfileVersion": 1` and version from master is `"lockfileVersion": 2`. I recommend taking latest `package-lock.json` from master, and run installation on local again but with npm > 6

package.json
jonaslagoni: Would it make sense that `derberg/npm-dependency-manager-for-your-github-org` was context depending on the version strategy, and "understand" what to update and what not to? 🤔 I.e. in this case "understand" that because we use a next-generation branch, only updates from that can be used going forward instead of those from master? 🧐

derberg:
finally different errors 🚀 😄 
have a look, sometimes message do not match reality. Also try fixing by switching from `toMatch` instead of `toEqual`. I saw @magicmatatjahu did it in one of his PRs. Probably because of switch to `jest`

package.json
derberg: in our scenario `bump` runs in scope of modelina. When it runs in `next` branch scope, it surely can figure that out, that it runs in `next` branch, and do something specific. But the problem is when it runs in scope of `master`, how will it know to not do anything in dependant projects? in `cli` it works on default master branch 🤷🏼 

I don't think there is any other solution than:
- have some .bump.config.yml file that you can add to repo, where you can specify what repos should be ignored, and you would store it in master


or

- modify settings of workflow that updates all repos, manage `bump.yml` workflow as separate workflow, independent from `nodejs` topic. Then we can have `autobump` topic. Only repos with this topic would get updates, and `modelina` would be opted out (not only modelina as we have the same case with parser now) from bump workflow updates and could this way have custom settings (like repos to ignore or entire workflow removed from master)

cc @KhudaDad414 @magicmatatjahu @smoya

jonaslagoni: > But the problem is when it runs in scope of master, how will it know to not do anything in dependant projects?

Will it not be able to process the dependent project package.json file to determine the current version and see whether it uses a pre-build id or which major version it uses?

derberg:
@jonaslagoni you need help?

jonaslagoni:
@derberg I need a better way to integrate with the old parser 😅 See https://github.com/asyncapi/modelina/pull/1050 and https://github.com/asyncapi/parser-js/pull/687

package.json
derberg: technically possible, but why should it have such functionality?

basing on what condition will it know that RC used in package.json is the last one, and actually main release should be used?

jonaslagoni: I might be missing some crucial point here, but in the bump workflow, it should be possible to get the new version of the library, and then take the version of the library it tries to bump, and correlate the semver version with each other to know whether to bump it or not.

Of course, this should not automatically update from pre-release to a new major version. Or I guess it can 🤷

package.json
```diff
@@ -10,13 +10,14 @@
...
-    "@asyncapi/generator": "^1.9.12",
-    "@asyncapi/modelina": "^0.59.9",
+    "@asyncapi/generator": "^1.9.14",
+    "@asyncapi/modelina": "^1.0.0-next.40",
     "@asyncapi/parser": "^1.17.1",
     "@asyncapi/studio": "^0.15.4",
     "@oclif/core": "^1.18.0",
     "@oclif/errors": "^1.3.5",
     "@oclif/plugin-not-found": "^2.3.1",
+    "ajv": "^8.12.0",
```
jonaslagoni, line 20: This is a weird one, and I would be happy if someone had a different solution.

If I dont specify this dependency I keep getting the stacktrace:

```
    TypeError [ERR_INVALID_ARG_TYPE]: The "warning" argument must be of type string or an instance of Error. Received an instance of ModuleLoadError

      at Plugin.warn (node_modules/@oclif/core/lib/config/plugin.js:256:17)
      at node_modules/@oclif/core/lib/config/plugin.js:239:30
          at async Promise.all (index 7)
      at async Plugin._manifest (node_modules/@oclif/core/lib/config/plugin.js:232:24)
      at async Plugin.load (node_modules/@oclif/core/lib/config/plugin.js:127:25)
      at async Config.load (node_modules/@oclif/core/lib/config/config.js:85:9)
      at async Function.load (node_modules/@oclif/core/lib/config/config.js:79:9)
      at async Object.run (node_modules/@oclif/test/lib/load-config.js:13:26)
      at async Object.run (node_modules/@oclif/test/lib/command.js:15:30)
      at async Object.run (node_modules/fancy-test/lib/base.js:68:25)
```

After a bit of debugging, the underlying error is `Cannot find module 'ajv/dist/compile/codegen'`. This originates from the Modelina dependency, that now uses the latest parser implementation...

My hunch is that the ajv dependency are clashing between versions, but it's just weird that studio works fine 🧐 

```
% npm list ajv
@asyncapi/cli@0.27.2 /Users/lagoni/Documents/github/cli
├─┬ @asyncapi/generator@1.9.14
│ ├─┬ @asyncapi/raml-dt-schema-parser@2.0.1
│ │ └─┬ ramldt2jsonschema@1.2.3
│ │   └─┬ json-schema-migrate@0.2.0
│ │     └── ajv@5.5.2 
│ └── ajv@6.12.6 
├─┬ @asyncapi/modelina@1.0.0-next.40
│ ├─┬ @apidevtools/swagger-parser@10.1.0
│ │ └── ajv@8.12.0 
│ └─┬ @asyncapi/parser@2.0.0-next-major.11
│   ├─┬ @stoplight/spectral-core@1.15.1
│   │ └── ajv@8.11.2 
│   ├─┬ @stoplight/spectral-functions@1.7.1
│   │ └── ajv@8.11.2 
│   ├─┬ @stoplight/spectral-rulesets@1.14.1
│   │ └── ajv@8.11.2 
│   ├── ajv@8.12.0 
│   ├─┬ ajv-formats@2.1.1
│   │ └── ajv@8.11.2 
│   └─┬ webapi-parser@0.5.0
│     └── ajv@6.5.2 
├─┬ @asyncapi/parser@1.17.1
│ └── ajv@6.12.6 
├─┬ @asyncapi/studio@0.15.4
│ └─┬ @asyncapi/parser@2.0.0-next-major.10
│   └── ajv@8.11.2 
├─┬ @semantic-release/npm@7.1.3
│ └─┬ npm@7.24.2
│   └─┬ node-gyp@7.1.2
│     └─┬ request@2.88.2
│       └─┬ har-validator@5.1.5
│         └── ajv@6.12.6
``` 

So, by specifically installing this dependency it fixes everything, and I simply dont know why...

jonaslagoni:
I am gonna need some help with the test CI, cause I have no idea why it's failing, any ideas? @derberg

.gitignore
```diff
@@ -17,7 +17,6 @@ node_modules
 asyncapi.json
 asyncapi.yaml
 asyncapi.yml
-/test/docs
```
jonaslagoni, line 20: Because each file is cleaned up in the `fromTemplate` test, there is no reason to ignore it, and because one of the tests require that it's not ignored.

test/commands/generate/fromTemplate.test.ts
```diff
@@ -9,49 +9,60 @@ const generalOptions = [
...
+        'should throw error if output folder is in a git repository',
+        (ctx, done) => {
+          expect(ctx.stderr).toContain(
+            'Error: "./test/docs/2" is in a git repository with unstaged changes.'
+          );
+          cleanup('./test/docs/2');
+          done();
+        }
+      );
+  });
```
jonaslagoni, line 51: I dont quite understand why this worked before, but I guess that's just race conditions in a nutshell 🤷 

I refactored the tests so each tests are completely siloed from the other. Further, this test is specifically set up to trigger the error where before, I have no idea how it was triggered 😆

jonaslagoni:
PR is ready for final review 🙂

Souvikns (dismissed):

derberg (dismissed):

jonaslagoni:
Fixed the conflict from master, feel free to merge whenever you have the reviews you think is needed 🙂

derberg (approved):

PR #376, opened 2022-10-18
feat: integrate v2 ParserJS in validate command

magicmatatjahu:
- Integrate new ParserJS (with Spectral) in the `validate` command
- Integrate new way of validation in other commands - user can see validation diagnostics in `generate models` and `diff` commands and determine if command should fail or should perform logic if diagnostics severity is allowed.
- Write tests

Example input:

```bash
asyncapi validate ./test/specification.yaml --diagnostics-format stylish
```

Example output:

image

**Related issue(s)**
Part of https://github.com/asyncapi/parser-js/issues/481


Souvikns (dismissed):
Looks awesome, one thing to add, when we run the validate command and there is a governance issue it says `File ./test/specification.yml and/or referenced documents have governance issues.` but the file is valid, shouldn't it say that the file is valid but have governance issue.

Souvikns:
I use cli to quickly check if the spec I am writing is valid, and yeah this would help

magicmatatjahu:
@Souvikns Ok, good :) I will change that but before let's see what think about that @derberg

magicmatatjahu:
@derberg ping 😄

src/parser.ts
```diff
@@ -0,0 +1,106 @@
+import { Flags } from '@oclif/core';
+import { Parser } from '@asyncapi/parser/cjs';
+import { AvroSchemaParser } from '@asyncapi/parser/cjs/schema-parser/avro-schema-parser';
+import { OpenAPISchemaParser } from '@asyncapi/parser/cjs/schema-parser/openapi-schema-parser';
+import { RamlSchemaParser } from '@asyncapi/parser/cjs/schema-parser/raml-schema-parser';
+import { getDiagnosticSeverity } from '@stoplight/spectral-core';
+import { html, json, junit, stylish, teamcity, text, pretty } from '@stoplight/spectral-cli/dist/formatters';
+import { OutputFormat } from '@stoplight/spectral-cli/dist/services/config';
```
derberg, line 8: it is pretty risky to import these helpers from `dist` of the CLI.....breaking changes may hurt us later

package.json
```diff
@@ -12,11 +12,12 @@
     "@asyncapi/diff": "^0.4.0",
     "@asyncapi/generator": "^1.9.12",
     "@asyncapi/modelina": "^0.59.8",
-    "@asyncapi/parser": "^1.17.0",
+    "@asyncapi/parser": "^2.0.0-next-major.6",
```
derberg, line 15: and here we have the same problem we have with other PR from Jonas. This will be overwritten with next parser release back to `1.18`

derberg (commented):
just left 2 comments
other review will do tomorrow

test/commands/diff.test.ts
```diff
@@ -6,7 +6,7 @@ describe('diff', () => {
     test
       .stderr()
       .stdout()
-      .command(['diff', './test/specification.yml', './test/specification.yml', '--format=json'])
+      .command(['diff', './test/specification.yml', './test/specification.yml', '--format=json', '--no-log-diagnostics'])
```
derberg, line 9: when I run `diff` successfully I do not want to see diagnostics by default. When I run `diff` I'm interested with `diff`, unless files that I run through diff have errors and `diff` cannot run because of it

derberg (changes requested):
left just few comments

src/parser.ts
magicmatatjahu: It will give you errors during the CLI build that these exports don't exist in the path. There is no other way, and creating these functions from scratch or copying them to our codebase is pointless. There is no good solution here and this is the least invasive.

derberg: then we should at least make sure that in package.json we do not allow `^` for spectral-cli dependency

derberg (commented):
since https://github.com/asyncapi/.github/pull/204 the `bump.yml` workflow is not going to be replicated in repositories that do not have `get-global-node-release-workflows`. This means that you can now easily update it in parser repository, in the `master` branch to make sure regular releases will not revert the version update that you did here. So in parser bump workflow you need to add `cli` to list of ignored repos

magicmatatjahu:
@derberg Ok, suggestions applied! Could you check again?

derberg:
@magicmatatjahu can you address code smells?

magicmatatjahu:
@derberg You can check that we cannot address such a issues, because Modelina doesn't support in input parsed AsyncAPI document (but only in types) - in the logic we check that. However, that code is in HEAD on master https://github.com/asyncapi/cli/blob/master/src/commands/generate/models.ts#L164 so that code isn't added by my PR.

derberg (approved):
@magicmatatjahu right, thanks for point that out

magicmatatjahu:
@jonaslagoni Do you wanna check?

jonaslagoni (approved):
@magicmatatjahu not necessary, but now that I am here LGTM 👍

Souvikns (approved):

magicmatatjahu:
@derberg https://github.com/asyncapi/parser-js/pull/701 and https://github.com/asyncapi/parser-js/pull/702 Could you check? Thanks!

PR #384, opened 2022-11-08
feat: add optional parameters to typescript model generator

### LINKED ISSUE #383: Adding optional parameters to typescript model generator
#### Reason/Context
Please try answering few of those questions
- Why we need this improvement?
Specifying optional parameters to model generator should be easy from cli.

- How will this change help?
Add the following optional cli flags: modelType, enumType, moduleSystem, exportType

- What is the motivation? 
This will allow users or the cli to override the default values for these properties

#### Description
Please try answering few of those questions
- What changes have to be introduced?
Additional cli flags need to be added

- Will this be a breaking change?
No, this will be a backwards compatible change

- How could it be implemented/designed?
This feature request will implement optional flags for TypeScript only, model generators for other languages may have similar requirement.

juanheyns:
**Description**

Implements optional parameters for typescript model generator.

Screen Shot 2022-11-07 at 9 24 44 PM

Test results:

Screen Shot 2022-11-07 at 9 22 13 PM

Tested various other combinations. Implementation was pretty simple.

**Related issue(s)**
Fixes #383

### DIFF
--- a/src/commands/generate/models.ts
+++ b/src/commands/generate/models.ts
@@ -35,25 +35,25 @@ export default class Models extends Command {
     /**
      * TypeScript specific options
      */
-    modelType: Flags.string({
+    tsModelType: Flags.string({
       type: 'option',
       options: ['class', 'interface'],
       description: 'TypeScript specific, define which type of model needs to be generated.',
       required: false,
     }),
-    enumType: Flags.string({
+    tsEnumType: Flags.string({
       type: 'option',
       options: ['enum', 'union'],
       description: 'TypeScript specific, define which type of enums needs to be generated.',
       required: false,
     }),
-    moduleSystem: Flags.string({
+    tsModuleSystem: Flags.string({
       type: 'option',
       options: ['ESM', 'CJS'],
       description: 'TypeScript specific, define the module system to be used.',
       required: false,
     }),
-    exportType: Flags.string({
+    tsExportType: Flags.string({
       type: 'option',
       options: ['default', 'named'],
       description: 'TypeScript specific, define which type of export needs to be generated.',
@@ -77,7 +77,7 @@ export default class Models extends Command {
 
   async run() {
     const passedArguments = await this.parse(Models);
-    const { modelType, enumType, moduleSystem, exportType, namespace, packageName, output } = passedArguments.flags;
+    const { tsModelType, tsEnumType, tsModuleSystem, tsExportType, namespace, packageName, output } = passedArguments.flags;
     const { language, file } = passedArguments.args;
     const inputFile = await load(file) || await load();
     const parsedInput = await parse(inputFile.text());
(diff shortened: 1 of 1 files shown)


magicmatatjahu:
cc @jonaslagoni

magicmatatjahu:
@juanheyns Hi! Please change PR's title to the `feat: adding optional parameters to typescript model generator`.

src/commands/generate/models.ts
```diff
@@ -16,31 +16,68 @@ export default class Models extends Command {
...
+      options: ['ESM', 'CJS'],
+      description: 'TypeScript specific, define the module system to be used.',
+      required: false,
+    }),
+    exportType: Flags.string({
+      type: 'option',
+      options: ['default', 'named'],
+      description: 'TypeScript specific, define which type of export needs to be generated.',
+      required: false,
+    }),
```
magicmatatjahu, line 61: I don't think so that this solution is good for future features. At the moment we support 5-6 languages and in future we can support many more and hardcode some flags for particular language is a bad solution. I think that we should go with idea like having `--config` which should be stringified JSON and the shape of the`--config` should be determine by chosen language. What do you think?

cc @jonaslagoni

magicmatatjahu: And value for the `--config` flag should meet the interface of options that can be passed to a given generator - e.g. for TS - https://github.com/asyncapi/modelina/blob/next/src/generators/typescript/TypeScriptGenerator.ts#L17 - of course without options where you need to pass some runtime references.

jonaslagoni: I actually think we should have both 🤔

Cause some would much rather just use pure parameters, especially if they need to quickly generate the models, then they dont care about the config file.

magicmatatjahu: But you can have conflicts with parameters names for other languages, that's my whole point 😄 Also if we go with `--config` we don't need to worry about updating CLI when new Modelina will release, because we will pass options shape.

jonaslagoni: If there are conflicts, we can easily adapt the parameters 🙂 I vote to have the parameters and in future major versions you can always remove them if you find them too challenging, when the config feature is added 😄 

> Also if we go with --config we don't need to worry about updating CLI when new Modelina will release, because we will pass options shape.

That is true yes.

But the configuration feature is not likely to be implemented this year.

magicmatatjahu: So I opt for having prefix in flags like `--tsExportType`

juanheyns: Updated flag names with a ts prefix.

juanheyns: Do you want to implement a `--config` option as part of this PR or as part of a future enhancement?

jonaslagoni: Definitely future 🙂

jonaslagoni (dismissed):
LGTM 👍

derberg (commented):
@magicmatatjahu lgtm but you were first reviewer so I need your 🟢

jonaslagoni:
@magicmatatjahu any update?

magicmatatjahu (approved):

magicmatatjahu:
@derberg ping :)

Issue #205, opened 2022-01-31
Fetching examples from master causes errors during release


derberg:
#### Describe the bug
because of
```
const SPEC_EXAMPLES_ZIP_URL = 'https://github.com/asyncapi/spec/archive/refs/heads/master.zip';
```

that is in https://github.com/asyncapi/cli/blob/master/scripts/fetch-asyncapi-example.js#L9 script, we will always have few hours during spec release where CLI will fail in development, on ci/cd. It is because it is fetching stuff from master, that was already released, like now, we released spec 2.3. But CLI is not updated with parser that supports 2.3 cause it is not yet released

so we need to use:
```
const SPEC_EXAMPLES_ZIP_URL = 'https://github.com/asyncapi/spec/archive/refs/tags/v2.2.0.zip';
```

but then it means we have to have a manual bump of CLI, unless we improve the script to somehow check what version it should fetch. Like maybe use parser before fetching, to check what is the latest version of the spec that it is using 🤔

imabp:
Hey @derberg I hope this fix works :  #213

imabp:
Hey @derberg 
The latest release of cli (v0.13.0 ), supports parser version  1.14.0 which supports the spec version 2.3.0
If we are following this pattern. 

<table>

<tr>
<td>parser</td>
<td>spec</td>
</tr>
<tr>
<td>1.13.x</td>
<td>2.2.x</td>
</tr>
<tr>
<td>1.14.x</td>
<td>2.3.x</td>
</tr>
</table> 

Using this pattern, `MAJOR.MINOR.PATCH` it seems 

for every  `MINOR` update in spec,  there is `MINOR` update required for parser.
And if we stay consistent with this approach, we can implement the script in that logic.

derberg:
@imabp do you plan to continue working on this one? https://github.com/asyncapi/cli/pull/213

PR #390, opened 2022-11-16
feat: add new example to `asyncapi new` command

derberg:
The idea of a new example is to have it reflect the same example that is used in docs.

why? so tutorials can have a simple command that reader can use to bootstrap example asyncapi file if they do not have any:
```
asyncapi new --example=tutorial.yml --no-tty
```

Resolve https://github.com/asyncapi/cli/issues/205

### DIFF
--- a/scripts/fetch-asyncapi-example.js
+++ b/scripts/fetch-asyncapi-example.js
@@ -6,7 +6,7 @@ const unzipper = require('unzipper');
 const path = require('path');
 const parser = require('@asyncapi/parser');
 
-const SPEC_EXAMPLES_ZIP_URL = 'https://github.com/asyncapi/spec/archive/refs/tags/v2.3.0.zip';
+const SPEC_EXAMPLES_ZIP_URL = 'https://github.com/asyncapi/spec/archive/refs/heads/master.zip';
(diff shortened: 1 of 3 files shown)


scripts/fetch-asyncapi-example.js
```diff
@@ -6,7 +6,7 @@ const unzipper = require('unzipper');
 const path = require('path');
 const parser = require('@asyncapi/parser');
 
-const SPEC_EXAMPLES_ZIP_URL = 'https://github.com/asyncapi/spec/archive/refs/tags/v2.3.0.zip';
+const SPEC_EXAMPLES_ZIP_URL = 'https://github.com/asyncapi/spec/archive/refs/heads/master.zip';
```
derberg, line 9: yeah, for reasons like https://github.com/asyncapi/spec/pull/873 I think we need to always point to master.

derberg: as a result we can close:
- https://github.com/asyncapi/cli/pull/213
- https://github.com/asyncapi/cli/issues/205

Souvikns (approved):

magicmatatjahu (approved):

alequetzalli:
THANK YOU!!! ✨✨✨

PR #391, opened 2022-11-16
feat: add bundle command

### LINKED ISSUE #219: Add `asyncapi bundle` command

Souvikns:
**Description**

Integrating [bundler](https://github.com/asyncapi/bundler/) into CLI by introducing the command `asyncapi bundle`. 

Usage: 
```
asyncapi bundle ./file1.yml ./file2.yml --output asyncapi.yaml --referenceIntoComponents

```

**Related issue(s)**

Resolves #219


src/commands/bundle.ts
```diff
@@ -0,0 +1,81 @@
+import { Flags } from '@oclif/core';
+import { Example } from '@oclif/core/lib/interfaces';
+import Command from '../base';
+import bundle from '@asyncapi/bundler';
+import fs from 'fs';
+import path from 'path';
+import { ErrorLoadingSpec } from '../errors/specification-file';
+
+export default class Bundle extends Command {
+  static description = 'Bundle multiple asyncapi files together.';
```
derberg, line 10: is it only for multiple files, what if I have one file, it still bundles references right?

src/commands/bundle.ts
```diff
@@ -0,0 +1,81 @@
...
+
+  static examples: Example[] = [
+    'asyncapi bundle ./asyncapi.yaml -o final.yaml',
+    'asyncapi bundle ./asyncapi.yaml ./spec.yaml --reference-into-components'
+  ];
+
+  static flags = {
+    help: Flags.help({ char: 'h' }),
+    output: Flags.string({ char: 'o', description: 'The output file name. Omitting this flag will create a main.yaml.' }),
+    'reference-into-components': Flags.boolean({ char: 'r', description: 'Bundle the message $refs into components.' }),
```
derberg, line 21:
```suggestion
    'reference-into-components': Flags.boolean({ char: 'r', description: 'Bundle the message $refs into Components object.' }),
```

src/commands/bundle.ts
```diff
@@ -0,0 +1,81 @@
...
+  static strict = false;
+
+  static examples: Example[] = [
+    'asyncapi bundle ./asyncapi.yaml -o final.yaml',
+    'asyncapi bundle ./asyncapi.yaml ./spec.yaml --reference-into-components'
+  ];
+
+  static flags = {
+    help: Flags.help({ char: 'h' }),
+    output: Flags.string({ char: 'o', description: 'The output file name. Omitting this flag will create a main.yaml.' }),
```
derberg, line 20:
```suggestion
    output: Flags.string({ char: 'o', description: 'The output file name. Omitting this flag will create a main.yaml.' }),
```

`main.yml` or maybe `bundle.yml`? I would also keep promoting `asyncapi` as suffix, so `asyncapi-main` or `asyncapi-bundle`

src/commands/bundle.ts
```diff
@@ -0,0 +1,81 @@
...
+    'reference-into-components': Flags.boolean({ char: 'r', description: 'Bundle the message $refs into components.' }),
+    base: Flags.string({ char: 'b' , description: 'Path to the file which will act as a base. This is required when some properties are to needed to be overwritten.'}),
```
derberg, line 22:
```suggestion
    base: Flags.string({ char: 'b' , description: 'Path to the file which will act as a base. This is required when some properties are to needed to be overwritten.'}),
```

what do you mean by this description? `This is required when some properties are to needed to be overwritten.`

derberg: good to have example for it in help

main.yml
```diff
@@ -0,0 +1,40 @@
+asyncapi: 2.2.0
```
derberg, line 1: did you really want to commit it?

src/commands/bundle.ts
derberg: also, with all other commands except of `generator` that has technical limitation, we should by default output to terminal, if `-o` is not provided

test/commands/bundle/bundle.test.ts
```diff
@@ -0,0 +1,104 @@
+import { test } from '@oclif/test';
+import fs from 'fs';
+import path from 'path';
+
+function cleanup(filepath: string) {
+  fs.unlinkSync(filepath);
+}
+
+const spec = `asyncapi: 2.2.0
```
derberg, line 9: can we use `2.5.0`?

test/commands/bundle/bundle.test.ts
```diff
@@ -0,0 +1,104 @@
+import { test } from '@oclif/test';
+import fs from 'fs';
+import path from 'path';
+
+function cleanup(filepath: string) {
```
derberg, line 5: looks like good candidate for reusable helper -> https://github.com/asyncapi/cli/blob/master/test/commands/generate/fromTemplate.test.ts#L12

src/commands/bundle.ts
derberg: I also haven't seen a test for `-b` or `--base` flag

derberg (changes requested):
I left some comments. Can we agree not to use `sync`-like functions in base code. In tests it is fine but in main CLI code I would prefer we always use `async/await` style

src/commands/bundle.ts
Souvikns: Yeah, it first picks up a single file resolve references (reference it into components if required) then combines it with other spec files (which have gone through the same process).

src/commands/bundle.ts
Souvikns: > what do you mean by this description? This is required when some properties are to needed to be overwritten.

For example when you have two separate files and after combining you want to have a separate new info object or add a new license, you can pass a object that would be used to change those things as per required. 

I am currently testing it, I am facing some issues as it is not working as intended with the new `reference-into-components` feature.

src/commands/bundle.ts
```diff
@@ -0,0 +1,86 @@
...
+            path.resolve(process.cwd(), flags.base || ''),
+            'utf-8'
+          )
+          : undefined,
+      }
+    );
+
+    if (!output) {
+      console.log(document.yml())
+    } else {
```
Souvikns, line 48: Outputing to the console could used to pipe to file, like `asyncapi bundle ./specfile.yaml > asyncapi.yaml` by default it outputs in `yml` format should we have a format parameter because someone will want to have it in json, or just follow the same format as the input file?

derberg: imho we should follow input. Flag for manipulating output can be added upon request. Unless you want to work on it now, up to you

src/commands/bundle.ts
derberg: so maybe for now remove it from PR so it doesn't block you?

Souvikns: Already fixed, it in the bundler

derberg:
@Souvikns remember to click arrows when ready for another review
Screenshot 2022-12-01 at 14 28 33

README.md
```diff
@@ -136,6 +136,7 @@ COMMANDS
   new       creates a new AsyncAPI file
   start     starts a new local instance of Studio
   validate  validate an AsyncAPI file
+  bundle    bundle multiple asyncapi files together
```
derberg, line 139: sorry forgot about this one in last review

bundler is not only for bundling multiple asyncapi files, it can also bundle all references for a single document, right?

```suggestion
  bundle    bundle one or multiple asyncapi documents and their references together
```

src/commands/bundle.ts
```diff
@@ -0,0 +1,76 @@
...
+    'asyncapi bundle ./asyncapi.yaml --output final-asyncapi.yaml',
+    'asyncapi bundle ./asyncapi.yaml ./features.yaml --reference-into-components',
+    'asyncapi bundle ./asyncapi.yaml ./features.yaml --base ./asyncapi.yaml --reference-into-components'
+  ];
+
+  static flags = {
+    help: Flags.help({ char: 'h' }),
+    output: Flags.string({ char: 'o', description: 'The output file name. Omitting this flag will create a main.yaml.' }),
+    'reference-into-components': Flags.boolean({ char: 'r', description: 'Bundle the message $refs into components object.' }),
+    base: Flags.string({ char: 'b', description: 'Path to the file which will act as a base. This is required when some properties are to needed to be overwritten.' }),
```
derberg, line 26:
```suggestion
    base: Flags.string({ char: 'b', description: 'Path to the file which will act as a base. This is required when some properties need to be overwritten.' }),
```

What `This is required when some properties need to be overwritten` means? I tried to figure from bundle project history why it was added, also looked into test case, but in test cases we do not evaluate bundle against some real result, so I could not figure quickly what is the difference between bundling with and without `base`

src/commands/bundle.ts
```diff
@@ -0,0 +1,76 @@
...
+  static examples: Example[] = [
+    'asyncapi bundle ./asyncapi.yaml > final-asyncapi.yaml',
+    'asyncapi bundle ./asyncapi.yaml --output final-asyncapi.yaml',
+    'asyncapi bundle ./asyncapi.yaml ./features.yaml --reference-into-components',
+    'asyncapi bundle ./asyncapi.yaml ./features.yaml --base ./asyncapi.yaml --reference-into-components'
+  ];
+
+  static flags = {
+    help: Flags.help({ char: 'h' }),
+    output: Flags.string({ char: 'o', description: 'The output file name. Omitting this flag will create a main.yaml.' }),
```
derberg, line 24: `Omitting this flag will create a main.yaml` this is not true anymore right?

derberg (changes requested):

src/commands/bundle.ts
Souvikns: Yeah it is not true, will update the flag

src/commands/bundle.ts
Souvikns: maybe this could help https://github.com/asyncapi/bundler/issues/101 , this is something most people wouldn't be interested in, but it is like nice to have feature. You can use `base` flag to overwrite or retain info object when you are merging different asyncapi files.

derberg: when I read explanation, I understand that `base` is basically something that holds basic information that should stay in asyncapi.yml and all the other files are added to it, cannot override what is in base already

but then I see the linked issue and the expected behavior section. Looks like what is there in base is lost and expected result is `This service is in charge of processing user logouts` and not what is in base `This service is in charge of processing user signups`

sorry for all these questions :(

Souvikns:
Not sure why but now  it is unable to install dependencies https://github.com/asyncapi/cli/actions/runs/4163522302/jobs/7204027231#step:7:69

derberg:
I think it is random npm issue, saw it few times already in other projects -> https://github.com/asyncapi/cli/actions/runs/4163522302/jobs/7204027231#step:7:41

I rerun, let's see

Souvikns:
@derberg could this be a ubuntu related issues, on mac at least `npm i` works

derberg (commented):
it makes no sense 😭 

maybe try to bump rimraf to latest -> 4.1.2

we use it only in the npm scripts so we should not be probably afraid of the major changes that version 4 brought

derberg:
@Souvikns yo, with all the recent changes here and there, try fixing conflicts and maybe 🤞🏼 this time will be good

derberg:
@Souvikns you have them there again. Remember to ping me once you solve them. Rerequest review using GitHub buttons. I do not get notifications that you pushed code

derberg (approved):

Issue #184, opened 2022-01-17
Make it easier to create templates and glee projects


jonaslagoni:
#### Reason/Context
We have had talks before about making it easier to create new generator templates [through something like `npx create-asyncapi-template`](https://bonsaiilabs.com/create-npx-starter-command/). However, I think that our CLI could fulfill that role to offer a better user experience for developers, to easily bootstrap new templates or glee projects.

What do you think, does it fit the CLI use-case?

derberg:
imho yes, AsyncAPI CLI should be a single place where we integrate all AsyncAPI tools to make work from the terminal a pleasure. So folks install just one tool to use all the others. One 💍 to rule them all 🤓 

it is just a matter of coming up with proper command names and good DX

imabp:
Hey, this sounds amazing!! 🤩
Would like to work on this.

jonaslagoni:
Would `bootstrap` make sense? 🤔 

```
asyncapi bootstrap glee --output ./output-dir
asyncapi bootstrap glee mqtt --output ./output-dir 

asyncapi bootstrap template --output ./output-dir
//Incase you want to copy an existing template
asyncapi bootstrap template @asyncapi/java-cloud-stream --output ./output-dir
```

derberg:
I think `scaffold` is more accurate technical word that would explain what is done, not sure `bootstrap` is correct 🤔 

can you provide some more explanation of what each command would actually generate?

imabp:
Yes @derberg , `scaffold` sounds less confusing than bootstrap.

So I was thinking to start it with, something like support of Async API for NextJS API Directory.
Objective:
The user can create NextJS Serverside Asynchronous APIs from the AsyncAPI spec file.

Assuming, AsyncAPI and NextJS are installed globally.

Steps 
1. User should be able to run -> 
`asyncapi scaffold nextjs --output <NextJS_API_Directory>` 

2. This builds a directory with particular endpoints as separate filenames (That's how NextJS function), and a NextJS friendly code. 
The `--typescript` flag will help to choose the language.

----
Your suggestions: @jonaslagoni

fmvilas:
What about `new`? We already have it and would be just a matter of extending it:

```
asyncapi new --> New AsyncAPI file, as it is now.
asyncapi new file --> Same as above.
asyncapi new project --> New Glee project
asyncapi new glee --> New Glee project
asyncapi new generator --> New Generator template (or generator if we end up calling them like this)
[...]

Issue #337, opened 2021-07-18
Create `create-template` in AsyncAPI organization 


jonaslagoni:
#### Reason/Context
Stumbled upon this and thought it would be perfect for the generator. We could offer a way to use the [init function](https://docs.npmjs.com/cli/v7/commands/npm-init) which will allow people to more quickly setup new templates.

By creating `create-template` in the AsyncAPI scope we can do the following example commands:
- `npm init @asyncapi/template --react`
- `npm init @asyncapi/template --nunjucks`
- `npm init @asyncapi/template`

#### Description
This will significantly speed up the process of creating new ones and make it more accessible for people. This is similar to what React have - https://github.com/facebook/create-react-app

derberg:
I think we should follow the same direction that will be done with https://github.com/asyncapi/cli/pull/418

Shurtu-gal:
@derberg @jonaslagoni can I work on this?

derberg:
yeah, but problem is not the implementation, we need a clear definition of the project that we would like to have generated with this new command. Some brainstorming session is kinda needed

Shurtu-gal:
Still Relevant 🤝

Shurtu-gal:
Still relevant 🤝

AayushSaini101:
I want to work on this. :)

DhairyaMajmudar:
Hi @derberg @jonaslagoni I would love to work on this issue, actually I am developing a similar kind of npm package https://github.com/DhairyaMajmudar/majirya-cli

This package a CLI Tool which simplifies mern stack and T3 stack web development.

Amzani:
Hi @DhairyaMajmudar @AayushSaini101 was the first to express his interest in working on this issue.
Feel free to pick the other issues in the project.

Thanks

DhairyaMajmudar:
Sure looking forward to apply for an another issue, thank you : )

AayushSaini101:
We can implement a command that is asyncapi init that will generate a template for the user, and we have already have template of template structure https://github.com/asyncapi/template-for-generator-templates we can simply use it @Amzani @Souvikns

aeworxet:
### Bounty Issue's service comment

Text labels: `bounty/2024-Q3`, `bounty/advanced`, `bounty/coding`
First assignment to third-party contributors: [2024-06-21 00:00:00 UTC+12:00](https://time.lol/#2024-06-21T00:00:00+12:00)
[...]

PR #418, opened 2022-12-16
feat: implement 'new glee' command

peter-rr:
**Related issue(s)**
See also #184

### DIFF
--- a/src/commands/new/glee.ts
+++ b/src/commands/new/glee.ts
@@ -1,30 +1,25 @@
 import {Flags} from '@oclif/core';
 import { promises as fPromises } from 'fs';
-import Command from '../base';
+import Command from '../../base';
 import * as inquirer from 'inquirer';
-import { start as startStudio, DEFAULT_PORT } from '../models/Studio';
+import { start as startStudio, DEFAULT_PORT } from '../../models/Studio';
 import { resolve } from 'path';
 
 const { writeFile, readFile } = fPromises;
 const DEFAULT_ASYNCAPI_FILE_NAME = 'asyncapi.yaml';
 const DEFAULT_ASYNCAPI_TEMPLATE = 'default-example.yaml';
 
-export default class New extends Command {
-  static description = 'Creates a new asyncapi file';
+export default class NewGlee extends Command {
+  static description = 'Creates a new Glee project';
 
   static flags = {
     help: Flags.help({ char: 'h' }),
-    'file-name': Flags.string({ char: 'n', description: 'name of the file' }),
-    example: Flags.string({ char: 'e', description: 'name of the example to use' }),
-    studio: Flags.boolean({ char: 's', description: 'open in Studio' }),
-    port: Flags.integer({ char: 'p', description: 'port in which to start Studio' }),
-    'no-tty': Flags.boolean({ description: 'do not use an interactive terminal' }),
+    name: Flags.string({ char: 'n', description: 'name of the project', default: 'glee-project' }),
   };
 
-  static args = [];
-
   async run() {
-    const { flags } = await this.parse(New); // NOSONAR
+    // Create new asyncapi file
+    const { flags } = await this.parse(NewGlee); // NOSONAR
     const isTTY = process.stdout.isTTY;
 
     if (!flags['no-tty'] && isTTY) {
@@ -43,6 +38,36 @@ export default class New extends Command {
         this.warn('Warning: --studio flag was passed but the terminal is not interactive. Ignoring...');
       }
     }
+    
+    // // Create new glee project
+    // const { args } = await this.parse(New);
+    // const gleePath = args['glee-project'];
+
+    // if (gleePath) {
+    //   // Create glee project folder
+    //   await fPromises.mkdir(gleePath);
+
+
+    //   // Create 'functions' folder inside glee project folder
+    //   await fPromises.mkdir(`${gleePath}/functions`);
+
+
+    //   // DO NOT create asyncapi.yaml file -> Get the file and copy to it to glee folder
+    //   const specFile = this.createAsyncapiFile(fileName, template);
+    //   if (specFile) {
+    //     await fPromises.copyFile(__dirname, `${gleePath}`);
+    //   } else {
+    //     await fPromises.writeFile(`${gleePath}`, '../../assets/examples/examples.json', { encoding: 'utf8' });
+
+    //   }
+
+    //   // Create package.json file
+
+    //       // DO NOT "npm install to generate file" -> Access to "https://github.com/asyncapi/create-glee-app/blob/master/templates/default/package.json"
+    //       // and copy it (same like was done with asyncapi file)
+    //       // DO NOT add glee dependency (already included in the accessed file)
+
+    // }
   }
 
   /* eslint-disable sonarjs/cognitive-complexity */
--- a/src/commands/new/index.ts
+++ b/src/commands/new/index.ts
@@ -1,0 +1,2 @@
+import file from './file.js';
+export default file;
--- a/package-lock.json
+++ b/package-lock.json
@@ -400,8 +400,11 @@
       }
     },
     "node_modules/@asyncapi/minimaltemplate": {
-      "resolved": "test/minimaltemplate",
-      "link": true
+      "resolved": "file:test/minimaltemplate",
+      "dev": true,
+      "dependencies": {
+        "@asyncapi/generator-react-sdk": "^0.2.25"
+      }
     },
     "node_modules/@asyncapi/modelina": {
       "version": "0.59.9",
@@ -23554,12 +23557,6 @@
           "optional": true
         }
       }
-    },
-    "test/minimaltemplate": {
-      "dev": true,
-      "dependencies": {
-        "@asyncapi/generator-react-sdk": "^0.2.25"
-      }
     }
   },
   "dependencies": {
@@ -23820,7 +23817,7 @@
       }
     },
     "@asyncapi/minimaltemplate": {
-      "version": "file:test/minimaltemplate",
+      "dev": true,
       "requires": {
         "@asyncapi/generator-react-sdk": "^0.2.25"
       }
--- a/src/commands/new/file.ts
+++ b/src/commands/new/file.ts
@@ -1,0 +1,134 @@
+import {Flags} from '@oclif/core';
+import { promises as fPromises } from 'fs';
+import Command from '../../base';
+import * as inquirer from 'inquirer';
+import { start as startStudio, DEFAULT_PORT } from '../../models/Studio';
+import { resolve } from 'path';
+
+const { writeFile, readFile } = fPromises;
+const DEFAULT_ASYNCAPI_FILE_NAME = 'asyncapi.yaml';
+const DEFAULT_ASYNCAPI_TEMPLATE = 'default-example.yaml';
+
+export default class NewFile extends Command {
+  static description = 'Creates a new asyncapi file';
+
+  static flags = {
+    help: Flags.help({ char: 'h' }),
+    'file-name': Flags.string({ char: 'n', description: 'name of the file' }),
+    example: Flags.string({ char: 'e', description: 'name of the example to use' }),
+    studio: Flags.boolean({ char: 's', description: 'open in Studio' }),
+    port: Flags.integer({ char: 'p', description: 'port in which to start Studio' }),
+    'no-tty': Flags.boolean({ description: 'do not use an interactive terminal' }),
+  };
+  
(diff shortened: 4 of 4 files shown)


src/commands/new/glee.ts
```diff
@@ -0,0 +1,42 @@
...
+    const projectName = flags.name;
+
+    const PROJECT_DIRECTORY = path.join(process.cwd(), projectName); // cwd-> /Users/pedro.ramos/cli
+    const GLEE_TEMPLATES_DIRECTORY = resolve(__dirname, '../../../create-glee-app/templates/default'); // __dirname-> /Users/pedro.ramos/cli/lib/commands/new
+
+    try {
+      await fPromises.mkdir(PROJECT_DIRECTORY);
+      console.log('Glee project folder created!');
+    } catch (err) {
+      console.error('An existing folder is already using the same name of your Glee project. Please specify a different name for the new project, or just remove the existing folder from the current directory.');
```
fmvilas, line 31: Suggesting a better message error IMHO:

```suggestion
      console.error(`Unable to create the project. We tried to use "${projectName}" as the directory of your new project but it already exists (${PROJECT_DIRECTORY}). Please specify a different name for the new project. For example, run the following command instead:\n\nasyncapi new glee --name ${projectName}-1`);
```

src/commands/new/glee.ts
```diff
@@ -0,0 +1,42 @@
...
+      return;
+    }
+    
+    try {
+      await fs.copy(GLEE_TEMPLATES_DIRECTORY, PROJECT_DIRECTORY);
+      console.log('Content copied to your project folder!');
```
fmvilas, line 37: This is a good example for you to learn. The user doesn't know how their project is being created behind the scenes. They don't know anything about us copying it or how we do it internally. In this message, you're explaining the result of running your code instead of explaining the result of executing `asyncapi new glee`, which is what the user is doing. I propose we change it to the following:

```suggestion
      console.log(`Your Glee project "${projectName}" has been created successfully!\n\nNext steps:\n\n  cd ${projectName}\n  npm install\n  npm run dev\n\nAlso, you can already open the project in your favorite editor and start tweaking it.`);
```

src/commands/new/index.ts
```diff
@@ -0,0 +1,2 @@
+import file from './file.js';
```
fmvilas, line 1: I think you can remove the `.js` extension:

```suggestion
import file from './file';
```

fmvilas (changes requested):
This is going great! 💪 

I left a few suggestions. Also, if I recall correctly, the `asyncapi new glee` command should have an alias called `asyncapi new project`. You should add that by creating the `project.ts` file in the same directory. It can simply be a file that imports the `glee.ts` and exports everything it's exporting (same as with `new/index.ts`, which is an alias of `new/file.ts`). Also, make sure console messages suggest `asyncapi new project` or `asyncapi new glee`, depending on which one has been used.

Keep up the great work 🙌

src/commands/new/glee.ts
```diff
@@ -0,0 +1,42 @@
...
+    const { flags } = await this.parse(NewGlee); // NOSONAR
+
+    const projectName = flags.name;
+
+    const PROJECT_DIRECTORY = path.join(process.cwd(), projectName); // cwd-> /Users/pedro.ramos/cli
+    const GLEE_TEMPLATES_DIRECTORY = resolve(__dirname, '../../../create-glee-app/templates/default'); // __dirname-> /Users/pedro.ramos/cli/lib/commands/new
+
+    try {
+      await fPromises.mkdir(PROJECT_DIRECTORY);
+      console.log('Glee project folder created!');
```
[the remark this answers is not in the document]
peter-rr: >  Also, make sure console messages suggest `asyncapi new project` or `asyncapi new glee`, depending on which one has been used.

I guess we could apply your suggestion also to this console message, right? 🤔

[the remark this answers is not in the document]
peter-rr: Apart from that, I think this console message should be removed cause it would be redundant after implementing this other suggestion: https://github.com/asyncapi/cli/pull/418#discussion_r1087651323

src/commands/new/glee.ts
peter-rr: I guess I should suggest here `asyncapi new glee --name ${projectName}-1` or `asyncapi new project --name ${projectName}-1` depending on which subcommand has been used ('glee' or 'project'). My question is: how could I access to that value so I can use it in the code? 🤔

src/commands/new/glee.ts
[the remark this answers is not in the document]
fmvilas: > Apart from that, I think this console message should be removed since it would be redundant after implementing this other suggestion: [#418 (comment)](https://github.com/asyncapi/cli/pull/418#discussion_r1087651323)

Exactly. Good point 👍

src/commands/new/glee.ts
fmvilas: Yeah, there are many ways but since you're using classes I think you can take advantage of class inheritance. It's hard to explain here but I'll try. On the project.ts file you do something like this:

```ts
import NewGleeCommand from './glee'

export default class NewProjectCommand extends NewGleeCommand {
  constructor(...args) {
    super(..args)
    this.commandName = 'project'
  }
}
```

and on the glee.ts file you put:

```ts
export default class NewGleeCommand extends Command {
  protected commandName = 'glee'
```

Then on each of the places where you need to retrieve that name, you do `this.commandName` and that will give you either `proejct` or `glee`, depending on the command that was used.

It's hard to explain here but I'm happy to jump into a call with you tomorrow if you need it.

fmvilas (changes requested):

src/commands/new/glee.ts
```diff
@@ -0,0 +1,60 @@
...
+
+    try {
+      await fPromises.mkdir(PROJECT_DIRECTORY);
+    } catch (err) {
+      switch (err.code) {
+      case 'EEXIST':
+        console.error(`Unable to create the project. We tried to use "${projectName}" as the directory of your new project but it already exists (${PROJECT_DIRECTORY}). Please specify a different name for the new project. For example, run the following command instead:\n\n  asyncapi new ${this.commandName} --name ${projectName}-1\n`);
+        break;
+      case 'EACCES':
+        console.error(`Unable to create the project. We tried to access the "${PROJECT_DIRECTORY}" directory but it was not possible due to file access permissions. Please check the read/write permissions at your current working directory.`);
```
fmvilas, line 37:
```suggestion
        console.error(`Unable to create the project. We tried to access the "${PROJECT_DIRECTORY}" directory but it was not possible due to file access permissions. Please check the write permissions of your current working directory ("${process.cwd()}").`);
```

src/commands/new/glee.ts
```diff
@@ -0,0 +1,60 @@
...
+        break;
+      case 'EPERM':
+        console.error(`Unable to create the project. We tried to create the "${PROJECT_DIRECTORY}" directory but the operation requires elevated privileges. Please check the privileges for your current user.`);
+        break;
+      default:
+        console.error('Unable to create the project. Something went wrong during the process. Please check the following error message for further info about the issue actually happening:');
```
fmvilas, line 43: I think we can simplify the message:

```suggestion
        console.error('Unable to create the project. Something went wrong during the process. Please check the following message for further info about the error:');
```

src/commands/new/glee.ts
```diff
@@ -0,0 +1,60 @@
...
+    }
+    
+    try {
+      await fs.copy(GLEE_TEMPLATES_DIRECTORY, PROJECT_DIRECTORY);
+      await fPromises.rename(`${PROJECT_DIRECTORY}/env`, `${PROJECT_DIRECTORY}/.env`);
+      await fPromises.rename(`${PROJECT_DIRECTORY}/gitignore`, `${PROJECT_DIRECTORY}/.gitignore`);
+      await fPromises.rename(`${PROJECT_DIRECTORY}/README-template.md`, `${PROJECT_DIRECTORY}/README.md`);
+      console.log(`Your project "${projectName}" has been created successfully!\n\nNext steps:\n\n  cd ${projectName}\n  npm install\n  npm run dev\n\nAlso, you can already open the project in your favorite editor and start tweaking it.`);
+    } catch (err) {
+      console.error('Unable to create the project. Something went wrong during the process. Please check the following error message for further info about the issue actually happening:');
```
fmvilas, line 56:
```suggestion
      console.error('Unable to create the project. Please check the following error message for further info about the issue actually happening:');
```

Removing the second sentence as it's not providing any value.

fmvilas (changes requested):
Left some comments but it's already looking good to get out of the Draft PR stage 🙌

fmvilas (changes requested):
You have conflicts in the package-lock.json file. My suggestion is that you fix it by doing `git pull` (you'll get the conflicts in your machine), remove `package-lock.json` and `node_modules`, and then run `npm install`. After that, just continue the usual `git commit` workflow.

derberg:
just my suggestion is that better do not remove `package-lock.json` as the installation that will follow will pump all the available patches of all dependencies. I once in the past got nice tip that it is enough to resolve conflicts in `package.json` and then run `npm i --verbose` and `package-lock.json` will be properly fixed

peter-rr:
> just my suggestion is that better do not remove `package-lock.json` as the installation that will follow will pump all the available patches of all dependencies. I once in the past got nice tip that it is enough to resolve conflicts in `package.json` and then run `npm i --verbose` and `package-lock.json` will be properly fixed

I've tried out that suggestion as you can see on my [last commit](https://github.com/asyncapi/cli/pull/418/commits/863ff09c1021f9756e89fed492ef4b6d8c87a1be) but it seems the conflicts in `package-lock.json` are not resolved yet.

I applied the changes in `package.json` from the latest version in master branch, remove `node_modules` and then run `npm i --verbose`. Do you know if I'm missing something? 🤔 @derberg @fmvilas

derberg:
@peter-rr on monday I had a call with one CLI contributor and he had exactly the same error and `npm i --verbose` worked lovely. Maybe you got some errors regarding dependency conflicts related to `npm` version compatibility. Try `npm i --verbose --force`

fmvilas (approved):
LGTM 👍

Souvikns (dismissed):

peter-rr:
Working on adding some tests related to the command before merging 🛠️

peter-rr:
CI is passing now after adding the tests. Do you think more tests are needed for this 1st iteration? @Souvikns @fmvilas

fmvilas:
@peter-rr we replied on https://asyncapi.slack.com/archives/CQVJXFNQL/p1678105129062329. Did you decide to try out something in the end?

peter-rr:
Merge conflicts solved and no more tests added for this iteration. 
@Souvikns @fmvilas Could you approve the changes so we can merge the PR? Thanks :)

Souvikns (changes requested):
I just ran the command locally and it works like a charm, one thing though I see it is not generating any spec file, when I do `npm run dev` it gives me this error. 
Screenshot 2023-03-20 at 12 08 51 PM


I think we can solve this by generating a spec file to go with the functions.

peter-rr:
> I think we can solve this by generating a spec file to go with the functions.

Great, let me take a look 👀

Souvikns:
getting this error in tests 
```
(node:51125) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 21 error listeners added to [WriteStream]. Use emitter.setMaxListeners() to increase limit
```

I see that the MaxListeners were increased, https://github.com/asyncapi/cli/blob/6e1484b702dda9678ff17d34e8a48656b7cbf95e/test/jest.setup.ts#L14


I am unable to understand why this has to be increased and does this affect this error. 

@derberg, @magicmatatjahu

peter-rr:
> I just ran the command locally and it works like a charm, one thing though I see it is not generating any spec file, when I do `npm run dev` it gives me this error. Screenshot 2023-03-20 at 12 08 51 PM
> 
> I think we can solve this by generating a spec file to go with the functions.

This should be working fine after updating the `.gitignore` file. Could you check it?

Souvikns (approved):

derberg:
@peter-rr @Souvikns this PR resolves https://github.com/asyncapi/cli/issues/184 imho, and if we want similar stuff for Generator templates, IMHO would be better if @peter-rr opens a new issue really

derberg:
@peter-rr @Souvikns sorry, actually https://github.com/asyncapi/cli/issues/337 is there already, so @peter-rr imho just update PR description that #184 will be resolved with merge of this PR

Souvikns (approved):

PR #430, opened 2023-01-16
feat: add Modelina configuration file support

jonaslagoni:
**Description**
Currently, it's hard to utilize the full feature suite of Modelina, because they are hidden behind callbacks (to allow full customization of model rendering). To enable these features we need to support a basic configuration file, that the user can use to customize the model generation phase when running the `asyncapi generate models` command.

This PR is just a proof of concept to be used for discussing how to proceed.

To enable the configuration file a user can do the following:
```
asyncapi generate models typescript -c="./typescript_model.js"
```

Where the configuration file is:
```js
const {TS_COMMON_PRESET, TypeScriptOptions, IndentationTypes} = require('@asyncapi/modelina');
/** @type {TypeScriptOptions} */
module.exports = {
  enumType: 'union',
  modelType: 'interface',
  indentation: {
    size: 10,
    type: IndentationTypes.SPACES
  },
  mapType: 'record',
  renderTypes: false,
  moduleSystem: 'CJS',
  constraints: {
    modelName: (context) => {
      return `Custom${context.modelName}`;
    },
    propertyKey: (context) => {
      return `custom_prop_${context.objectPropertyModel.propertyName}`;
    }
  },
  typeMapping: {
    Any: (context) => {
      // Always map AnyModel to number
      return 'number';
    }
  },
  presets: [
    {
      preset: TS_COMMON_PRESET,
      options: {
        example: true
      }
    }
  ]
}
```

This gives 100% control of the generators to the user.

**Related issue(s)**
Related to https://github.com/asyncapi/modelina/issues/780

### DIFF
--- a/test/commands/generate/models.test.ts
+++ b/test/commands/generate/models.test.ts
@@ -15,13 +15,23 @@ describe('models', () => {
       expect(ctx.stdout).toMatchSnapshot();
       done();
     });
-    
+
   test
     .stderr()
     .stdout()
-    .command([...generalOptions, 'random', './test/specification.yml', `-o=${ path.resolve(outputDir, './random')}`])
+    .command([...generalOptions, 'random', './test/specification.yml', `-o=${path.resolve(outputDir, './random')}`])
     .it('fails when it dont know the language', (ctx, done) => {
       expect(ctx.stderr).toEqual('Error: Expected random to be one of: typescript, csharp, golang, java, javascript, dart, python, rust\nSee more help with --help\n');
+      expect(ctx.stdout).toEqual('');
+      done();
+    });
+
+  test
+    .stderr()
+    .stdout()
+    .command([...generalOptions, 'typescript', './test/specification.yml', `-o=${path.resolve(outputDir, './ts_config')}`, `-c=${path.resolve(__dirname, './models_configuration/incorrect_models.js')}`])
+    .it('should handle incorrect configuration file correctly', (ctx, done) => {
+      expect(ctx.stderr).toMatchSnapshot();
       expect(ctx.stdout).toEqual('');
       done();
     });
@@ -35,12 +45,12 @@ describe('models', () => {
       expect(ctx.stdout).toMatchSnapshot();
       done();
     });
-    
-  describe('for TypeScript', () => {  
+
+  describe('for TypeScript', () => {
     test
       .stderr()
       .stdout()
-      .command([...generalOptions, 'typescript', './test/specification.yml', `-o=${ path.resolve(outputDir, './ts')}`])
+      .command([...generalOptions, 'typescript', './test/specification.yml', `-o=${path.resolve(outputDir, './ts')}`])
       .it('works when file path is passed', (ctx, done) => {
         expect(ctx.stderr).toEqual('');
         expect(ctx.stdout).toContain(
@@ -48,13 +58,23 @@ describe('models', () => {
         );
         done();
       });
-  });
 
-  describe('for JavaScript', () => {  
     test
       .stderr()
       .stdout()
-      .command([...generalOptions, 'javascript', './test/specification.yml', `-o=${ path.resolve(outputDir, './js')}`])
+      .command([...generalOptions, 'typescript', './test/specification.yml', `-o=${path.resolve(outputDir, './ts_config')}`, `-c=${path.resolve(__dirname, './models_configuration/typescript_models.js')}`])
+      .it('should handle configuration file', (ctx, done) => {
+        expect(ctx.stderr).toEqual('');
+        expect(ctx.stdout).toMatchSnapshot();
+        done();
+      });
+  });
+
+  describe('for JavaScript', () => {
+    test
+      .stderr()
+      .stdout()
+      .command([...generalOptions, 'javascript', './test/specification.yml', `-o=${path.resolve(outputDir, './js')}`])
       .it('works when file path is passed', (ctx, done) => {
         expect(ctx.stdout).toContain(
           'Successfully generated the following models: '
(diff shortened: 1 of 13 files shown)


fmvilas:
I love the approach. Gives a ton of flexibility. Have you thought about how to handle dependencies? I mean, what if in this file I want to import one dependency like `lodash`? Where should it go? A local package.json might be too much given that I may not even be running the command from a Node.js project or a project at all 🤔 Maybe scanning the file first to install the dependencies in their latest version? 🤔

jonaslagoni:
I actually did not think about that, but definitely a valid concern 🤔 Hmm...

At the point where you need to create a node project to run the command, do you not think it will be easier to add your own integration instead of using the CLI? So you [have a script such as:](https://github.com/asyncapi/modelina/blob/next/examples/typescript-generate-example/index.ts)
```js
import { TypeScriptGenerator, TS_COMMON_PRESET } from '@asyncapi/modelina';
import SomeDependency from 'something';

const generator = new TypeScriptGenerator({
  presets: [
    {
      preset: TS_COMMON_PRESET,
      options: {
        example: true
      }
    }
  ]
});

export async function generate(): Promise<void> {
  const models = await generator.generate(input);
  for (const model of models) {
    console.log(model.result);
  }
}
generate();
```

When we extend this feature to the AsyncAPI generator and templates, adding a solution that enables dependencies would be needed anyway... Might need to take a [page out of the AsyncAPI generator](https://github.com/asyncapi/generator/blob/f0daa471c6ea4a91a2c7f18d616c0de06c30dafc/lib/generator.js#L359) 🤔

derberg:
You could always provide a clear instruction on how one can bundle dependencies into the config script. This is how GitHub Actions work now. You bundle all the code into one file that GitHub workflow uses and dist code. So you have instruction, if you want config file to contain some external dependencies then install them and then bundle with `@vercel/ncc`.

Anyway it is more on modelina side, right? not cli. You provide guide there, and just add it as link in `description: 'Path to config file'`

jonaslagoni:
> You could always provide a clear instruction on how one can bundle dependencies into the config script. This is how GitHub Actions work now. You bundle all the code into one file that GitHub workflow uses and dist code. So you have instruction, if you want config file to contain some external dependencies then install them and then bundle with @vercel/ncc.

Can you clarify a bit more about what you mean? How bundling dependencies and code could work?

> Anyway it is more on modelina side, right? not cli. You provide guide there, and just add it as link in `description: 'Path to config file'`

Yea I think so too, see https://github.com/asyncapi/modelina/issues/780

But I want to see what functionality overlaps between CLI and Generator integrations before placing the logic in Modelina. Because arguments and the configuration file are overlapping in the CLI and the generator will probably have multiple configuration files (TBD), so I would rather not place common logic in Modelina before actually having proof it's needed.

I think the best way is to slowly progress and see what works and what does not. Always easy to remove stuff once we see commonalities 🙂

derberg:
> Can you clarify a bit more about what you mean? How bundling dependencies and code could work?

1. you create this:
```
import { TypeScriptGenerator, TS_COMMON_PRESET } from '@asyncapi/modelina';
import SomeDependency from 'something';

const generator = new TypeScriptGenerator({
  presets: [
    {
      preset: TS_COMMON_PRESET,
      options: {
        example: true
      }
    }
  ]
});

export async function generate(): Promise<void> {
  const models = await generator.generate(input);
  for (const model of models) {
    console.log(model.result);
  }
}
generate();
```
2. then you install dependencies
3. then you install vercel bundler and run `ncc build your-config.js --external="@asyncapi/modelina"` (my assumption is that you do not want to bundle modelina in the bundle as it runs inside modelina anyway 😄 )
4. use generated `index.js` as the config that you pass to CLI

of course I did not test ☝🏼 with modelina and cli. Purely theoretical idea 😄 

---

> Because arguments and the configuration file are overlapping in the CLI and the generator will probably have multiple configuration files (TBD)

sorry I didn't get this. What overlaps, the `-c` that you add on CLI? with what exactly?

jonaslagoni:
> of course I did not test ☝🏼 with modelina and cli. Purely theoretical idea 😄

Ahh okay got it! Yea that is definitely another solution 👍 

> sorry I didn't get this. What overlaps, the -c that you add on CLI? with what exactly?

Oftentimes configuration files are used as the baseline, and then arguments overwrite it i.e. a user could do:
```
asyncapi generate models typescript -c="./typescript_model.js" --tsModelType="interface"
```

So even if the config file specified to output models to `class` the argument takes priority. Like this proof of concept does.

derberg:
> Oftentimes configuration files are used as the baseline, and then arguments overwrite it i.e. a user could do:

ah yes, but this can be solved by making these mutually exclusive. So if you use config, other props are ignored.

test/specification.yml
Shurtu-gal, line 1: @jonaslagoni There has been a recent change in the folder structure for tests in #712. This can be moved into https://github.com/asyncapi/cli/tree/master/test/fixtures.

test/commands/generate/models.test.ts
Shurtu-gal, line 1: Similarly, this and other files can be moved into the new folder. https://github.com/asyncapi/cli/tree/master/test/integration/generate

Shurtu-gal (changes requested):

test/commands/generate/models.test.ts
Shurtu-gal: Also the tests have been shifted to mocha now so some functions need to be changed.

Amzani:
@Shurtu-gal @jonaslagoni is this PR still needed?

jonaslagoni:
Its not needed no, once #1376 is merged it will automatically be included once we add it in Modelina 🙂

PR #434, opened 2023-01-20
test: fix tests for model generation

jonaslagoni:
**Description**
With the new update to Modelina the test packages are no longer a valid package name, so just changing it to something that makes more sense


Souvikns:
@jonaslagoni I think you need to reinstall node_modules

jonaslagoni:
I can, but the release workflow will revert the changes anyway, I think 🤔 So it won't really solve anything, other then this PR can be merged 🧐

PR #435, opened 2023-01-20
feat: add kotlin support for model generation

sambhavgupta0705:
**Description**

- ... added kotlin support for model generation.

**Related issue(s)**

Resolves #432

PR #458, opened 2023-02-16
fix: solve typescript models generation error and some failing tests

peter-rr:
**Description**
This fix will solve the typescript models generation error mentioned in the issue below plus some failing unit tests. 

**Related issue(s)**
Relates to #435


test/commands/generate/fromTemplate.test.ts
```diff
@@ -95,7 +95,7 @@ describe('template', () => {
       .it('should print debug logs', (ctx, done) => {
         expect(ctx.stdout).toContain(
           `Template sources taken from ${path.resolve(
-            './test/minimaltemplate'
+            './node_modules/@asyncapi/minimaltemplate'
```
magicmatatjahu, line 98: It probably causes the failing tests. Why did you change path?

peter-rr: Just because it was failing. But it got passed after that change 🤔 Let me take a look at it 👀

peter-rr: After reverting that change now the test is passing. Seems to be failing just at my local but don't understand why. Should I make some change on my local repo folder's structure?

peter-rr:
Failing tests already solved. Ready for review 🚀 
@magicmatatjahu @jonaslagoni

magicmatatjahu (approved):

test/commands/optimize.test.ts
```diff
@@ -84,17 +84,26 @@ describe('optimize', () => {
...
 
+  describe('with no context file', () => {
+    beforeEach(() => {
+      try {
+        testHelper.deleteDummyContextFile();
+      } catch (e) {
+        if (e.code !== 'ENOENT') {
+          throw e;
+        }
+      }
```
magicmatatjahu, line 97: You can create helper (separate function with this body) for that in the `test/testHelper.ts`

peter-rr: Yeah, I was considering that option but finally decided to repeat this piece of code in all those specific tests. Basically, we were in trouble with `deleteDummyContextFile()` since it was called twice for the tests checking that the context file does not exist, and the 2nd call was failing due to the context file had been already deleted. As all the logic was encapsulated in the `testHelper.ts` it was difficult to identify where was exactly happening the problem in the specific tests. Besides, I had to simplify the logic of `deleteDummyContextFile()` cause it was failing randomly due to conflicts related to the management of the execution for the different threads.

Apart from that, I don't think it's a good idea to include that 'try-catch' block in the helper since the errors would be "hidden" in that case and not spotted, so that contributors would end up making use just of the function with the 'try-catch' and ignoring the original 'deleteDummyContextFile()' function, I guess 🤔 

Let's say I'd prefer not to follow DRY principles when it comes to unit tests, something like the following approach: https://dev.to/mbarzeev/why-practicing-dry-in-tests-is-bad-for-you-j7f

Issue #460, opened 2023-03-06
Generate marshalling functions for TypeScript models


jonaslagoni:
#### Reason/Context
In Modelina for TypeScript, you can make it generate the models with marshaling functions and it should also be possible through the CLI.

To do this I suggest we add a flag called `tsMarshalling` as a boolean. If true the generator should include the following options:
```
  presets: [
    {
      preset: TS_COMMON_PRESET,
      options: {
        marshalling: true
      }
    }
  ]
```
Which should be added here: https://github.com/asyncapi/cli/blob/438de0741e9ef29672b6e3af483b90e796903a7d/src/commands/generate/models.ts#L118 as the code example show: https://github.com/asyncapi/modelina/blob/master/examples/typescript-generate-marshalling/index.ts

These will be the generated results: https://github.com/asyncapi/modelina/blob/master/examples/typescript-generate-marshalling/__snapshots__/index.spec.ts.snap

Remember to also add a test here: https://github.com/asyncapi/cli/blob/438de0741e9ef29672b6e3af483b90e796903a7d/test/commands/generate/models.test.ts#L39

Feel free to reach out if you have any questions or get stuck!

jonaslagoni:
@AayushSaini101 there are no assignments or anything in that way, so it's all about just doing it 😊 Go for it 👍

NileshPant1999:
@AayushSaini101 if you are not doing this i will be adding this change will create a pr shortly

NileshPant1999:
this is being discussed here
https://github.com/asyncapi/cli

PR #471, opened 2023-03-12
Feature: Added Support For generating marshalling functions in typesc…

NileshPant1999:
…ript

**Description**
I am trying to solve the issue as per my understanding this is the probable solution if this is right please give green i will update the tests and change the relevant documention after that
- ...
- ...
- ...

**Related issue(s)**
This Will Solve the Issue [https://github.com/asyncapi/cli/issues/460](url)

### DIFF
--- a/src/commands/generate/models.ts
+++ b/src/commands/generate/models.ts
@@ -68,6 +68,9 @@ export default class Models extends Command {
       required: false,
       default: 'default',
     }),
+    tsMarshalling: Flags.boolean({
+      default: false,
+    }),
     /**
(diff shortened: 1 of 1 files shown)


jonaslagoni (changes requested):
Remember to add a test ad well @NileshPant1999 🙂 otherwise it looks good 👍

derberg:
@NileshPant1999 need some help?

derberg:
there are some conflicts
@NileshPant1999 if you want to continue with this one, I suggest opening a new PR and add missing tests

PR #484, opened 2023-03-23
test: fix different test related issues

derberg:


test/commands/context.test.ts
```diff
@@ -6,14 +6,26 @@ import TestHelper from '../testHelper';
...
+  beforeAll(() => {
     testHelper.createDummyContextFile();
   });
 
-  afterEach(() => {
+  afterAll(() => {
     testHelper.deleteDummyContextFile();
   });
 
+  describe('config:context:current', () => {
```
derberg, line 17: this should fix random:
```
EPERM: operation not permitted, open 'D:\a\cli\cli\test.asyncapi'

      27 |
      28 |   createDummyContextFile(): void {
    > 29 |     writeFileSync(DEFAULT_CONTEXT_FILE_PATH, JSON.stringify(this._context), { encoding: 'utf-8' });
         |                  ^
      30 |   }
      31 |   
      32 |   deleteDummyContextFile(): void {
```

because as I can see, the context tests do not really need fresh context file for each test case and we can have it create just once

derberg:
@Souvikns fixing 3 different errors related to testing. Please have a look and lemme know what you think.

Souvikns (approved):

Issue #244, opened 2022-02-23
Studio 404 not found


aykonsvk:
#### Describe the bug

I get this error:
```
$ ./node_modules/.bin/asyncapi start:studio
(node:89108) [MODULE_NOT_FOUND] Error Plugin: @asyncapi/cli: Cannot find module '@oclif/plugin-help/lib/command'
Require stack:
- /Users/m/PhpstormProjects/spec/examples/social-media/node_modules/@fmvilas/oclif-plugin-spaced-commands/lib/hooks/init.js
- /Users/m/PhpstormProjects/spec/examples/social-media/node_modules/@oclif/config/lib/config.js
- /Users/m/PhpstormProjects/spec/examples/social-media/node_modules/@oclif/config/lib/index.js
- /Users/m/PhpstormProjects/spec/examples/social-media/node_modules/@oclif/command/lib/command.js
- /Users/m/PhpstormProjects/spec/examples/social-media/node_modules/@oclif/command/lib/index.js
- /Users/m/PhpstormProjects/spec/examples/social-media/node_modules/@asyncapi/cli/bin/run
module: @oclif/config@1.18.3
task: runHook init
plugin: @asyncapi/cli
root: /Users/m/PhpstormProjects/spec/examples/social-media/node_modules/@asyncapi/cli
See more details with DEBUG=*
(Use `node --trace-warnings ...` to show where the warning was created)
Studio is running at http://localhost:3210?liveServer=3210
Watching changes on file asyncapi.yaml
```

but browser opens up and shows message` 404 - The requested path could not be found`

#### How to Reproduce

- installed locally/globally using `npm install  @asyncapi/cli`
- when I got first error with oclif, I added more packages. 

```
{
  "dependencies": {
    "@asyncapi/cli": "^0.14.1",
    "@oclif/command": "^1.8.16",
    "@oclif/plugin-help": "^5.1.11",
    "oclif": "^2.4.5"
  }
}
```          

- run `./node_modules/.bin/asyncapi start:studio`  or `./node_modules/.bin/asyncapi new` 

the guide to create a new file works and I can verify that a new file was created.

macOS Monterey, node - v17.6.0, npm - 8.5.1

#### Expected behavior

It will open a studio in the browser with the studio.

Souvikns:
[...]

PR #332, opened 2022-09-07
fix: studio 404 not found

to-kr:
Conditional resolve of Studio Build depending on installation type.

**Related issue(s)**
Resolves #244

PR #489, opened 2023-03-27
fix: solve 404 error by better resolving of path to studio

derberg:
Fixes https://github.com/asyncapi/cli/issues/244

Continuation of https://github.com/asyncapi/cli/pull/332 that initial contributor was not interested to continue


derberg:
@Souvikns can you have a look?

Souvikns (approved):

src/models/Studio.ts
```diff
@@ -45,8 +44,13 @@ export function start(filePath: string, port: number = DEFAULT_PORT): void {
   });
 
   const server = createServer((request, response) => {
+    //not all CLI users use npm. Some package managers put dependencies in different weird places
+    //this is why we need to first figure out where exactly is the index.html located 
+    //and then strip index.html from the path to point to directory with the rest of the studio
+    const indexLocation = require.resolve('@asyncapi/studio/build/index.html');
```
Souvikns, line 50: This is looking for a global install studio right?

derberg: for any closest
so if you use CLI as local dependency, this one will be picked up

PR #501, opened 2023-03-31
fix: assets release using oclif

CyberHippo:
**Description**

Update `oclif` dependencies to benefit from https://github.com/oclif/oclif/pull/987 which is missing from `souvikns-oclif` leading to permissions errors.

**Details**

- use [official `oclif` npm package](https://www.npmjs.com/package/oclif) instead of [`souvikns-oclif`](https://www.npmjs.com/package/souvikns-oclif)
- bump versions of `oclif` dependencies
- update syntax for `oclif pack` commands 

**Related issue(s)**

- Fixes #425

PR #514, opened 2023-04-12
fix: adjust release pipeline to node 18

derberg:
Since we did https://github.com/asyncapi/cli/pull/501/files#diff-fd411b4d11d3f5d1d4406fc43b9696941eaec0a37f53b34e04cd1adac4b8e193 we also need to adjust release workflow as it won't pass


derberg:
@Souvikns 🙏🏼

Souvikns (approved):

PR #512, opened 2023-04-09
refactor: update bump-homebrew-formula.yml 

achiverram28:
**Description**

This is the pull request to refactor the `https://github.com/asyncapi/cli/blob/master/.github/workflows/bump-homebrew-formula.yml `. This involves changing from` push` event to `cron` and also running update of brew formula

Verifying :
29135EBB-3103-4EC8-A04B-C2755E65694C_4_5005_c

`Refactors#503`

PR #532, opened 2023-04-24
ci: update brew release workflow to work on schedule

### LINKED ISSUE #503: Do something with automated brew formula update
### Context

whenever we release here, bot also creates PR to brew to update formula -> https://github.com/asyncapi/cli/blob/master/.github/workflows/bump-homebrew-formula.yml

because sometimes we have days when we merge a lot of PRs, we become harmful and spammers to `brew` folks -> https://github.com/Homebrew/homebrew-core/pull/127077

### Solution

I was thinking that maybe we should just change from `push` event to `cron` and it will be ok if we run update of brew formula:
- once a day, end of the day
- and PR is opened on `brew` when there are changes, of course not every day

### IMPORTANT

Before we fix this problem, @Souvikns @magicmatatjahu @boyney123 let us make sure we do not merge more than one `feat` or `fix` a day 🙏🏼 

Thoughts ?

derberg:
Fixes #503 

We cannot wait for https://github.com/asyncapi/cli/pull/512 forever, we need to make sure brew community is no longer affected

### DIFF
--- a/.github/workflows/bump-homebrew-formula.yml
+++ b/.github/workflows/bump-homebrew-formula.yml
@@ -1,28 +1,36 @@
 name: Bump Homebrew formula for CLI
 
 on:
-  # It cannot run on release event as when release is created then version is not yet bumped in package.json
-  # This means we cannot extract easily latest version and have a risk that package is not yet on npm
-  push:
-    branches:
-      - master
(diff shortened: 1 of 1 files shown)


derberg:
@Souvikns now this CI workflow also drops a message to slack in case brew release will fail for some reason

Souvikns (approved):

Issue #361, opened 2022-09-27
`asyncapi --version` should also output version of other asyncapi dependencies


derberg:
So instead of version of AsyncAPI CLI only we should also list generator and others

https://github.com/asyncapi/generator/pull/818/files#diff-c1cc304e56dd03cbb5cf65c32d83c9b7ab069800b2ce780249ac87d320e76789 is one of main reasons. For example in case of generator and templates, it is super important to know that you are using proper generator version. So we need an easy way for CLI user to check that out with one command

Thoughts? @Souvikns @magicmatatjahu @boyney123

Souvikns:
> So we need an easy way for CLI user to check that out with one command

Yeah, I agree with this, also since CLI is implementing other tools, should CLI be able to install specific versions of the tool on the go? So users can change the generator version they want to work with commands like `asyncapi install @asyncapi/generator@1.4.0`

magicmatatjahu:
@Souvikns I don't think so that it's good idea, due to fact that API for given tool can change between version, so for example, in newest parser we switch from singleton parser instance to the Parser as class, so API and also exports from module changed, so calling `parse` was changed.

derberg:
ok, since we agree on that, the question is not on technical details.

afaik `--version` is something we get out of the box from oclif. So in case of this feature, we are talking about overwriting `--version` flag with custom implementation, right?

Souvikns:
What would be better, to have a section in the [release documentation](https://github.com/asyncapi/cli/releases/tag/v0.25.0) talking about the different version of asyncapi tools it depends on, Or have it in CLI's `--version` flag or both?

derberg:
`--version` flag is most important. For release notes we always link PRs, and it is easy to figure out.

derberg:
this one is still very much needed

thulieblack:
This issue is part of the AsyncAPI Bounty Program, interested contributors need to follow the guidelines below to ensure fairness and timely completion of assigned tasks:

### Task Assignment
The assignment of tasks will be prioritized in the following order:

- AsyncAPI maintainers (from any repository)
- Regular contributors (individuals who have merged three or more pull requests)
- Other (if an individual doesn't fall under the above, the maintainer can determine the criteria i.e regular volunteers, etc)

We encouraged everyone to apply as long as the task is for you (*falls under your skill set*). We will not be using the *first comment - get assigned* method for assignments. Instead, we will provide 3 days to consider all individuals interested before assigning any bounty task.

### Deadline
[...]

PR #555, opened 2023-05-12
feat: add `versions` to output versions of AsyncAPI tools used (#361)

aeworxet:
This PR adds to CLI command

```
asyncapi config versions
```
which outputs versions of AsyncAPI tools used, in form
```
@asyncapi/cli/0.44.2 linux-x64 node-v16.17.0
  ├@asyncapi/avro-schema-parser/3.0.0
  ├@asyncapi/bundler/0.3.11
  ├@asyncapi/converter/1.2.0
  ├@asyncapi/diff/0.4.1
  ├@asyncapi/generator/1.10.0
  ├@asyncapi/modelina/1.6.1
  ├@asyncapi/openapi-schema-parser/3.0.0
  ├@asyncapi/optimizer/0.1.18
  ├@asyncapi/parser/2.0.0
  ├@asyncapi/raml-dt-schema-parser/4.0.0
  └@asyncapi/studio/0.17.3

Repository: https://github.com/asyncapi/cli
```

Resolves https://github.com/asyncapi/cli/issues/361

### DIFF
--- a/src/commands/config/versions.ts
+++ b/src/commands/config/versions.ts
@@ -8,10 +8,10 @@ export default class Versions extends Command {
     help: Flags.help({ char: 'h' }),
   };
 
-  private dependencies: string[] = [];
-  private dependency = '';
-
   async run() {
+    const dependencies: string[] = [];
+    let dependency = '';
+
     // Preparation of the array with all dependencies '@asyncapi/*' along with
     // their versions.
     for (const key in this.config.pjson.dependencies) {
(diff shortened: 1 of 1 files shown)


src/commands/config/versions.ts
```diff
@@ -0,0 +1,56 @@
...
+
+export default class Versions extends Command {
+  static description = 'Show versions of AsyncAPI tools used';
+
+  static flags = {
+    help: Flags.help({ char: 'h' }),
+  };
+
+  private dependencies: string[] = [];
+  private dependency = '';
```
derberg, line 12: just curious why we need those outside `run` if they are anyway private and needed only inside the scope of the run?

derberg (commented):
LGTM, just one question left

src/commands/config/versions.ts
aeworxet: Yep. Got carried away by OOP. Changed.

derberg (approved):

PR #556, opened 2023-05-12
feat: add usage documentation for new commands

mhmohona:
**Description**

The script is been added has following workflow - 
- Create and update temp README file
- Generate command help documentation using oclif
- Extract usage information from README
- Update the usage.md file with the extracted usage information and a header
- Append the contents of the README file to the end of the usage file
- Remove the temp README file

**Related issue(s)** fixs: #123

cc: @derberg

### DIFF
--- a/scripts/update_usage_docs.js
+++ b/scripts/update_usage_docs.js
@@ -1,0 +1,29 @@
+const fs = require('fs');
+const util = require('util');
+const { exec } = require('child_process');
+const execPromisified = util.promisify(exec);
+
+const README_PATH = './README.md';
+const USAGE_PATH = '../docs/usage.md';
+
+fs.appendFileSync(README_PATH, '\n\n# Usage\n\n<!-- usage -->\n\n# Commands\n\n<!-- commands -->\n');
+
+execPromisified('oclif readme').then(() => {
+  const header = `---
+title: 'Usage'
+weight: 40
+---
+
+The AsyncAPI CLI makes it easier to work with AsyncAPI documents.
+`;
+
+  fs.writeFileSync(USAGE_PATH, header);
+
+  const readmeContents = fs.readFileSync(README_PATH, 'utf8');
+
+  fs.appendFileSync(USAGE_PATH, `\n${readmeContents}`);
+
+  fs.unlinkSync(README_PATH);
+}).catch((err) => {
+  console.error(err);
+});


derberg (commented):
just had a great call with @mhmohona on how this PR can be improved 💪🏼

mhmohona:
> just had a great call with @mhmohona on how this PR can be improved 💪🏼

Thank you so much for your amaizing guideline @derberg  ^_^

scripts/updateUsageDocs.js
```diff
@@ -0,0 +1,26 @@
+const { writeFile, readFile } = require('fs').promises;
```
derberg, line 1: `/* eslint-disable @typescript-eslint/no-var-requires */` needs to be added in first line

scripts/updateUsageDocs.js
```diff
@@ -0,0 +1,26 @@
...
+// Define an async function to write the header and the README contents to the usage documentation file
+async function run() {
+  await writeFile(USAGE_PATH, header, { encoding: 'utf8' });
+  const readmeContents = await readFile(README_PATH, 'utf8');
+
+  // Append the contents of the README file to the usage documentation file
+  await writeFile(USAGE_PATH, readmeContents, { encoding: 'utf8', flag: 'a' });
+}
+
+run();
```
derberg, line 26: please check the new line error

derberg (changes requested):
left two more comments

I think it would make sense to add some note in readme, in development guide section (if we have one, if not, just create) that the given markdown file should not be edited manually as it is generated by the CI

unless we can just generate inside `usage.md` file some kind of note about it, a note that will be visible in markdown but not on the website, would have to be between `<!-- AUTOGENERATED and info how -->`

what do you think?

derberg:
@mhmohona so what do you thinks about my comment concerning readme or `<!-- AUTOGENERATED and info how -->`

mhmohona:
@derberg oh, i typed the reply and forgot to click on comment button :3 
Yes, its a great idea! I think in script it should be added instead of readme. I

 was also thinking of writing a blog about on which approach I first wanted to go and later on what happened.

derberg:
@mhmohona definitely sounds good but that is a stuff for separate PR for sure

derberg (approved):
LGTM 🚀 

let's merge and see if automation works and usage.md will be updated

PR #568, opened 2023-05-22
Generate example function for TypeScript models

### LINKED ISSUE #462: Generate example function for TypeScript models
#### Reason/Context
In Modelina for TypeScript, you can make it generate the models generate functions that return a basic instance example, this should be possible through the CLI.

To do this I suggest we add a flag called `tsExampleInstance` as a boolean. If true the generator should include the following options:
```
  presets: [
    {
      preset: TS_COMMON_PRESET,
      options: {
        example: true
      }
    }
  ]
```
Which should be added here: https://github.com/asyncapi/cli/blob/438de0741e9ef29672b6e3af483b90e796903a7d/src/commands/generate/models.ts#L118 as the code example shows: https://github.com/asyncapi/modelina/blob/master/examples/typescript-generate-jsonbinpack/index.ts

These will be the generated results: https://github.com/asyncapi/modelina/blob/master/examples/typescript-generate-example/__snapshots__/index.spec.ts.snap

Remember to also add a test here: https://github.com/asyncapi/cli/blob/438de0741e9ef29672b6e3af483b90e796903a7d/test/commands/generate/models.test.ts#L39

Related documentation: https://github.com/asyncapi/modelina/blob/master/docs/languages/TypeScript.md#generate-example-data-function

Feel free to reach out if you have any questions or get stuck!

chinma-yyy:
**Description**

- Added a flag `tsExampleInstance` which if true adds an example of the model generated .
- Added a snapshot test for the same.

**Related issue(s)**
Resolves #462

### DIFF
--- a/src/commands/generate/models.ts
+++ b/src/commands/generate/models.ts
@@ -15,7 +15,7 @@ enum Languages {
   dart = 'dart',
   python = 'python',
   rust = 'rust',
-  kotlin='kotlin'
+  kotlin = 'kotlin'
 }
 const possibleLanguageValues = Object.values(Languages).join(', ');
 
@@ -61,7 +61,7 @@ export default class Models extends Command {
       description: 'TypeScript specific, define the module system to be used.',
       required: false,
       default: 'ESM',
-      
+
     }),
     tsIncludeComments: Flags.boolean({
       description: 'TypeScript specific, if enabled add comments while generating models.',
@@ -77,6 +77,11 @@ export default class Models extends Command {
     }),
     tsJsonBinPack: Flags.boolean({
       description: 'TypeScript specific, define basic support for serializing to and from binary with jsonbinpack.',
+      required: false,
+      default: false,
+    }),
+    tsExampleInstance: Flags.boolean({
+      description: 'Typescript specific, if enabled add example while generating models.',
       required: false,
       default: false,
     }),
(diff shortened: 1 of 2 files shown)


chinma-yyy:
There is a problem in the code formatting and I don't know how to resolve it. Please if anyone can help me with it.

src/commands/generate/models.ts
```diff
@@ -140,95 +145,103 @@ export default class Models extends Command {
...
-      throw new Error(`Could not determine generator for language ${language}, are you using one of the following values ${possibleLanguageValues}?`);
+          })
+        }
+        if (tsJsonBinPack) {
+          presets.push({
+            preset: TS_COMMON_PRESET,
+            options: {
+              marshalling: true
+            }
+          },
```
jonaslagoni, line 164: Do you see that the same preset are added twice now, just with slightly different options? 🙂 

This should not be the case, it can only be added once, and then the options should change slightly based on the parameters 🙂

chinma-yyy: It is not getting added twice even if there are both flags in the command of `tsExampleInstance` and `tsJsonBinPack`. I think Modelina has the support for this where the same preset is added with multiple values of options then it gets combined to give the output. Just the order of the functions generated is getting changed somewhat rest is the same for me.

chinma-yyy: Another approach would be to create an object variable 
```javascript
const options={
    example : tsExampleInstace ,
    marshalling : tsMarshalling
}

presets.push({
    preset: TS_COMMON_PRESET,
    options:options
})
```
and then use this while pushing the presets which can also resolve #460

jonaslagoni: Ahh you right, it's allowed as long as the options don't overlap or default behavior adds some stuff. Because otherwise it will be added twice 😄

Issue #559, opened 2023-05-15
help command is not working


prajjwalkapoor:
#### Describe the bug
`asyncapi help` is not recognized as a command. The functionality for the help is working with `asyncapi --help` but the recommendation says to use `asyncapi help` .

![image](https://github.com/asyncapi/cli/assets/68912239/f5f5a69d-5289-4cbd-9f0d-f21672d50305)

Souvikns:
Yeah, we used to have a help command and then we removed it and now only allow help with the `--help` flag. With the new update with oclif to have an extra help command we have to use a plugin, so by default, we should have only the `--help` flag. 

> https://oclif.io/docs/help_classes

sambhavgupta0705:
@Souvikns so we should remove the suggestion of `help` ?

Souvikns:
I don't think we have any suggestions, for help, I am looking at the [usage docs](https://github.com/asyncapi/cli/blob/master/docs/usage.md) and I don't think we have anywhere specified about help command.

derberg:
yeah, problem is I guess that we do not have control over the warning message that says `asyncapi help`

sambhavgupta0705:
yes I searched for it in the repo but didn't find any instance.

kaushik-rishi:
@Souvikns So we are not planning to add the help command right ?

Souvikns:
Ok so the warning is coming from a package that we use called `@oclif/plugin-not-found` where it creates a log message for `help` command https://github.com/oclif/plugin-not-found/blob/470096ba538e89fea043fa7a712404a450da6204/src/index.ts#L18 I don't think we can change that and it is misleading to the user to have a warning saying them to use `asyncapi help` command which does not exist so we have two options really 

- create a help command 
- it is actually a [hook](https://oclif.io/docs/hooks) and we can write one for our use case and control the warning

what would be a good option?

sambhavgupta0705:
IMHO 
>it is actually a [hook](https://oclif.io/docs/hooks) and we can write one for our use case and control the warning
 
 it is good to have only one command of help

derberg:
I think we should go with the hook

kaushik-rishi:
@derberg @Souvikns can i go ahead and work on this ?

sambhavgupta0705:
@kaushik-rishi I would also like to work on this ,have gone through oclif docs and cli structure

kaushik-rishi:
[...]

PR #593, opened 2023-05-25
fix: help command

sambhavgupta0705:
**Description**

- ...
- ...
- ...

**Related issue(s)**

#559


package.json
```diff
@@ -105,10 +104,13 @@
   "main": "lib/index.js",
   "oclif": {
     "commands": "./lib/commands",
+    "hooks": {
+      "command_not_found": [
+        "./lib/hooks/command_not_found/example"
```
Souvikns, line 109:
```suggestion
        "./lib/src/errors/command-not-found"
```
I think this should be the path of the hook, not really sure, looking at this example it looks like that

Souvikns (commented):
I think you should create a new folder called hooks, instead of adding this in the errors folder

package.json
kaushik-rishi: @Souvikns i've added the hook in `hooks/command-not-found` folder, it should work right ?

Souvikns: I think that should be a file path, check that after building the project does that path has any file

kaushik-rishi:
@Souvikns Just a quick question before i bail out, how are we supposed to prevent the 
`@oclif/plugin-not-found` from triggering, inside the hook that we are working with ?

Souvikns:
The plugin we are using right now is actually a hook, can you can use that as an example https://github.com/oclif/plugin-not-found

kaushik-rishi:
> The plugin we are using right now is actually a hook, you can use that as an example https://github.com/oclif/plugin-not-found

But we would still need the `oclif/plugin-not-found` for any other user entered commands which donot exist right ?

We basically need this new hook on top of oclif/plugin-not-found as far as i get it. please correct me.

package.json
kaushik-rishi: Yeah, it's just file paths that we have to link back in `package.json`. It's working build.

I just separated the hooks into folders based on events for further clarification if anyone wants to add new hooks.

sambhavgupta0705:
> The plugin we are using right now is actually a hook, you can use that as an example https://github.com/oclif/plugin-not-found

yes I read the blog and the hook was setup inside the plugin

Souvikns:
I don't think we need the old plugin we are replacing it with our custom hook

kaushik-rishi:
> image This is the output @Souvikns

https://github.com/asyncapi/cli/blob/479c31d8c4652831b0c6472ba49a77e8cf7b189d/src/hooks/command_not_found/help-command-handler.ts#L4

I just logged the options passed for reference. You can build on top of this.

sambhavgupta0705:
@kaushik-rishi I think we should not console.log this whole message and just console `help command doesn't exist try --help`

Souvikns:
That output is comping from logging the opts right, now you just have to have a custom warning about it. I think if you see that log that means that the hook is connected to the CLI

kaushik-rishi:
@sambhavgupta0705 please remove the errors folder. just reminding.

sambhavgupta0705:
> @sambhavgupta0705 please remove the errors folder. just reminding.

I didn't get it which errors folder

sambhavgupta0705:
@Souvikns the PR is ready for review

kaushik-rishi:
package-lock.json shouldn't technically change.

As we didn't install or uninstall any package I think we can safely revert package-lock.json changes.

kaushik-rishi:
> I don't think we need the old plugin we are replacing it with our custom hook

@Souvikns need your suggestions 🙂

i understand, but what about the fallback for other commands, the hook developed here is handling just the `help` command and suggesting the user to use `--help`. 
What if i use something like `asyncapi generte` (intentional typo), we would still need the `plugin-not-found` to give us a fallback option to suggestions from the plugin right ?

image


----

Also, @sambhavgupta0705 can we rather find a way to just bail out and exit after the console.log in the hook, not to forward the entire case to `plugin-not-found` module ?

image

sambhavgupta0705:
Yes we can add .exit functionality here.

kaushik-rishi:
@sambhavgupta0705 for your reference.
follow the last line of this documentation https://oclif.io/docs/hooks.
you can use `this.exit()` to exit out instead of going further.

let @Souvikns confirm if this is what we are looking forward to do.

sambhavgupta0705:
@kaushik-rishi I tried adding `this.exit()` but the plugin console was always get printed

kaushik-rishi:
> @kaushik-rishi I tried adding `this.exit()` but the plugin console was always get printed

It is probably due to asynchronous nauigure, i'm afraid i don't have a solution for this right away.

I can explain the situation better.
May be it's like once the hook is triggered both the plugins (the one we wrote and the plugin-not-found) and being called at the same time. It is exiting fine but before exiting it's printing the first line of the `plugin-not-found`s prompt.

sambhavgupta0705:
Yess

kaushik-rishi:
> Yess

I have a couple of ideas, Let @Souvikns catch up and review the discussion uptill now. 🙂

We can easily make our own hook for handling every command that isn't found. Let's carry it further.

Souvikns:
Yeah, the plugin used to suggest or at least try to suggest an alternative command if one is wrong, I think we can build this feature, you can take inspiration from the plugin if you want.

src/hooks/command_not_found/help-command-handler.ts
```diff
@@ -0,0 +1,9 @@
+import {Hook} from '@oclif/core';
+
+const helpCommandHandlerHook: Hook<'command_not_found'> = async function (opts) {
+  if (opts.id === 'help') {
+    console.log('help command doesn\'t exist try --help');
+  }
+};
```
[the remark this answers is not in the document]
Souvikns: Will this work, I mean if the control ever goes to the plugin it will say to use `asyncapi help` command and that does not exist.

package-lock.json
kaushik-rishi, line 1: please revert the `package-lock.json` changes.

Souvikns:
You write the code, and you can take inspiration from it, and write it in your own way, I don't know about the license, @derberg are we breaking any rules here?

Souvikns:
@sambhavgupta0705 @kaushik-rishi, @derberg is not available right now, you can go ahead and start building the suggestion feature, no need to reuse but if you want you can take inspiration from the existing library.

kaushik-rishi:
Sure @Souvikns

sambhavgupta0705:
> @sambhavgupta0705 @kaushik-rishi, @derberg is not available right now, you can go ahead and start building the suggestion feature, no need to reuse but if you want you can take inspiration from the existing library.

Yup just busy with some work,will start working on this by this week.

sambhavgupta0705:
hey @kaushik-rishi ,what other changes do we need to do?

kaushik-rishi:
> hey @kaushik-rishi ,what other changes do we need to do?

Hey @sambhavgupta0705 
Was a bit busy since a couple of days.
We can take inspiration from the [`@oclif/not-found`](https://github.com/oclif/plugin-not-found) and recreate some workflow to let the user know that they've entered a command that is not yet registered.

derberg:
`oclif` is done under MIT license, so it is definitely fine to copy some code. Just attribute original creators in the readme

Souvikns:
@sambhavgupta0705 let us know if anything is blocking you or if you need any help.

sambhavgupta0705:
@Souvikns just got busy with mentorship program work .I will soon find a way to resolve this ASAP

Souvikns:
@sambhavgupta0705 any updates on this?

sambhavgupta0705:
ohh sorry I totally forgot about this

Souvikns:
@sambhavgupta0705 can you resolve the conflict

peter-rr:
@Souvikns Should we close this PR as the bug related is already solved by https://github.com/asyncapi/cli/pull/843 ?

derberg:
closing as looks like not needed

PR #598, opened 2023-05-26
feat: add Dockerfile and its release pipeline

### LINKED ISSUE #358: Add Dockerfile and release pipeline
- example Dockerfile that we already have for the Generator CLI -> https://github.com/asyncapi/generator/blob/master/Dockerfile
- but instead of installing npm package, and using CLI from npm, we can actually build even smaller image, not based on node alpine. We would just plug into https://github.com/asyncapi/cli/blob/master/.github/workflows/upload-release-assets.yml workflow where we generate different distro versions, including Linux, and use linux binary in docker? Just an idea 🤷🏼 
- make sure it is plugged into the release flow. Example -> https://github.com/asyncapi/generator/blob/master/.github/workflows/release-docker.yml

Savio629:
**Description**

Draft PR
-Added Docker using npm (researching using linux-curl)
-Added worflow for docker
-Added docker:build script

**Related issue(s)**
Resolves #358


Savio629:
@derberg

derberg:
@Savio629 so far so good, just solve merge conflicts

are you still exploring alternative solution? 

I was also wondering if you could add https://github.com/marketplace/actions/docker-hub-readme-description-sync as another step in the automation, after pushing to docker? as apparently docker do not pick github readme by default

Savio629:
I'm exploring the alternative solution, I'll update you by tomorrow.
I'll solve the merge conflicts and update the workflows as mentioned.
Have you gone through the workflow? I have updated the part which was deprecated.

derberg:
@Savio629 can you also have a look on sonarcloud complain about security?

Savio629:
@derberg Fixed one of the warnings but the second warning tells me to run the dockerfile in `nonroot user` but while doing so git, chromium and aysncapi/cli are not able to install due to permissions...
I tried to add USER command after installing but then while running the image it throw an error regarding the permissions.

derberg:
@Savio629 but sonar said to use `--ignore-scripts` not `--install-scripts`

Savio629:
Yeah I updated the pr.
Mistyped...

Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+# Since 0.14.0 release of html-template chromium is needed for pdf generation.
+# More custom packages for specific template should not be added to this dockerfile. Instead, we should come up with some extensibility solution.
+RUN apk --update add git chromium && \
+    rm -rf /var/lib/apt/lists/* && \
+    rm /var/cache/apk/*
+
+# Installing latest released npm package
+RUN npm install --ignore-scripts -g @asyncapi/cli
+
+ENTRYPOINT [ "asyncapi" ]
```
derberg, line 17:
```suggestion
FROM node:16-alpine

# Create a non-root user
# Addressing sonar cloud: The node image runs with root as the default user.
RUN addgroup -S myuser && adduser -S myuser -G myuser

WORKDIR /app

# Since 0.14.0 release of html-template chromium is needed for pdf generation
ENV PUPPETEER_EXECUTABLE_PATH /usr/bin/chromium-browser
ENV PUPPETEER_SKIP_CHROMIUM_DOWNLOAD true

# Since 0.30.0 release Git is supported and required as a dependency
# Since 0.14.0 release of html-template chromium is needed for pdf generation.
# More custom packages for a specific template should not be added to this Dockerfile. Instead, we should come up with some extensibility solution.
RUN apk --update add git chromium && \
    rm -rf /var/lib/apt/lists/* && \
    rm /var/cache/apk/*

# Installing the latest released npm package
RUN npm install --ignore-scripts -g @asyncapi/cli

# Switch to the non-root user
USER myuser

ENTRYPOINT [ "asyncapi" ]
```

this should solve the security issue

Savio629: Yeah i tried the same method but then while running the image it throws permission error
@derberg

Savio629: Found a fix!

derberg: interesting, I tried it on local and there were no errors, anyway, I see all green now 🚀

derberg (approved):

derberg:
/rtm

Thanks 🙏

.github/workflows/release-docker.yml
```diff
@@ -0,0 +1,37 @@
...
+
+      - name: Release to Docker
+        run: | 
+          echo ${{secrets.DOCKER_PASSWORD}} | docker login -u ${{secrets.DOCKER_USERNAME}} --password-stdin
+          npm run docker:build
+          docker tag asyncapi/cli:latest asyncapi/cli:${{ steps.version.outputs.value }}
+          docker push asyncapi/cli:${{ steps.version.outputs.value }}
+          docker push asyncapi/cli:latest
+
+      - name :  Sync README.md and Description to Docker Hub
```
Ferror, line 29: Does it really need a secund checkout action?

Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+ENV PUPPETEER_SKIP_CHROMIUM_DOWNLOAD true
+# Since 0.30.0 release Git is supported and required as a dependency
+# Since 0.14.0 release of html-template chromium is needed for pdf generation.
+# More custom packages for specific template should not be added to this dockerfile. Instead, we should come up with some extensibility solution.
+RUN apk --update add git chromium && \
+    rm -rf /var/lib/apt/lists/* && \
+    rm /var/cache/apk/*
+
+# Installing latest released npm package
+RUN npm install --ignore-scripts -g @asyncapi/cli
```
Ferror, line 19: I think the cli version should be configurable, because then you don't know which version is inside the container

Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+
+# Switch to the non-root user
+USER myuser
+
+ENTRYPOINT [ "asyncapi" ]
```
Ferror, line 24: Empty line missing

package.json
```diff
@@ -177,4 +178,4 @@
     "get-version": "echo $npm_package_version"
   },
   "types": "lib/index.d.ts"
-}
+}
```
Ferror, line 181: Empty line missing

derberg:
@Ferror the PR is already merged, please open up a followup to improve it 🙏🏼

Issue #549, opened 2023-05-03
Attempting to generate Rust models produces an inscrutable error message


coriolinus:
#### Describe the bug

Attempting to generate Rust models produces an inscrutable error message.

#### How to Reproduce

Version: `@asyncapi/cli/0.40.4 wsl-x64 node-v16.15.1`

```sh
$ asyncapi generate models rust openapi/render_service/v1/render_service.openapi.yaml
Error: EEXIT: 1
```
#### Expected behavior

Either successful generation of models in Rust, or some kind of human-readable explanation of what went wrong.

kaushik-rishi:
@coriolinus can you share the file for which you are trying to generate the models ?

coriolinus:
That file is still closed-source for now, unfortunately.

kaushik-rishi:
@coriolinus oh no problem!
can you try the same command with the file mentioned in the asyncapi docs.
[this one](https://gist.github.com/kaushik-rishi/f9e7ae0f0e49666995ac37607309e56d)

coriolinus:
That does appear to work: 

```bash
$ asyncapi generate models rust asyncapi.yaml
Thu 25 May 2023 09:08:03 AM CEST
Successfully generated the following models:
## Model name: AnonymousSchema1
// AnonymousSchema1 represents a AnonymousSchema1 model.
#[derive(Clone, Debug, Deserialize, Serialize)]
pub struct AnonymousSchema1 {
    #[serde(rename="id", skip_serializing_if = "Option::is_none")]
    pub id: Option<i32>,
    #[serde(rename="lumens", skip_serializing_if = "Option::is_none")]
    pub lumens: Option<i32>,
    #[serde(rename="sentAt", skip_serializing_if = "Option::is_none")]
    pub sent_at: Option<String>,
    #[serde(rename="additionalProperties", skip_serializing_if = "Option::is_none")]
    pub additional_properties: Option<std::collections::HashMap<String, serde_json::Value>>,
}
```

Though I note that the model isn't quite proper; the `#[serde(skip_serializing_if)]` annotation handles serialization, but not deserialization. You'd want `#[serde(default)]` as well so that the model can symmetrically deserialize its serialization output. 

Digression aside, it looks like the problem is related to the input file I gave it. It's not syntactic--the Zally linter passed my input file--but something there caused the Rust generator to abort without error messages.

coriolinus:
Here's a moderately well minified example which shows the failure:

```yaml
[...]

PR #618, opened 2023-05-31
feat: helper message for `invalid` specification file during model generation

### LINKED ISSUE #616: CLI exits  the process on invalid file input rather than outputing the error message
#### Describe the bug
When using the `generate` command for `models` , if an invalid specification file is provided then the cli exits with status code 1 but does not stdout any specific message about what error happened. 

#### How to Reproduce
Steps to reproduce the issue. Attach all resources that can help us understand the issue:
- Use command `asyncapi generate models typescript asyncapi.yml`

- Screenshots

image

- Link to GitHub repository with project that has issues
- Files that can help reproduce the issue, like your AsyncAPI file, either paste inside the issue in a code block or in a sharable [gist](https://gist.github.com/)
File used 

```yml
asyncapi: '2.5.0'
info:
  title: Streetlights API
  version: '1.0.0'
  license:
    name: Apache 2.0
    url: 'https://www.apache.org/licenses/LICENSE-2.0'
server:
  mosquitto:
    url: mqtt://test.mosquitto.org
    protocol: mqtt
channels:
  light/measured:
    publish:
      summary: Inform about environmental lighting conditions for a particular streetlight.
      operationId: onLightMeasured
      message:
        name: LightMeasured
        payload:
          type: object
          properties:
            id:
              type: integer
              minimum: 0
              description: Id of the streetlight.
            lumens:
              type: integer
              minimum: 0
              description: Light intensity measured in lumens.
            sentAt:
              type: string
              format: date-time
              description: Date and time when the message was sent.

```
*Here the file is invalid because the server parameter is at wrong location*
#### Expected behavior

- An error message be displayed something like : `Generator Error: Input is not a corrent AsyncAPI document so it cannot be processed.`
- This will result in understanding what the error is about.

## Additional Context
- The issue is in this line 
https://github.com/asyncapi/cli/blob/9c3bde73ba9e62a1b92dbda0d3c217aa94c92626/src/parser.ts#L83-L84
The issue is that instead of only returning the status of the document it is exiting the process. Rather the approach here https://github.com/asyncapi/cli/blob/9c3bde73ba9e62a1b92dbda0d3c217aa94c92626/src/commands/generate/models.ts#L146-L148 is that we handle the status and make an output of the reason.

chinma-yyy:
**Description**

- Added error logging for invalid specification file

image

**Related issue(s)**
Resolves #616 
Fixes https://github.com/asyncapi/cli/issues/549

### DIFF
--- a/src/commands/generate/models.ts
+++ b/src/commands/generate/models.ts
@@ -2,7 +2,7 @@ import { CSharpFileGenerator, JavaFileGenerator, JavaScriptFileGenerator, TypeSc
 import { Flags } from '@oclif/core';
 import Command from '../../base';
 import { load } from '../../models/SpecificationFile';
-import { parse, validationFlags } from '../../parser';
+import { formatOutput, parse, validationFlags } from '../../parser';
 
 import type { AbstractGenerator, AbstractFileGenerator } from '@asyncapi/modelina';
 
@@ -142,8 +142,10 @@ export default class Models extends Command {
     const { tsModelType, tsEnumType, tsIncludeComments, tsModuleSystem, tsExportType, tsJsonBinPack, namespace, csharpAutoImplement, csharpArrayType, csharpNewtonsoft, csharpHashcode, csharpEqual, csharpSystemJson, packageName, output } = flags;
     const { language, file } = args;
     const inputFile = (await load(file)) || (await load());
-    const { document, status } = await parse(this, inputFile, flags);
+    const { document, diagnostics ,status } = await parse(this, inputFile, flags);
     if (!document || status === 'invalid') {
+      const severityErrors = diagnostics.filter((obj) => obj.severity === 0);
+      this.log(`Input is an empty file or not a correct AsyncAPI document so it cannot be processed.${formatOutput(severityErrors,'stylish','error')}`);
       return;
     }
 
--- a/src/commands/generate/models.ts
+++ b/src/commands/generate/models.ts
@@ -145,7 +145,7 @@ export default class Models extends Command {
     const { document, diagnostics ,status } = await parse(this, inputFile, flags);
     if (!document || status === 'invalid') {
       const severityErrors = diagnostics.filter((obj) => obj.severity === 0);
-      this.log(`Input is an empty file or not a correct AsyncAPI document so it cannot be processed.${formatOutput(severityErrors,'stylish','error')}`);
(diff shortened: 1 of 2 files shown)


kaushik-rishi:
I don't feel it's a good idea to display the warnings, infos, hints, if we're not actually even considering the error components with other severity. Let me check how the logger's working and i'll get back in a bit @chinma-yyy 

But rest of it LGTM.

src/commands/generate/models.ts
```diff
@@ -142,8 +142,10 @@ export default class Models extends Command {
     const { tsModelType, tsEnumType, tsIncludeComments, tsModuleSystem, tsExportType, tsJsonBinPack, namespace, csharpAutoImplement, csharpArrayType, csharpNewtonsoft, csharpHashcode, csharpEqual, csharpSystemJson, packageName, output } = flags;
     const { language, file } = args;
     const inputFile = (await load(file)) || (await load());
-    const { document, status } = await parse(this, inputFile, flags);
+    const { document, diagnostics ,status } = await parse(this, inputFile, flags);
     if (!document || status === 'invalid') {
+      const severityErrors = diagnostics.filter((obj) => obj.severity === 0);
+      this.log(`Input is an empty file or not a correct AsyncAPI document so it cannot be processed.${formatOutput(severityErrors,'stylish','error')}`);
```
Souvikns, line 148: @chinma-yyy looking at your screenshot I have a question why does it say that input is an empty file because it is not, you are showing warning on whats wrong with the file, so it is not an empty file. 

What I am trying to say is, you have two situations on error 
1. path file is wrong and the spec does not exist 
2. path file is right but there is some issue with the file. 

I think you should change the error message.

Souvikns (changes requested):

kaushik-rishi:
@chinma-yyy Can you update the screenshot in the PR ?

src/commands/generate/models.ts
chinma-yyy: Yup, thought that the `document` object would be undefined if the file is empty that's why we have made the `!document` condition there. Anyways I have made the changes

Souvikns (approved):
LGTM

derberg (approved):
added to description that https://github.com/asyncapi/cli/issues/549 will also be fixed, thanks

PR #646, opened 2023-06-12
feat: diff will show error on breaking changes

aayushmau5:
**Description**

`diff` command will now show error(with exit code 1) when breaking changes are detected. This is the default behaviour.
There's also a flag `--no-error` you can provide to not show this error.
Ex.
```
diff ./old.yml ./new.yml --no-error
```

The current behaviour is that diff will show error no matter the `type` of changes required. For example, 
```
diff ./old.yml ./new.yml --type=non-breaking
```
will show a breaking change error in case there are breaking changes detected between provided documents. 

**Related issue(s)**

None

### DIFF
--- a/src/commands/diff.ts
+++ b/src/commands/diff.ts
@@ -8,6 +8,7 @@ import Command from '../base';
 import { ValidationError } from '../errors/validation-error';
 import { SpecificationFileNotFound } from '../errors/specification-file';
 import {
+  DiffBreakingChangeError,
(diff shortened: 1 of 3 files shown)


aayushmau5:
cc @jfcote You can try out the changes.

JFCote (approved):
LGTM!

Souvikns (approved):

aayushmau5:
@Souvikns when can we merge this?

PR #673, opened 2023-07-01
ci: upgrade release workflow to new version

### LINKED ISSUE #605: Need for urgent changes in GitHub Actions automation
This issue defines a list of tasks that need to be performed in this repo to make sure it's ci/cd automation works long term without any issues.

It is up to maintainers to decide if it must be addressed in one or multiple PRs.

Below are 3 different sections describing 3 different important ci/cd changes.

**IMPORTANT-START** 
For GitHub workflows that contain `This workflow is centrally managed in https://github.com/asyncapi/.github/` you do not have to perform any work. These workflows were already updated through the update in `.github`. The only exception is the workflows related to nodejs release. More details in **Upgrade Release pipeline - in case of nodejs projects** section
**IMPORTANT-END** 

### Deprecation of way data is shared between steps

Every single GitHub Action workflow that has `echo "::set-output name={name}::{value}"` need to be updated to follow `echo "{name}={value}" >> $GITHUB_OUTPUT`

We do not yet know when `set-output` will stop working. Previous disable date was 31.05 but now then say community needs more time.

[For more details read official article from GitHub](https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/)

### Deprecation of node12

> 2nd bullet point is still relevant for you even if your projects in not nodejs project

- Every single workflow that uses [setup-node](https://github.com/actions/setup-node) action needs an update to follow `v3` version of this action, and make sure minimum node 14 is used
- Now this part is more complex. Problem with `node12` is that node-based GitHub Actions were using it in majority as a runtime environment. Look for example at [this action.yaml file for setup-node action v2](https://github.com/actions/setup-node/blob/v2/action.yml#L38). So the job that you have to do is go through all the workflows, and verify every single action that you use, make sure you are using the latest version that is not based on `node12`. I already did review a lot of actions as part of [this PR](https://github.com/asyncapi/.github/pull/226) so maybe you will find some actions there and can copy from me. For example `action/checkout` needs to be updated to v3.

Node12 end of support in action is probably September 27th.

[For more details read official article from GitHub](https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/)

### Upgrade Release pipeline - in case of nodejs projects

> ignore this section if your project is not nodejs project

You have 2 options. You can:

**A.** choose to switch to new release pipeline using instruction from https://github.com/asyncapi/.github/issues/205

**B.** stay with old release pipeline, and manually update GitHub workflows and actions used in it, you can inspire a lot from this PR https://github.com/asyncapi/.github/pull/226

I definitely recommend going with **A**

Workflows related to release:
- .github/workflows/if-nodejs-release.yml
- .github/workflows/if-nodejs-version-bump.yml
- .github/workflows/bump.yml

chinma-yyy:
**Description**

- Followed the steps mentioned in asyncapi/.github/issues/205] and removed the dependencies   

**Related issue(s)**

Resolves #605


Souvikns (approved):
This part looks alright, we now just have to add the topics in the repo right?

Souvikns:
I have added the topics needed, @derberg let me know if we are missing something,

PR #674, opened 2023-07-01
feat: add `tsExampleInstance` flag and test for generate models 

### LINKED ISSUE #462: Generate example function for TypeScript models

chinma-yyy:
**Description**

- Added `tsexampleInstance` flag which will add the example option to the model
- Created an `options` object which can be reused and declared once based on the flags
- Added the relevant test script and updated the Docs

**Related issue(s)**

Resolves #462


chinma-yyy:
@jonaslagoni  Will need help on the part of description of the flag.

chinma-yyy:
> @jonaslagoni Will need help on the part of description of the flag.

@jonaslagoni !!

jonaslagoni:
@chinma-yyy what do you need help with? 🙂 

Looks fine to me 👍

derberg:
> @jonaslagoni waiting !

@chinma-yyy were you shouting? 😅

Issue #574, opened 2023-05-22
TS Error when run generation


Tenischev:
#### Describe the bug
TypeScript error when run generation from template/

#### How to Reproduce
1. Install node (tested with 16 and 18)
2. Install latest cli
3. Checkout asyncapi/java-spring-template
4. Run command `asyncapi generate fromTemplate tests/mocks/mqtt.yml @asyncapi/java-spring-template -o spring-output-test`
5. Error:
```
error TS6046: Argument for '--module' option must be: 'none', 'commonjs', 'amd', 'system', 'umd', 'es6', 'es2015', 'es2020', 'es2022', 'esnext', 'node12', 'nodenext'.
```

Additional details:
OS - Windows 11
```
asyncapi --version
@asyncapi/cli/0.40.8 win32-x64 node-v16.20.0
```

#### Expected behavior
No error when run generation from template

derberg:
@Souvikns I think you have access to windows, can you help with this issue 🙏

Tenischev:
here step by step drill down of commands, maybe this will help
![image](https://github.com/asyncapi/cli/assets/4137916/ab836439-f98e-47e9-936f-c373784c6de7)

Souvikns:
OK, I am getting the same error on Mac as well.

Souvikns:
The command is running on the codebase without any issue, but in production, it is not working, I checked and all other commands are working, might be something wrong with the generator.

Not a fix but you can run the command by passing `--module` flag run `asyncapi generate fromTemplate --module=none <rest of the parameters>`

Tenischev:
> Not a fix but you can run the command by passing `--module` flag run `asyncapi generate fromTemplate --module=none <rest of the parameters>`

Nope
![image](https://github.com/asyncapi/cli/assets/4137916/f802a3b6-a881-4a5f-b529-34024109a593)

derberg:
generator do not support `--module`, the error comes like from TS compiler I think

Souvikns:
@Tenischev can you install typescript globally and try to run the command, it worked for me. 

<img width="756" alt="Screenshot 2023-05-23 at 4 19 11 PM" src="https://github.com/asyncapi/cli/assets/41781438/88f45957-e3da-4b9a-87dd-196069d69c50">

derberg:
does it mean that `typescript` has to be moved to `dependencies`? it shouldn't be needed in production right? `typescript` should be required only in build time

Souvikns:
[...]

PR #681, opened 2023-07-14
fix: update dependencies

Souvikns:
**Description**

This PR is just updating the dependencies for CLI. 

**Related issue(s)**

see also #574


derberg:
just realized it might not be enough and we need the same for `generator` repo, but yeah, lets see

derberg (approved):

Issue #599, opened 2023-05-29
Prepare CLI for v1.0.0 release 


Souvikns:
> This issue is meant for the AsyncAPI mentorship program. 

This is an extension of https://github.com/asyncapi/cli/issues/251, as we are done with most of the work needed for v1. This project has smaller chunks of tasks instead of one big feature and most work is related to documentation and writing tests. 

### Scope of the project 
We are going to release our first stable version of CLI as v1 and we need some help with it like 

-  reach 90% test coverage
- enable cross-platform installation- (only windows chocolatey)
- Documentation update - (update usage docs for all the commands with examples)
- Simple Integration tests to prove the binary works

### Technologies Required 
- Typescript 
- nodejs 
- oclif
- jest

derberg:
> Documentation update - (update usage docs for all the commands with examples)

fyi this will be automated with https://github.com/asyncapi/cli/pull/556

also join https://github.com/asyncapi/cli/issues/573 as we need to decide on how we release. Majors can be more often and might be that before someone completes mentorship v1 or even v2 will be there already. So rethink title, maybe not prepare for v1 but maybe call it "hardening" or whatever name is not related to v1 😄

Priyansh61:
Hi @Souvikns @derberg , 

Iwould like to work on this proposal idea. Can you please give me some guidelines for writing the proposal for this project?

Thanks,

manulpatel:
Hello @derberg @Souvikns! I am interested in being a part for v1.0.0 release of cli. I would be very happy to work on this and team up with other mentees as a part of the Mentorship program. It seems the major task remaining is the integration of `cupid `in cli. Could you give some hints on any specifics required in in the proposal?

AceTheCreator:
Hey Folks, 

To clarify how to submit a proposal, you should contact the project Mentor/s on Slack and send them your submission in the form of a document or ask them how best they'd prefer you submit your proposals.

tusharrao198:
Hi @Souvikns @derberg . I am interested in working on this issue. Can you please guide me on how to start ?

codingmickey:
> Hi @Souvikns @derberg . I am interested in working on this issue. Can you please guide me on how to start ?

+++

Shurtu-gal:
> This issue has been automatically marked as stale because it has not had recent activity 😴
> 
> It will be closed in 120 days if no further activity occurs. To unstale this issue, add a comment with a detailed explanation.
> 
[...]

PR #353, opened 2022-09-21
fix: switch from mocha and chai to the jest

magicmatatjahu:
**Description**

- switch from `mocha` and `chai` to the `jest`. Simple reason. Mocha and chai (specifically, a `ts-node` which transpiles code from TS to JS) has a some problems with importing ESM. I found that problems when I started writing tests for https://github.com/asyncapi/cli/pull/347 and new parser uses new module syntaxes. We still need `chai` in devDependencies, because `@oclif/test` uses it underneath.
- bump `@oclif/core` to the latest `^1.18.0` version.

**Related issue(s)**
See more https://github.com/asyncapi/cli/pull/347

PR #712, opened 2023-08-02
test: change folder structure

Shurtu-gal:
**Description**

- I have made this PR considering that I also have to add unit and integration tests.
- Before, as there were only functionality tests, I put them in the functionality folder.
- These refactoring will help in adding new tests under other categories.

**Related issue(s)**
#599


Souvikns (commented):
I see you have moved the tests to a different folder, can you update the pr comment and talk about why you are making this change and how it will benefit us?

Shurtu-gal:
@Souvikns, I have updated the comment.

Shurtu-gal:
@Souvikns @fmvilas @derberg I have to decide whether I should keep using jest for testing purposes or use mocha for it. I would love your opinions. 

Reasons for not using mocha :- https://github.com/asyncapi/cli/pull/353

Reasons for shifting to mocha:- 
 - It has better support for oclif. Hence gives more accurate coverage

#### With Mocha
image

#### With Jest 
image


Some alternatives can be to mock the commands as functions and test them instead of running them as cli. But I don't have a viable approach for doing so ( Would love your guidance out here 😮‍💨 )

chinma-yyy:
The issue is with the code transpiling. I think we need not shift to mocha instead we can change the collect coverage parameter to live/**/*.js which will collect coverage from the compiled code which will give us the accurate result. Would love to hear on this more. I tried changing the transformer and some config didn't find something useful

Shurtu-gal:
@chinma-yyy Let me also try then 🤖

Shurtu-gal:
> LGTM
> after merging just please go through remaining 10PRs and add comment to the ones that add tests, with info what people have to do to update and follow latest `master`

👍

derberg:
@Souvikns @Shurtu-gal I guess it can be merged? Feel free to do so when ready

peter-rr (commented):
LGTM!

Issue #751, opened 2023-08-09
Failing context operations like add, use, list, etc.


AnishKacham:
Context operations do not work even after initializing context

#### How to Reproduce
Create an empty directory, and spin up a sample asyncapi spec document.
for this I used the sample [mentioned]( https://gist.github.com/AnishKacham/f7cf6aadaa11c436d40f9910bef51a60
) in the [docs](https://www.asyncapi.com/docs/tutorials/create-asyncapi-document) :

`asyncapi config context init`
you should see a `.asyncapi-cli` file get created with the following content: 
```
{"store":{}}
```
now running `asyncapi config context add foo ./asyncapi.yaml` gives the following message even if context is initialized already
```
You have no context file configured. Run "asyncapi config context init" to initialize it.
```
- Screenshots
![image](https://github.com/asyncapi/cli/assets/79566582/c6a11442-a857-4764-b399-4817a1545126)

- Link to GitHub repository with project that has issues
  - https://github.com/asyncapi/cli
  
#### Expected behavior
- Context should be added to the `.asyncapi-cli` file

derberg:
@aeworxet hey any ideas?

AnishKacham:
update on this issue. The command works perfectly fine if I'm building cli locally and then using it as:
 ```
./bin/run config context add foo .\asyncapi.yaml
```

derberg:
@AnishKacham when you tested a binary, how did you install it? `npm install @asyncapi/cli -g` or?

AnishKacham:
> @AnishKacham when you tested a binary, how did you install it? `npm install @asyncapi/cli -g` or?

Yes. That is indeed how I installed asyncapi cli. the `./bin/run` was from cloning the repository and running `npm run build`

derberg:
getting the same with binary installed over npm:

```bash
$ asyncapi --version
@asyncapi/cli/0.52.3 darwin-x64 node-v16.13.0

$ asyncapi config context init
Initialized context /{...}/template-for-generator-templates-asyncapi-fork/.asyncapi-cli

$ cat .asyncapi-cli
{"store":{}}

$ asyncapi config context add dummy test/fixtures/dummy.yml
You have no context file configured. Run "asyncapi config context init" to initialize it.

Souvikns:
I think it is working for me
[...]

PR #752, opened 2023-08-09
feat: add flag to set context when it is created

### LINKED ISSUE #726: Add a new flag to set context when it is created
#### Reason/Context

To create a new context and then set it as current we have to run two commands namely 
- `asyncapi config context add`
- `asyncapi config context use`

We could really add a flag in the `context add` command so that we automatically set the config as current when we are creating it, so something like this 

```
asyncapi config context add [context-name] [file-path] --set-current
```

#### Description

I think this is preet simple as we just have to add a new flag in [here](https://github.com/asyncapi/cli/blob/dc99a3d376784babb710be2fd872f291bccdd3b5/src/commands/config/context/add.ts#L7), for more information on how to create flags in Oclif commands read [here](https://oclif.io/docs/flags)

https://github.com/asyncapi/cli/blob/dc99a3d376784babb710be2fd872f291bccdd3b5/src/commands/config/context/add.ts#L7

AnishKacham:
**Description**

- Adds a new flag `--set-current` / `-s` to the context add command that sets the context on addition
- Tested locally by building the cli
  
image

**Related issue(s)**
Resolves #726 
See also #751

### DIFF
--- a/src/commands/config/context/add.ts
+++ b/src/commands/config/context/add.ts
@@ -1,6 +1,6 @@
 import { Flags } from '@oclif/core';
 import Command from '../../../base';
-import { addContext } from '../../../models/Context';
+import { addContext, setCurrentContext } from '../../../models/Context';
 import {
   MissingContextFileError,
   ContextFileWrongFormatError,
@@ -10,6 +10,12 @@ export default class ContextAdd extends Command {
   static description = 'Add a context to the store';
   static flags = {
     help: Flags.help({ char: 'h' }),
+    'set-current': Flags.boolean({
+      char: 's',
+      description: 'Set context being added as the current context',
+      default: false,
+      required: false,
+    })
   };
 
   static args = [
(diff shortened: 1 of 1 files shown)


AnishKacham:
Hi, @Souvikns would you review this PR first or should we get #751 resolved first? 
The binary file from building the cli repository seems to work fine for this feature's purpose.

Souvikns:
> Hi, @Souvikns would you review this PR first or should we get #751 resolved first? The binary file from building the cli repository seems to work fine for this feature's purpose.


well, I don't think that the issue will be affected in any way by this PR, as it just adds a flag maybe we can merge it.

Souvikns (approved):
Looks good, docs are auto-generated but double-check if you have to update anything. When you are ready to merge just comment with `\rtm`

AnishKacham:
\rtm

Souvikns:
so sorry mistype, it will be `/rtm`

PR #756, opened 2023-08-11
feat: add flag 'd' for context add command to print any errors caught

AnishKacham:
**Description**

- add `--debug` / `-d` flag that prints any errors completely that are caught in the catch block.
``` 
BEFORE THE CHANGE
$ .\bin\run config context add foo .\asyncapi.yaml
You have no context file configured. Run "asyncapi config context init" to initialize it.

AFTER THE CHANGE
$ .\bin\run config context add foo .\asyncapi.yaml --debug
---DEBUGGING INFO---

 MissingContextFileError [ContextError]: These are your options to specify in the CLI what AsyncAPI file should be used:

    at C:\Anish\Code\cli\lib\models\Context.js:198:19
    at Generator.throw (<anonymous>)
    at rejected (C:\Anish\Code\cli\node_modules\tslib\tslib.js:165:69)

---END DEBUGGING INFO---

You have no context file configured. Run "asyncapi config context init" to initialize it.
 ```

**Related issue(s)**
See also #751

### DIFF
--- a/src/commands/config/context/add.ts
+++ b/src/commands/config/context/add.ts
@@ -10,6 +10,12 @@ export default class ContextAdd extends Command {
   static description = 'Add a context to the store';
   static flags = {
     help: Flags.help({ char: 'h' }),
+    debug: Flags.boolean({
+      char: 'd',
+      description: 'Print any errors that arise',
+      default: false,
+      required: false,
+    })
   };
 
   static args = [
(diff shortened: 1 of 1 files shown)


AnishKacham:
Closing this as the root cause for this PR has been fixed in the following PR https://github.com/asyncapi/cli/pull/778 that fixes Issue https://github.com/asyncapi/cli/issues/751

PR #765, opened 2023-08-18
Fix commands optimize logs when using parameter -o overwrite

rjambrecic:
**Description**

- ...
- ...
- ...

**Related issue(s)**

### DIFF
--- a/src/commands/optimize.ts
+++ b/src/commands/optimize.ts
@@ -113,7 +113,7 @@ export default class Optimize extends Command {
         this.log(`Created file ${newPath}...`);
       } else if (this.outputMethod === Outputs.OVERWRITE) {
         await writeFile(specPath ?? 'asyncapi.yaml', optimizedDocument, { encoding: 'utf8' });
-        this.log(`Created file ${newPath}...`);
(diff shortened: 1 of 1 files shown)


src/commands/optimize.ts
```diff
@@ -113,7 +113,7 @@ export default class Optimize extends Command {
         this.log(`Created file ${newPath}...`);
       } else if (this.outputMethod === Outputs.OVERWRITE) {
         await writeFile(specPath ?? 'asyncapi.yaml', optimizedDocument, { encoding: 'utf8' });
-        this.log(`Created file ${newPath}...`);
+        this.log(`Created file ${specPath ?? 'asyncapi.yaml'}...`);
```
derberg, line 116: @rjambrecic afaik we should also change the log message to say `Update file` instead of `Created file`?

also could you introduce a new variable so we do not duplicate `specPath ?? 'asyncapi.yaml'` that is already in line 115?

derberg (changes requested):
@rjambrecic thanks for the fix, I left one comment

cc @KhudaDad414

derberg:
no followup from contributor, closing

PR #778, opened 2023-08-31
fix: read context file sync way when calling context model

derberg:
Fixes https://github.com/asyncapi/cli/issues/751

- now `oclif.manifest.json` is also generated on local. Git ignored, but it will be published to npm as it is explicitly mentioned in package.json `files`
- we read context file using node sync method as the parent function cannot run inside `async` function


derberg:
@Souvikns @magicmatatjahu easy merge

Souvikns (approved):

PR #823, opened 2023-10-01
fix: switch from jest to mocha

Shurtu-gal:
**Description**

- Switch from jest to mocha for better accuracy of coverage as per [discussion](https://github.com/orgs/asyncapi/discussions/864)
- Current coverage :-
image

**Related issue(s)**
#599


Souvikns (approved):
Good job @Shurtu-gal 👍🏼.

PR #847, opened 2023-10-17
feat: add chocolatey support

Shurtu-gal:
**Description**

- Adds chocolatey support

**Related issue(s)**
#599


Shurtu-gal:
TODO :- 

- [ ] Get version from release itself

.github/workflows/release-chocolatey.yml
```diff
@@ -0,0 +1,42 @@
...
+          cd ./deploy/chocolatey
+          pwsh -File ./replace.ps1 -version ${{ steps.release_version.outputs.version }}
+        
+      
+      - name: Run Chocolatey Pack
+        run: |
+          cd ./deploy/chocolatey
+          choco pack ./asyncapi-cli.nuspec
+          choco apikey add --source "'https://push.chocolatey.org/'" --key ${{ secrets.CHOCOLATEY_API_KEY }}
+          choco push ./asyncapi.${{ steps.release_version.outputs.version }}.nupkg --source "'https://push.chocolatey.org/'"
```
Shurtu-gal, line 42: .

deploy/chocolatey/asyncapi-cli.nuspec
```diff
@@ -0,0 +1,29 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Do not remove this test for UTF-8: if “Ω” doesn’t appear as greek uppercase omega letter enclosed in quotation marks, you should use an editor that supports UTF-8, not this one. -->
+<package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
+  <metadata>
+    <id>asyncapi</id>
```
Shurtu-gal, line 5: .

Shurtu-gal (commented):
@derberg If you would be testing it somewhere please use id such as `asyncapi-test-123466` instead of `asyncapi`
At these places :-

.github/workflows/release-chocolatey.yml
```diff
@@ -0,0 +1,42 @@
+name: Release Chocolatey Package
+on:
+  release:
+    types: [published]
+  workflow_dispatch:
```
derberg, line 5: we should avoid manual releases, I guess it is here just for testing?

how about doing it the same as with brew -> that we release whatever is new once a day? since we have a lot of automation, might be that we have minor release 2-3 times a day, so we should avoid hitting chocolatey community, like we did with brew

.github/workflows/release-chocolatey.yml
```diff
@@ -0,0 +1,42 @@
...
+          }
+          else {
+            $version = "${{ github.event.release.tag_name }}" -replace "^v", ""
+          }
+          echo "Setting version to $version"
+          echo "::set-output name=version::$version"
+
+      - name: Make nuspec from the template
+        run: |
+          cd ./deploy/chocolatey
```
derberg, line 33: so these files that we have there, in this directory they stay as they are? and we commit them to repo?

if it is only deployment related, why not having it in `.github` folder? in the end, nobody should ever manually run deployment from local

.github/workflows/release-chocolatey.yml
```diff
@@ -0,0 +1,42 @@
...
+          pwsh -File ./replace.ps1 -version ${{ steps.release_version.outputs.version }}
+        
+      
+      - name: Run Chocolatey Pack
+        run: |
+          cd ./deploy/chocolatey
+          choco pack ./asyncapi-cli.nuspec
+          choco apikey add --source "'https://push.chocolatey.org/'" --key ${{ secrets.CHOCOLATEY_API_KEY }}
```
derberg, line 41: how did you get `CHOCOLATEY_API_KEY`, asyncapi needs an official account there?

.github/workflows/release-chocolatey.yml
derberg: please also add failure handling so we know if deploy to chocolatey fails -> https://github.com/asyncapi/cli/blob/master/.github/workflows/bump-homebrew-formula.yml#L32

deploy/chocolatey/tools/LICENSE.txt
```diff
@@ -0,0 +1,207 @@
+﻿
```
derberg, line 1: is it chocolatey requirement? why we need another license, why can't we push the one from root, what is the difference?

derberg (commented):
left some comments, please also update docs

deploy/chocolatey/tools/LICENSE.txt
Shurtu-gal: Yeah, this is the one packaged in tools hence needed.

.github/workflows/release-chocolatey.yml
Shurtu-gal: Yeah, asyncapi needs an official account there. Till now I was testing on my account would be great if somebody could make one and register the name **(i.e id)** we want to assign to cli app. ( Opinion :- Use asyncapi as that's what is used as the name in path too) 

@Souvikns Any take here.

.github/workflows/release-chocolatey.yml
Shurtu-gal: Sure.

.github/workflows/release-chocolatey.yml
Shurtu-gal: Yeah. I have removed it.

Shurtu-gal:
All required changes are done @derberg.

Shurtu-gal:
> I think you forgot to update docs?

Yeah, forgot to see it in the above-most one. Updated the installation docs just now.

derberg (approved):
all good from my side, I think we are ready to merge and test in production 😄 

@Souvikns anything?

Souvikns (approved):

Issue #841, opened 2023-10-10
[CLI] Measure AsyncAPI adoption


smoya:
Related to https://github.com/asyncapi/community/issues/879

## Scope

- [x] Define metrics to collect
- [ ] Collect metrics in code

smoya:
Having in mind that CLI command executions are stateless, and short-lived processes, how are we gonna keep the concept of a session so we can track consequent executions of commands (let's call it funnels).
Funnels can be tracked by using some correlation id, but that id should be stored somewhere, disk in particular.

The following questions should be answered soon or later in order to keep track of those sequence of executions:

1. What's the TTL (Time To Live) of each session? it is fixed? It is renewed on each interaction?
2. What do we use for generating the session ID? UUID? Something based on particular seed?.
3. Where should we store that session ID? `/tmp` dir? Shall we also mention that to the users?

derberg:
I'm more interested with vision on:
- is it on by default
- how we inform the user
- how easy it is to opt out
- and yes, we definitely need a detailed documentation what we do collecting data and what data we collect, with code pointers

regarding session? UUID would be definitely useful to identify "single" user. But why do we care about session. It is about AsyncAPI adoption right? what funnels have to do with this? For me this is different topic.

btw, we also need a way to say "this is CI/CD run"

smoya:
> ut why do we care about session. It is about AsyncAPI adoption right? what funnels have to do with this? For me this is different topic.

Different topic from what, exactly?
Session is needed in order to keep track of the following (which can be found in the description of https://github.com/asyncapi/community/issues/879):

> "From all the users doing asyncapi validate successfully, 40% run asyncapi generate next, 20% run asyncapi validate again, and the rest simply stop there.

smoya:
I'm gonna share what I think about the following:

> * What's the TTL (Time To Live) of each session? it is fixed? It is renewed on each interaction? Do we care about this but instead we use always the same UUID unless not found (and we create a new one) ?

It is a requirement to send something that acts as a correlation ID when sending those events, otherwise we won't be able to link them. New Relic, for example, has the Funnels feature. See https://docs.newrelic.com/docs/query-your-data/nrql-new-relic-query-language/nrql-query-tutorials/funnels-evaluate-data-series-related-events/

[...]

PR #859, opened 2023-10-23
feat: add POC for measuring adoption

smoya:
**Description**

> [!IMPORTANT]  
> Not ready yet, just marking as ready for review to let CI run tests

This PR is meant as a simple POC for demonstrating how we could collect metrics for measuring adoption. 
This code uses the following shared library https://github.com/smoya/asyncapi-adoption-metrics. It can be shared between CLI, Studio, and any other tool that wants to keep track of those metrics.

In this example, we are just measuring the happy path of `validate` CLI command. The sink (aka backend where metrics are sent) in this POC is just logging to StdOut (`console.log`).
The output when validating a simple and small doc atm will output the collected metrics, I.e.:

```
asyncapi-adoption.action.executed       COUNTER [
  _asyncapi_version: '2.6.0',
  _asyncapi_servers: 1,
  _asyncapi_channels: 1,
  _asyncapi_messages: 1,
  _asyncapi_operations_send: 1,
  _asyncapi_operations_receive: 0,
  _asyncapi_schemas: 1,
  success: true,
  validation_result: 'valid',
  action: 'validate'
]       1
```

Of course, a lot of things are missing surrounding this, like the important warning to the user saying we are collecting metrics, a way to disable, etc etc etc.
Also, we might want to encapsulate this logic as well in a single file or something we can reuse for other commands in CLI

cc @Amzani @peter-rr @fmvilas @derberg 

**Related issue(s)**
https://github.com/asyncapi/cli/issues/841


src/commands/validate.ts
```diff
@@ -29,6 +33,22 @@ export default class Validate extends Command {
...
+        metadata['success'] = true;
+        metadata['validation_result'] = result;
+        await this.recorder.recordActionExecution('validate', metadata);
+        await this.recorder.flush();
+      }
+    } catch (e: any) {
+      if (e instanceof Error) {
+        this.log(`Skipping submitting anonymous metrics due to the following error: ${e.name}: ${e.message}`);
+      }
+    }
```
fmvilas, line 52: Can't we just move this logic to `../parser`? Otherwise, we'd be parsing twice, right?

smoya: In this case (`validate` cmd), there is no call to `../parser.ts:parse()` method (which internally calls `parser.parse()`) but just `../parser.ts:parse()`, which it doesn't really parse the document into a known model but just validates the document by calling `parser.validate()`.
In short, we need to transform the doc into a known model (Parser-API v1, v2, etc) so we can later extract the data we need for the metrics.

There are other commands that, at some point, call the  `../parser.ts:parse()` methd, such as the `generate models` or `diff`. We could add a method there that calls the parser without logging errors, even though it's not a big improvement in terms of reusing code.
Those methods are logging validation errors to the stderr of the cmd, which is not a desirable behaviour here since we just want to log metrics in an invisible way.

I hope I answered your question/concern.

smoya:
Something we might want to move on is to use Oclif hooks so we can create the metric metadata on the `prerun` hook, and send the metric with a success: true in the `postrun` hook or rather with a `success: false` a the catch part.

In that way we would be always sending a metric, no matters if the command fails or not, and log the cmd result (OK|KO).

peter-rr:
@smoya Let me work on the metrics recording implementation for all the actions/commands as you did for `validate.ts`

src/commands/validate.ts
fmvilas: That makes sense. Thanks!

test/commands/generate/models/csharp/AnonymousSchema_1.cs
```diff
@@ -0,0 +1,29 @@
+namespace 'asyncapi.models'
```
smoya, line 1: @peter-rr with your latest commits (In particular [this](https://github.com/asyncapi/cli/pull/859/commits/ea19157a1aadc8a20df0867fb6b6df0855dd268e)) you brought a lot of changes that shouldn't belong to this branch and PR. I don't really get why those changes are here if, in theory, you did a merge from `master` branch to here. 

I suggest we schedule a meeting call so we can pair and fix this together and see what can we do next to try avoid this in the future. Does it sounds 👍 to you?

smoya: We had a call and finally got fixed.

smoya:
> [!IMPORTANT]  
> Not ready yet, just marking as ready for review to let CI run tests

peter-rr:
> [!IMPORTANT]
> Ready for review 👍 This message overrides the previous comment by @smoya

derberg (commented):
most important for me is to have in docs a new markdown document dedicated to this topic where:
- it is explained we collect by default
- what we collect
- where we collect
- how to disable - and why we ask not to disable
- is there a recommended environment where we want people to disable metrics - like CI/CD - because maybe they will pollute data. Or instead of docs and counting on people that they read them, we should stop metrics when `process.env.CI === true` ?

peter-rr:
@derberg Yup! That's already considered as one of the [next actions](https://github.com/asyncapi/cli/issues/841#issuecomment-1896210710) to be carried out 💪

peter-rr:
> to have in docs a new markdown document dedicated to this topic

Do you mean in a separate markdown document just related to metrics collection? @smoya and I were thinking on adding a new section at CLI's README file with all the info you mentioned regarding metrics.

> * where we collect

Should we mention we are making use of New Relic Metrics API in our case? Or not necessary to include this info?



> * we should stop metrics when `process.env.CI === true` ?
Isn't that case covered [here]
(https://github.com/smoya/cli/blob/d57655b0e27a1009f6123427680710f962c4a11d/src/base.ts#L87)? Or should we add another specific case or condition for CI?

fmvilas:
> Isn't that case covered [here](https://github.com/smoya/cli/blob/d57655b0e27a1009f6123427680710f962c4a11d/src/base.ts#L87)? Or should we add another specific case or condition for CI?

Nope. That's not to pollute metrics when we run `npm test`. We should also stop sending metrics if `process.env === 'CI'` because that would mean that a CLI command is being executed on a CI/CD pipeline and that's a tricky case when it comes to privacy and security that people for sure won't like. Plus it's not our target. We want to see how people use the CLI, not how scripts/machines use the CLI.

> Do you mean in a separate markdown document just related to metrics collection? @smoya and I were thinking on adding a new section at CLI's README file with all the info you mentioned regarding metrics.

I think we should be as transparent as possible here. If someone wants to find out what are we measuring and where are we putting this information they should be able to easily find it and disable it if needed. IMHO, this is too important to be a carry over task. It should be a must to merge this PR. Otherwise, we're risking that this task would never get done.

> Should we mention we are making use of New Relic Metrics API in our case? Or not necessary to include this info?

IMHO as much information as possible. We have nothing to hide so we shouldn't hide it.

peter-rr:
> I think we should be as transparent as possible here. If someone wants to find out what are we measuring and where are we putting this information they should be able to easily find it and disable it if needed. IMHO, this is too important to be a carry over task. It should be a must to merge this PR. Otherwise, we're risking that this task would never get done.
> 
> >

Cool 👍 Let me work on a separate document for that purpose then.

derberg:
@peter-rr I don't think standard is `process.env.NODE_ENV !== 'CI'`
I think it is env called `CI` (`process.env.CI === true`) that once set to `true` - metrics should not be collected. But yeah I googled https://adamj.eu/tech/2020/03/09/detect-if-your-tests-are-running-on-ci/ and I also found https://stackoverflow.com/questions/73973332/check-if-were-in-a-github-action-travis-ci-circle-ci-etc-testing-environme which might mean we need to support several different variables 🤷🏼 - you need to research basically, but I'm pretty sure it is not `NODE_ENV` set to `CI`

peter-rr:
> @peter-rr I don't think standard is `process.env.NODE_ENV !== 'CI'`
> I think it is env called `CI` (`process.env.CI === true`) that once set to `true` - metrics should not be collected. But yeah I googled https://adamj.eu/tech/2020/03/09/detect-if-your-tests-are-running-on-ci/ and I also found https://stackoverflow.com/questions/73973332/check-if-were-in-a-github-action-travis-ci-circle-ci-etc-testing-environme which might mean we need to support several different variables 🤷🏼 - you need to research basically, but I'm pretty sure it is not `NODE_ENV` set to `CI`

Indeed, after doing some research I've found the env name is `process.env.CI`. See:
-  https://dev.to/kapi1/solved-treating-warnings-as-errors-because-of-process-env-ci-true-bk5
- https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables
I've tested this change with `CI=true ./bin/run validate <spec-file-path>` and also `CI=false ./bin/run validate <spec-file-path>`, and everything is working as expected.

Change requested is ready for review now 👍 https://github.com/smoya/cli/pull/10

Amzani (changes requested):
> ›   To disable tracking, set the "ASYNCAPI_METRICS" env variable to "false" when executing the command. For instance:
 ›
 ›   ASYNCAPI_METRICS=false asyncapi validate spec_file.yaml

This could be improved for users that opt-out, appending `ASYNCAPI_METRICS=false` is not DX-friendly. 

Proposal: 
- Ask users to `export ASYNCAPI_METRICS="false"`
- Add a new command `asyncapi analytics off|on`

>  ›   Warning: AsyncAPI anonymously tracks command executions to improve the specification and tools, ensuring no sensitive data reaches our servers. It aids in comprehending how AsyncAPI tools are used
 ›   and adopted, facilitating ongoing improvements to our specifications and tools.

- This message is polluting every command execution, we could show it only once during installation/update or the first command execution.
- Remove `Warning` as this is informational to allow users to disable analytics. After this informational message, we start collecting metrics.

peter-rr:
> * Ask users to `export ASYNCAPI_METRICS="false"`

This is the easiest solution I think, and can be done easily by editing the informational message.
> * Add a new command `asyncapi analytics off|on`

This one would be more DX-friendly. I guess it could be considered as a subcommand of `asyncapi config`, something like `asyncapi config analytics off|on`


> * This message is polluting every command execution, we could show it only once during installation/update or the first command execution.

Agree 👍 We could also consider to show the message only at the beginning of every user session, but not sure if that option might be too invasive in case the user starts many different sessions during the same day.

> * Remove `Warning` as this is informational to allow users to disable analytics. After this informational message, we start collecting metrics.

Should we add some message title like `Information:` or just remove `Warning`?

Amzani:
> This one would be more DX-friendly. I guess it could be considered as a subcommand of asyncapi config, something like asyncapi config analytics off|on

Sounds good 👍.

> Agree 👍 We could also consider to show the message only at the beginning of every user session, but not sure if that option might be too invasive in case the user starts many different sessions during the same day.

It's still too invasive, why we can't just show it one time at any command execution and then store that we have shown this message in some file like `.asyncapi-analytics`, btw we can use the same file to store all related analytics information 

1. Analytics disabled/enabled (asyncapi config analytics off|on)
2. Analytic information shown (yes|no)
3. User ID or Correlation ID

As this PR is only focused on Adoption metrics, we can implement  1) and 2)

 > Should we add some message title like Information: or just remove Warning?

I'll just remove it for now, CLI UI will be improved later in https://github.com/asyncapi/cli/issues/872

peter-rr:
@Amzani 
At the moment I've changed the previous warning message to and informational message, and "How to disable tracking" section at markdown document has been updated in https://github.com/smoya/cli/pull/9

Working in the meanwhile on a fancier and more DX-friendly solution like `asyncapi config analytics off|on`. I agree with the idea about storing all the analytics info in the same file 👍

peter-rr:
@Amzani 
Command `asyncapi config analytics` and `.asyncapi-analytics` config file already implemented and ready for review: https://github.com/asyncapi/cli/pull/859/commits/e3fc3c79476703c9f78c52ee1c585510df0bfebd

Amzani (changes requested):

src/commands/config/analytics.ts
```diff
@@ -0,0 +1,33 @@
...
+  async run() {
+    try {
+      const { flags } = await this.parse(Analytics);
+      const isDisabled = flags.disable;
+      const analyticsConfigFile = join(process.cwd(), '.asyncapi-analytics');
+      const analyticsConfigFileContent = JSON.parse(readFileSync(resolve(process.cwd(), '.asyncapi-analytics'), 'utf-8'));
+  
+      if (isDisabled) {
+        analyticsConfigFileContent.analyticsEnabled = 'false';
+        await writeFile(analyticsConfigFile, JSON.stringify(analyticsConfigFileContent), {encoding: 'utf8'});
```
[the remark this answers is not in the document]
peter-rr: Done! 👍 I've decided to also cover the scenario where the user does not specify any flag.

src/commands/config/analytics.ts
```diff
@@ -0,0 +1,33 @@
...
+        analyticsConfigFileContent.analyticsEnabled = 'false';
+        await writeFile(analyticsConfigFile, JSON.stringify(analyticsConfigFileContent), {encoding: 'utf8'});      
+        this.log('Analytics disabled.');
+      } else {
+        analyticsConfigFileContent.analyticsEnabled = 'true';
+        await writeFile(analyticsConfigFile, JSON.stringify(analyticsConfigFileContent), {encoding: 'utf8'});
+        this.log('Analytics enabled.');
+      }
+    } catch (e: any) {
+      this.error(e as Error);
```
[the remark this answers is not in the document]
peter-rr: I've covered the `ENOENT` and `EEXIST` errors. Not sure about some other possible errors for this command 🤔

Amzani (changes requested):
Well done @peter-rr 
We need to add some tests under `test/integration`

peter-rr:
@Amzani Integration tests already added. Let me know if something else missing :)

peter-rr:
@Amzani Should we use safer alternatives, such as `SHA-256`, as recommended [by Sonarcloud](https://sonarcloud.io/project/security_hotspots?id=asyncapi_cli&pullRequest=859&resolved=false&inNewCodePeriod=true)? Or should we ignore it?

Amzani:
@peter-rr we have discussed it privately! let's go ahead with `SHA-256`

peter-rr:
Hey @Souvikns, could you take a look so we can get the final review on this PR? 💪 I'm working now on fixing the current conflict regarding the `package-lock.json`.

Souvikns (approved):
LGTM 👍🏼

peter-rr:
@Souvikns The CLI build process seems to be failing since the latest version `v1.6.1` has been released. Could you please have a look?  https://github.com/asyncapi/cli/actions/runs/8201865027/job/22431460372?pr=859

peter-rr:
@Souvikns Conflict already solved 💪

KhudaDad414 (commented):
LGTM

src/base.ts
```diff
@@ -82,12 +82,12 @@ export default abstract class extends Command {
...
     await super.finally(error);
     this.metricsMetadata['success'] = error === undefined;
-    this.metricsMetadata['source'] = this.generateSHA1();
+    this.metricsMetadata['source'] = await this.generateSHA1(this.specFile?.getSource() || '');
     await this.recordActionFinished(this.id as string, this.metricsMetadata, this.specFile?.text());
   }
   
-  generateSHA1(): any {
-    const fileBuffer = readFileSync(this.specFile?.getSource() || '');
+  async generateSHA1(filePath: string): Promise<any> {
```
smoya: I don't want to be intrusive but md5 is way faster than sha1 and for this I think it will be enough.

Additionally, are you sure you don't have access already to the content of the file? I'm more than sure you will have it in memory already, so no need to read from filesystem again (which is slow since I/O is required)

src/base.ts
```diff
@@ -11,4 +38,93 @@ export default abstract class extends Command {
...
     }
   }
+
+  async recordActionFinished(action: string, metadata: MetricMetadata = {}, rawDocument?: string) {
+    if (rawDocument !== undefined) {
+      try {
+        const {document} = await this.parser.parse(rawDocument);
+        if (document !== undefined) {
+          metadata = MetadataFromDocument(document, metadata);
+          metadata['source'] = this.generateSHA256(document.info().title());
```
smoya, line 48: Why sha256? Really this is so expensive!

I don't want to be intrusive but md5 is way faster than sha1, obviously faster than sha256 and for this I think it will be enough.

Also, using only the title is not enough. Tons of title will collide. I instead suggest you to use a combination of Title, description, version, etc.

Amzani: I thought that was some requirement from NewRelic, but I didn't read carefully that it was just a SonarCloud recommendation.

Even the `combination of Title, description, version, etc` will collide as the `new file` command generates the same file content.

For reference, this will be used in calculating [Time to first API Design](https://github.com/Amzani/asyncapi-metrics-dashboard/issues/2#issuecomment-1946098809)

```sql
SELECT filter(latest(endTimestamp),
                      WHERE action = 'generate:fromTemplate'
                      AND source is not null
                      AND success = true)/1000 - 
              filter(latest(endTimestamp),
                      WHERE action = 'new:file'
                      AND source is not null
                      AND success = true)/1000 AS 'TS'
FROM Metric FACET source, user
```

After revisiting this query, the source will change any way between file creation and generation. 

I recommend disabling collecting the source for now and solving this in a new PR.
cc @peter-rr , @smoya, @KhudaDad414

peter-rr: Agree 👍 Let me disable the source collection then and open a new issue about it.

peter-rr: Source collection already disabled, and new issue related to the topic has been opened: https://github.com/asyncapi/cli/issues/1225

Amzani:
@peter-rr anything blocking this PR from merging ?, we would like to get some adoption insights :)

peter-rr:
> @peter-rr anything blocking this PR from merging ?, we would like to get some adoption insights :)

Now I just need to update the dependency with [asyncapi-adoption-metrics](https://github.com/smoya/asyncapi-adoption-metrics) repo and also resolve the current conflicts. I'll let you know in case I find any blocker.

peter-rr:
@Amzani 
Conflicts already solved 👍 Just waiting for approval to be merged.

peter-rr:
Yo folks @derberg @magicmatatjahu ,
Could you have a look at this PR for a final review?

Souvikns (approved):
LGTM 👍🏼

Souvikns:
Having some issues while installing dependencies, I tried installing the dependencies locally and it works fine, I have no idea what is happening in the GitHub action. Looks like some cache error. 

```bash
npm ERR! code EEXIST
npm ERR! syscall open
npm ERR! path C:\npm\cache\_cacache\tmp\f46cd6fa
npm ERR! errno -4075
npm ERR! EEXIST: file already exists, open 'C:\npm\cache\_cacache\tmp\f46cd6fa'
npm ERR! File exists: C:\npm\cache\_cacache\tmp\f46cd6fa
npm ERR! Remove the existing file and try again, or run npm
npm ERR! with --force to overwrite files recklessly.

```

Amzani:
@Souvikns isn't it something related to this
https://github.com/asyncapi/cli/actions/caches
Screenshot 2024-03-26 at 21 47 57

Souvikns:
Looks like cleaning the cache by force, could potentially solve this, but in `gh action` we have to update the workflow file to add a new step to clean cache. @derberg can we make changes to the workflow file for this action run? What options do we have here?

PR #877, opened 2023-11-04
fix: make sure validation of invalid document exits `1`

### LINKED ISSUE #861: fail-severity warn info
#### Describe the bug
The fail-severity doesn't generate an exit code 1 even for the warn/info severity.  
@asyncapi/cli/0.58.12 linux-x64 node-v18.16.0

#### How to Reproduce
`asyncapi validate schema.yaml --fail-severity=warn`
image
`echo $?`
`0`

#### Expected behavior
The exit code should be `1` but got `0`.

shawakash:
**Description**

- Fixes the issue #861  having validation exit code error by checking the return status
- Added a enum for validation return status

**Related issue(s)**
Resolves #861

### DIFF
--- a/src/commands/validate.ts
+++ b/src/commands/validate.ts
@@ -30,7 +30,7 @@ export default class Validate extends Command {
     }
 
     const response = await validate(this, specFile, flags);
-    if (response == ValidationStatus.Invalid) {
+    if (response === ValidationStatus.Invalid) {
       this.exit(1);
     }
   }
--- a/src/commands/validate.ts
+++ b/src/commands/validate.ts
@@ -23,12 +23,16 @@ export default class Validate extends Command {
     const { args, flags } = await this.parse(Validate); //NOSONAR
     const filePath = args['spec-file'];
     const watchMode = flags.watch;
-
+    
     const specFile = await load(filePath);
     if (watchMode) {
       specWatcher({ spec: specFile, handler: this, handlerName: 'validate' });
     }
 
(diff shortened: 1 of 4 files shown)


shawakash:
@smoya  Please review it

MayraYecenia (approved):

derberg (approved):
lgtm

@Souvikns ?

src/commands/validate.ts
```diff
@@ -30,5 +30,9 @@ export default class Validate extends Command {
     }
 
     await validate(this, specFile, flags);
+    if(await validate(this, specFile, flags) == ValidationStatus.Invalid) {
```
derberg, line 33: actually why validate twice?

also `this.exit(0)` is not really necessary

shawakash: @derberg  changed it

Souvikns:
@shawakash there are some lint issues.

peter-rr:
@Souvikns Should we close this PR as the bug related is already solved by https://github.com/asyncapi/cli/pull/1068 ?

Souvikns:
We can close it, but this PR adds some refactoring, that we can use.

peter-rr:
> We can close it, but this PR adds some refactoring, that we can use.

@Souvikns 
Yeah, we can keep it as a refactor then. Should we change the title from `fix:` to `refactor:` in that case?

@shawakash 
Do you need some help to move forward with this PR?

PR #887, opened 2023-11-13
feat: add Glee tutorial template

fmvilas:
**Description**

It adds the `--template` (`-t`) option to the `asyncapi new glee` and `asyncapi new project` commands. It let's you choose which template to use to generate the Glee project.

I'm also providing a new template to use for the Code Generation tutorial we'll have on the website after releasing v3.

##### TODO
- [x] Upgrade `package.json` of each template so they use the latest Glee release.


Souvikns (approved):
Both `asyncapi new glee` and `asyncapi new project` create a glee project, still should we have both commands, doesn't it create confusion?

derberg:
`asyncapi new project` can definitely be removed

as if it comes to the requirements for this PR, the template from this PR is not implementing req/reply so why waiting for new glee features?

fmvilas:
> as if it comes to the requirements for this PR, the template from this PR is not implementing req/reply so why waiting for new glee features?

That's definitely not needed. I removed it. I was experiencing an error that I thought it was caused by the req/reply PR but was actually caused by https://github.com/asyncapi/glee/pull/589.

fmvilas:
Regarding `asyncapi new project`, I created this issue: https://github.com/asyncapi/cli/issues/913.

derberg (approved):

Souvikns (approved):

PR #888, opened 2023-11-14
fix: update examples owned by CLI to v3

derberg:
2 examples from CLI are owned by CLI, not fetched from `spec` repo - so updating them manually

the tutorial one is specifically important as used in https://v3.asyncapi.com/docs/tutorials/create-asyncapi-document


derberg:
@Souvikns marked it as `fix` to trigger release to get examples included in the package

Souvikns (approved):

PR #945, opened 2023-12-01
fix: generating fromTemplate with refs in spec

### LINKED ISSUE #944: CLI cannot generate fromTemplate when specification contains file path $refs
#### Describe the bug
When trying to generate with the html-template the process will fail when the specification file uses a `$ref` element that points to another file.

It will just show the following message and exit (The error message itself unfortunately is not really helpful):
```
Generator Error: Input is not a correct AsyncAPI document so it cannot be processed.
```

After adding a bit of logging it seems to me that it fails to resolve the path and cannot load the referenced file.

#### How to Reproduce

Running the following command will fail:
```shell
asyncapi generate fromTemplate ./index.yaml @asyncapi/html-template
```

when used with the following specification files:

```yaml
# index.yaml
asyncapi: 2.5.0
info:
  title: Example
  version: 0.1.0
channels:
  user/signedup:
    $ref: subfolder/user-signedup.yaml
```

```yaml
# subfolder/user-signedup.yaml
subscribe:
  message:
    description: An event describing that a user just signed up.
    payload:
      type: object
      additionalProperties: false
      properties:
        fullName:
          type: string
        email:
          type: string
          format: email
        age:
          type: integer
          minimum: 18
```

#### Expected behavior
The cli manages to generate output using the html-template and the given specification.

Running the generator directly instead of the cli does in fact produce the expected result:
```shell
ag index.yaml @asyncapi/html-template
```

Laupetin:
**Description**

Updates the code for generation via template to add the path of the specification file to the options passed to the generator.
This makes the generator able to resolve paths specified in `$ref` properties.

**Related issue(s)**
Fixes #944


derberg:
@Laupetin hey, thanks for the PR.

I just had a look at related issue and started digging a bit as it was confusing why it works in Generator CLI (next time please specify version that you use, it is super important) and do not work in AsyncAPI CLI.

Please have a look at https://github.com/asyncapi/parser-js/issues/797#issuecomment-1618966765
I have a feeling this PR workarounds this underlying problem

Laupetin:
@derberg Hi, thanks for checking it out. Sorry for the delayed answer.

I can reproduce this issue with `@asyncapi/cli/1.2.35` and it works with `@asyncapi/generator 1.16.0`.

From my understanding the root of this issue is that the CLI first loads the spec from the file to a string in-memory.
Then it calls the parser to load the spec _fromString_ (generateFromString).

The parser therefore does not know the directory in which the spec is located in.
It then fails to resolve relative paths.

What does work with the CLI is having an _absolute_ path inside the root spec file.
Taking the example from the linked issue and assuming the spec file is located in `/home/user/projects/sample/index.yaml`:

```yaml
# index.yaml
asyncapi: 2.5.0
info:
  title: Example
  version: 0.1.0
channels:
  user/signedup:
    $ref: /home/user/projects/sample/subfolder/user-signedup.yaml
```
^ This would work. It is not very practical however when having the spec in a git repository that could be checked out with different paths as well.

I could not manage to get anything working with the fix inside the linked comment. 
It is possible that i misunderstood though.
I tried `file:subfolder/user-signedup.yaml` and `file://subfolder/user-signedup.yaml`, both did not seem to work.

peter-rr (commented):
I've been trying to reproduce the bug related https://github.com/asyncapi/cli/issues/944 and seems to be solved (output is correctly generated) working under the current CLI version `1.14.1` 👍 

@Laupetin Could you confirm you are not experiencing the issue anymore? In that case we may proceed to close this PR and also the bug related. Thanks!

Laupetin:
@peter-rr hey, thanks for checking it out. You are right, the exact case described in the issue seems to now generate correct output.
When testing the new version on the documentation of our project i noticed however that this only applies if the `index.yaml` of the examplary setup is in your current workdirectory.

If the paths instead are `docs/index.yaml` and `docs/subfolder/user-signedup.yaml` then it will still fail to build (the resolved directory seems to be based on the workdirectory and not the directory containing the `index.yaml`).
Trying using the generator instead of the cli as described in the issue does seem to work though.

Laupetin:
I rebased the branch on the current master branch. The following scenario does not work with the master branch but does with the changes of this PR (the `index.yaml` file is not in the current working directory):

```yaml
# docs/index.yaml
asyncapi: 2.5.0
info:
  title: Example
  version: 0.1.0
channels:
  user/signedup:
    $ref: subfolder/user-signedup.yaml
```

```yaml
# docs/subfolder/user-signedup.yaml
subscribe:
  message:
    description: An event describing that a user just signed up.
    payload:
      type: object
      additionalProperties: false
      properties:
        fullName:
          type: string
        email:
          type: string
          format: email
        age:
          type: integer
          minimum: 18
```

peter-rr (approved):
LGTM 🚀 
Everything working fine with fix applied.

Shurtu-gal:
Great catch @Laupetin 🚀, and thanks @peter-rr for looking into this as well. Passing path makes sense acc. to me.

Shurtu-gal (approved):

derberg:
@peter-rr for now we had to disable sonar cloud - to many problems comparing to the value it brings.

hope we can get alternative solution you reported done quickly

/rtm

PR #1039, opened 2023-12-16
feat: added presets for JavaFileGenerator

### LINKED ISSUE #1038: JavaFileGenerator in the generate models command ONLY generates the java model classes correctly if the openapi specs follow the java naming conventions
#### Describe the bug
At the moment the `JavaFileGenerator` in the generate models command ONLY generates the java model classes correctly if the openapi specs follow the java naming conventions.

For example if a schema in the openapi spec contains a property `myID` the `generate models java` command will generate a property `myId` in the corresponding class.

We cannot assume that authors of asyncapi specs strictly follow the rules of java naming conventions.

We should assure that all openapi specs can be correctly generated into java model classes independent of following java naming conventions when writing the specs or not.

#### Description
In https://github.com/asyncapi/cli/blob/master/src/commands/generate/models.ts the `JavaFileGenerator` currently is instantiated this way:
`
fileGenerator = new JavaFileGenerator();
`

The `JavaFileGenerator` should be instantiated this way:
`
fileGenerator = new JavaGenerator({
  presets: [JAVA_JACKSON_PRESET]
});
`

This way the schema properties not following the java naming conventions would annotated with the corresponding `JsonProperty` annotations.

This won't be a breaking change other than one has to add the `jackson-annotations` dependency to one's project.

#### How to Reproduce
Take any asyncapi spec with schema properties that are not written in camel case and run the cli with `generate models java`.
Have a look at the model classes to find all non camel case schema properties converted to camel case java properties, thus breaking the contract.

#### Expected behavior
Since asyncapi does not enforce the usage of java compliant naming conventions for the schema properties in their specs it should correctly generate model classes for specs that don't follow the java naming conventions.

Athul0491:
**Description**
- This PR adds presets to the ```JavaFileGenerator``` in the generate models command. 
- This would ensure that the java model classes are generated correctly even if java naming conventions are not followed.

**Related issue(s)**
Resolves #1038


oliverkuntze:
Any chance that someone has the time to review this? We are desperately waiting for this fix. :-)

Souvikns (approved):
LGTM

PR #1043, opened 2023-12-19
feat: add support for modelina in v3

fmvilas:
Activate support for v3 in asyncapi generate models 😄


KhudaDad414:
@fmvilas never thought it could be this easy to support something... 😆

derberg (changes requested):
wait, did you just want to merge without approval?

what about tests adjustment, so we know v3 really works?

Souvikns (approved):

fmvilas:
Not trying to merge without approval @derberg. Just leaving it there for when you approve it merges automatically.

derberg:
but we should not remove unnecessary test but have an update to the test to make sure v3 works

> Not trying to merge without approval @derberg. Just leaving it there for when you approve it merges automatically.

but that is not how things work with ready to merge. Once you push any small change, it is no longer ready to merge and label is anyway removed

jonaslagoni (commented):
In order to enable AsyncAPI v3 for Modelina you need to switch to the next branch. 

This does contain breaking changes for the models, as well as not being fully released more might pop up.

jonaslagoni:
@fmvilas you can bump the version to v3 now, that supports AsyncAPI v3 and is stable ✌️

derberg:
closing in favour of https://github.com/asyncapi/cli/pull/1155

Issue #1081, opened 2024-01-22
Enable new feature from Generator to support templates hosted as private packages


derberg:
#### Reason/Context

We just released a new feature in Generator -> https://www.asyncapi.com/docs/tools/generator/using-private-template

#### Description

Would be great to enable new flags in `asyncapi generate fromTemplate` command that enable user to pass custom `registry.url` and specific authorization method, like `registry.auth` or `registry.token`

@Souvikns I think it is a nice, pretty simple new feature that would be good for onboarding potential new maintainer as it is typical enhancement to existing command - so maintainer candidate learns code from perspective of commands and their tests

Shiva953:
would like to take this up!

AayushSaini101:
@derberg Is it available to work, I can complete this

derberg:
yeah, I guess so, @Shiva953 in related PR do not respons 2 weeks already

AayushSaini101:
> yeah, I guess so, @Shiva953 in related PR do not respons 2 weeks already
Thanks @derberg

AayushSaini101:
@Amzani @derberg Started working on it creating a draft PR

PR #1091, opened 2024-01-25
feat: support private registry options in `asyncapi generate fromTemplate` command

Shiva953:
**Description**

- This pull request introduces new flags in the `asyncapi generate fromTemplate` command, in order to support [private registry options](https://www.asyncapi.com/docs/tools/generator/using-private-template) such that user can pass custom `registry.url` and specific authorization method, like `registry.auth` or `registry.token`

**Related issue(s)**
#1081


Shiva953:
cc @derberg @Souvikns

src/commands/generate/fromTemplate.ts
```diff
@@ -100,6 +101,15 @@ export default class Template extends Command {
...
       description: 'Maps all schema references from base url to local folder'
     }),
+    'registry-url': Flags.string({
+      description: 'Specifies the URL of the private registry for fetching templates and dependencies'
+    }),
+    'registry-auth': Flags.string({
+      description: 'The registry username and password encoded with base64, formatted as username:password'
+    }),
+    'registry-token': Flags.string({
+      description: 'The npm registry authentication token, that can be passed instead of base64 encoded username and password'
```
Min2who, line 111: i think it will make more sense if we take `registryUsername` and `registryPassword` as input from the user
```suggestion
  'registry-username': Flags.string({
      description: 'The username for the private registry.',
    }),
    'registry-password': Flags.string({
      description: 'The password for the private registry.',
    }),
```

src/commands/generate/fromTemplate.ts
```diff
@@ -122,6 +132,9 @@ export default class Template extends Command {
       noOverwriteGlobs: flags['no-overwrite'],
       mapBaseUrlToFolder: parsedFlags.mapBaseUrlToFolder,
       disabledHooks: parsedFlags.disableHooks,
+      registryUrl: parsedFlags.registryUrl,
+      registryAuth: flags['registry-auth'],
```
Min2who, line 136: registry auth should be change with base 64 value as mentioned in  [this](https://www.asyncapi.com/docs/tools/generator/using-private-template)
You can do this in something like this
```bash
const registryAuth = registryUsername && registryPassword
    ? Buffer.from(`${registryUsername}:${registryPassword}`).toString('base64')
    : undefined;
```

src/commands/generate/fromTemplate.ts
```diff
@@ -122,6 +132,9 @@ export default class Template extends Command {
       noOverwriteGlobs: flags['no-overwrite'],
       mapBaseUrlToFolder: parsedFlags.mapBaseUrlToFolder,
       disabledHooks: parsedFlags.disableHooks,
+      registryUrl: parsedFlags.registryUrl,
```
Min2who, line 135: the default URL of `registryUrl` should be `https://registry.npmjs.org`

Min2who (changes requested):

src/commands/generate/fromTemplate.ts
Shiva953: Hmm, that was also my initial concern I raised under the issue. Though in the example described [here](https://www.asyncapi.com/docs/tools/generator/using-private-template) we are directly providing the base64 encoded value, wouldn't it make sense to do the same here too?

Shiva953: any updates on this?

Shiva953:
any updates on this?

Souvikns:
@Shiva953 can you check, some tests are not passing

derberg (commented):
please take a look into generator and how new flag works. There is documentation available.

option on generator is called `registry` and it is an object of 3 values

url - auth - token

you need to adjust your code accordingly and only stuff that is available

so if username and password is not available - then you cannot use it 😄 

the way people should use generator in CI is always through token and auth that represents encoded username and password

it is all because of security

so you need to enable CLI user to pass all there different flags (which you already have - you just have to properly pass them to generator

Amzani:
@Shiva953 any updates ?

Amzani:
We can close this PR as it's already solved: https://github.com/asyncapi/cli/pull/1399
Thanks @Shiva953 @AayushSaini101

PR #1138, opened 2024-02-02
feat: removing v3 check for bundle command

### LINKED ISSUE #1137: CLI is unable to bundle spec V3 documents using bundle command
#### Describe the bug

`ayncapi bundle` command does not support bundling spec v3 documents, running the command on v3 documents gives error 

```
Error: One of the files you tried to bundle is AsyncAPI v3 format, the bundle command does not support it yet, please checkout https://github.com/asyncapi/bundler/issues/133
```

#### How to Reproduce
To reproduce this error all you have to do is try to bundle an v3 spec using CLI. 

Run the following command to reproduce this error 
```
asyncapi bundle file-name
```

#### Expected behavior
`asyncapi bundle` command should be able to bundle Asyncapi V3 spec files.

Souvikns:
**Description**
V3 check was put on all commands in CLI to prevent compiling v3 spec if the tool didn't support it yet. Since [bundler](https://github.com/asyncapi/bundler/) now supports spec V3 this PR is removing that check from the CLI. 

**Related issue(s)**

Fixes #1137


smoya (changes requested):

derberg (commented):
I think it is `feat` as it is new feature

but also think you need update tests to make sure they run against v3, as we care the most about the latest version. Of course if you think we can still have separate test for v2.

peter-rr (approved):
LGTM 👍

Amzani (approved):
LGTM

derberg (approved):

PR #1155, opened 2024-02-12
feat: enable v3 and scala for model generation

jonaslagoni:
**Description**
This PR updates the `models` command to support AsyncAPI v3 and adds scala support.

It also fixes a bug in the example document by using the Kafka example for v3 as well as using the same file name as for v2 document to stay consistent. 

**Related issue(s)**
Fixes https://github.com/asyncapi/modelina/issues/1376


derberg (commented):
actually I think we need to first revert https://github.com/asyncapi/cli/commit/b682a8e64627b4d281703ca57056c1b7052c5e62 and make another release as this is a breaking change and then this PR should make a bump and there should be a test added that will also validate the output of model generation

@Souvikns thoughts?

jonaslagoni:
@derberg it wont be possible to add a test that catches breaking changes in the model generation. The amount of variants you would need to test makes it impossible

jonaslagoni:
Probably a good idea yea

Souvikns:
Blocking auto-updates makes sense but now we have to manually update most of the libraries, It will prevent the breaking changes but we would miss the minor features.

derberg:
yeah, it sucks 😞 there is no best approach here 😞 loosing autoupdates sucks

or we complicate too much and nobody cares about output changes 🤔 nobody complains so far that latest modelina was bumped without breaking change

jonaslagoni:
Replaced by https://github.com/asyncapi/cli/pull/1376

Issue #1098, opened 2024-01-29
Old version of Studio wrapped in CLI


eelcofolkertsma:
### Describe the bug
Some asyncapi/cli commands launch studio. The version launched does not handle v3.0.0 documents yet 

### How to Reproduce

- Install asyncapi/cli in your environment per instructions
- run command "asyncapi new" for some file and confirm to open in studio
- browser opens and displays invalid document with diagnostic "Version "3.0.0" is not supported. Please use "2.6.0" (latest) version of the specification. 

#### Screenshots
![image](https://github.com/asyncapi/cli/assets/71253997/42e2e9bd-11f6-4479-b2ee-1c6765e8b82c)
![image](https://github.com/asyncapi/cli/assets/71253997/78234e23-ddd3-426e-beed-cb6bcb5e95a9)

#### Link to GitHub repository with project that has issues
https://github.com/asyncapi/cli, master branch

### Expected behavior
A happy studio page showing Account Service asyncapi: 3.0.0 with zero validation error

PR #1182, opened 2024-02-19
fix: Studio cache issues

Amzani:
Fixes https://github.com/asyncapi/cli/issues/1098

Note: The browser might cache old studio version that doesn't support 3.0.0


src/models/Studio.ts
```diff
@@ -88,16 +89,16 @@ export function start(filePath: string, port: number = DEFAULT_PORT): void {
...
   });
-  
+
   wsServer.on('close', (socket: any) => {
     sockets.splice(sockets.findIndex(s => s === socket));
   });
 
   server.listen(port, () => {
-    const url = `http://localhost:${port}?liveServer=${port}`;
+    const url = `http://localhost:${port}?liveServer=${port}&_sv=${studioVersion}`;
```
derberg, line 101: are you sure this is a solution? In linked issue nobody specified that `1.4.0` of CLI was used, and older version did not have latest studio


also where is this `_sv` param coming from?

Amzani: Yes, it's not the most exciting solution, but we should find a way to completely disable the browser cache in the local HTTP server we are using to start Studio.

The issue happens for new CLI versions, even though we have the latest studio version. 

`_sv` -> studio version.

Amzani: https://www.loom.com/share/0a41904d9537411f98b2b46bc25f0344?sid=acb7be86-6da2-47f1-a568-5f328fe9221c

derberg: oh crap 🤔 

ok, so `_sv` has no meaning for studio really, it is just a hack to force "refresh". 


I think it is good, also useful as it informs what studio version you are using.

This is why, what about using full name for the query param? so it is immediately clear what that is?

Amzani: @derberg done, you take a look?

PR #1273, opened 2024-03-14
ci: skip trying to push an already pushed version

Shurtu-gal:
**Description**

Check if the version to be pushed has already been pushed to the remote.

**Related issue(s)**
https://asyncapi.slack.com/archives/CQVJXFNQL/p1710432214064749?thread_ts=1710238071.506779&cid=CQVJXFNQL

**Examples**
- [If a package is not pushed it will build normally](https://github.com/ash17290/cli/actions/runs/8287047957/job/22678466474)
- [Else will skip everything](https://github.com/ash17290/cli/actions/runs/8287051904/job/22678478986)


derberg (approved):
Thanks

PR #1304, opened 2024-03-27
feat: add asyncapi file birth timestamp in metrics collection.

### LINKED ISSUE #1161: [CLI] Measure AsyncAPI DX metrics
Related to https://github.com/Amzani/asyncapi-metrics-dashboard/issues/5

After merging https://github.com/asyncapi/cli/pull/859, we need to track unique AsyncAPI Files and users, see https://github.com/Amzani/asyncapi-metrics-dashboard/issues/2

## Definition of done

- We expected to use the `.asyncapi-analytics` file to store user IDs (UUID v4 is perfect)
- We use SHA1 (FileName) for identifying AsyncAPI File Names
- Analytics documentation is updated to inform users about what we collect
- We don't impact CLI performance

Example of hardcoded values I've used for testing purpose in `base.ts`

```
async finally(error: Error | undefined): Promise<any> {
    await super.finally(error);
    this.metricsMetadata['success'] = error === undefined;
    this.metricsMetadata['user'] = 'c320a14b-37f7-4a44-bdf0-9bf3ec22b652';
    this.metricsMetadata['source'] = this.generateSHA1(this.specFile?.getSource() || '');
    console.log('getFileURL', this.specFile?.getSource(), 'hash', this.generateSHA1(this.specFile?.getSource() || ''));
    await this.recordActionFinished(this.id as string, this.metricsMetadata, this.specFile?.text());
  }

```

### LINKED ISSUE #1225: [CLI] Add 'source' metadata for metrics collected

We need to include `source` metadata as a unique identifier for every file handled in CLI when collecting metrics for the commands executed.

This change will allow us to measure the `Time to first API Design` metric as described [here](https://github.com/Amzani/asyncapi-metrics-dashboard/issues/2#issuecomment-1946098809).

We need to consider two different issues we've found so far:

- The **hashing algorithm** to use: We should make a decision based on computational cost and security aspects according to New Relic requirements and also the recommendations from SonarCloud. Some research about algorithms like `md5`, `sha1` or `sha256` is needed.
- Find another way to calculate the `Time to first API Design` metric, or a different way to identify the source so it doesn't change between file creation and generation. See https://github.com/asyncapi/cli/pull/859#discussion_r1520842866

### Related issue(s)
Relates to https://github.com/asyncapi/cli/issues/1161

### LINKED ISSUE #1300: Add Time to First API Design DX metric
## Why do we need this improvement?

After merging https://github.com/asyncapi/cli/pull/859 
[DX Working group](https://github.com/orgs/asyncapi/discussions/1099) would like to track [Time to First API Design](https://github.com/Amzani/asyncapi-metrics-dashboard/issues/2#issuecomment-1946098809)

## How will this change help?

Improve our tooling

## Screenshots

_No response_

## How could it be implemented/designed?

Check https://github.com/Amzani/asyncapi-metrics-dashboard/issues/2#issuecomment-1946098809
As the AsyncAPI filename might change between the creation (`asyncapi new file`) and generation (code, doc), we could try this solution described [here](https://stackoverflow.com/questions/43170582/is-there-any-way-to-retrieve-a-file-folders-unique-id-via-nodejs) - If this is not working then we could just use the filename as we don't need to deal with edge cases.

## Scope

- [x] https://github.com/Amzani/asyncapi-metrics-dashboard/issues/9
- [ ] https://github.com/Amzani/asyncapi-metrics-dashboard/issues/10

## 🚧 Breaking changes

No

## 👀 Have you checked for similar open issues?

## 🏢 Have you read the Contributing Guidelines?

## Are you willing to work on this issue?

No, someone else can work on it.

KhudaDad414:
**Description**

- collect the asyncapi file birth timestamp.
- It will help us track each file for each user over time.

fixes #1225, fixes #1161, fixes partialy #1300 

EDIT:
fixes #1300

cc: @Amzani @smoya @peter-rr


src/models/SpecificationFile.ts
```diff
@@ -136,6 +137,7 @@ export async function load(filePathOrContextName?: string, loadType?: LoadType):
       return Specification.fromURL(filePathOrContextName);
     }
     await fileExists(filePathOrContextName);
+    if (command) {command.specFilePath = filePathOrContextName;}
```
fmvilas, line 140: 🤔 I have a bad feeling about injecting information on a foreign class like this. It doesn't feel elegant or "expected". Look at all the calls above like `await load(this, specFilePath)`. Or the signature of this method. It feels weird that we have to pass `this` or a `command` to load a file.

For instance, [this line](https://github.com/asyncapi/cli/pull/1304/files#diff-277d6f3a036a06f971ae75da2804a5b6b60cab44c03eb53d8d11a13ceef6e644R26) could become something like:

```js
this.specFile = await load(filePath);
this.specFilePath = this.specFile.getPath();
```

fmvilas (changes requested):

src/models/SpecificationFile.ts
KhudaDad414: Made the changes. 🙇

KhudaDad414:
[reason for tests not passing](https://github.com/asyncapi/.github/pull/280/files#r1548640815).

peter-rr (approved):
LGTM 👍

src/base.ts
```diff
@@ -69,6 +70,7 @@ export default abstract class extends Command {
 
   async recordActionMetric(recordFunc: (recorder: Recorder) => Promise<void>) {
     try {
+      this.setSource();
```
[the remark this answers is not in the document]
KhudaDad414: @peter-rr `recordActionMetric()` is called for both `asyncapi_adoption.action.invoked` and `asyncapi_adoption.action.finished`, so if there is a file, there will be a timestamp.

till now I don't see any benefit for collecting the file info at `asyncapi_adoption.action.invoked` since it will be the same as `asyncapi_adoption.action.finished` everytime.

in some commands like `asyncapi new file` it is impossible to collect this metric at `asyncapi_adoption.action.invoked` since no file has been created at that stage.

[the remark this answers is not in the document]
peter-rr: > till now I don't see any benefit for collecting the file info at `asyncapi_adoption.action.invoked` since it will be the same as `asyncapi_adoption.action.finished` everytime.

Yeah, makes sense 👍 

> in some commands like `asyncapi new file` it is impossible to collect this metric at `asyncapi_adoption.action.invoked` since no file has been created at that stage.

Agree! However, I don't understand why the metric is not collected for commands like `asyncapi validate` or `asyncapi convert`, in which the file already exists. Might be something related to the file processing time and `asyncapi_adoption.action.invoked` is sent before the existing file is processed? 🤔

src/base.ts
```diff
@@ -3,7 +3,7 @@ import { MetadataFromDocument, MetricMetadata, NewRelicSink, Recorder, Sink, Std
 import { Parser } from '@asyncapi/parser';
 import { Specification } from 'models/SpecificationFile';
 import { join, resolve } from 'path';
-import { existsSync } from 'fs-extra';
+import { existsSync, statSync } from 'fs-extra';
```
fmvilas, line 6: Why not using the asynchronous versions of these functions? 🤔 You can easily make `setSource` an async function and then it would be easy to call `await stat()` instead.

fmvilas (changes requested):

src/base.ts
peter-rr: Working fine with: `
const stats = await fPromises.stat(specFilePath);
`
Don't know if there is any better alternative to make `setSource` an async function.

KhudaDad414: @peter-rr I don't think there is any. we either have to use promises and `await` it. or use Sync functions. which I guess the first option is a lot better.

peter-rr (approved):
LGTM 👍

src/base.ts
peter-rr: LGTM 👍

Souvikns (dismissed):

Souvikns:
@fmvilas need your approval to merge this.

src/base.ts
```diff
@@ -78,6 +79,16 @@ export default abstract class extends Command {
...
   }
 
+  async setSource() {
+    const specFilePath = this.specFile?.getFilePath();
+    if (!specFilePath) { return; }
+    try {
+      const stats = await stat(specFilePath);
+      this.metricsMetadata['file_creation_timestamp'] = stats.birthtimeMs;
+    } catch (e: any) {
+      // do nothing.
```
fmvilas, line 89: What are the possible errors we could find here? File doesn't exist maybe? We don't have permission to access the file?

I have the feeling we should control (and report to New Relic) failures here.

fmvilas (changes requested):

src/base.ts
KhudaDad414: Since this function is ultimately being [called from](https://github.com/KhudaDad414/cli/blob/aca055c076769d36978a81bc5714adccfa2c6a36/src/base.ts#L95-L96) the `finally` method of the `command` with a file path, If there is any error with the file, We expect that the current call is actually reporting the error. 

clarified the comment in the code.

peter-rr: Just one question on this topic: if there is an error with the file, is it handled at the `catch` [method](https://github.com/KhudaDad414/cli/blob/aca055c076769d36978a81bc5714adccfa2c6a36/src/base.ts#L30-L39) as well, apart from the `finally` method? 🤔

KhudaDad414: @peter-rr if by "handled" you mean reporting to New Relic, then it is only being done in `finally` since it will run regardless of errors.

Amzani: @KhudaDad414 @peter-rr It's generally considered best practice to handle errors as close to their source as possible. This helps maintain clear code structure and makes it easier to understand and reason about error handling logic, in this case having some exception type checks under the `catch` would help.

Amzani:
@KhudaDad414 can you elaborate more on how this will partially solve https://github.com/asyncapi/cli/issues/1300 thanks

KhudaDad414:
@Amzani the `Track Time to First API design in the CLI` scope of the issue will be resolved. we already have the `Add Time to First API design to Metric dashboard` in [the dashboard](https://asyncapi-metrics-dashboard.vercel.app/dashboard/DeveloperExperience/1W).

src/base.ts
KhudaDad414: @Amzani this is how I understand the CLI would work when an error happens:

1) A CLI command is run against an invalid file path, or a file that the user doesn't have read permission etc...
2) Error is thrown.
2) `catch` method will log the error.
3) `finally` method will report it to New Relic.

`setSouce` is being called from the `finally` method. so the same error that caused the `finally` method to run will occur again. If the error is already logged and is being reported to New Relic, I am failing to understand why it should be handled and not be ignored here.

peter-rr: I would need some clarification here: are we all talking about the `catch` [method](https://github.com/asyncapi/cli/pull/1304/files/aca055c076769d36978a81bc5714adccfa2c6a36#diff-440b55a9ff9e0486020fc49197f3278d14446306fc0c5d5107b48139c20bfda6R30-R39) from the class, or the `catch` [block](https://github.com/asyncapi/cli/pull/1304/files/aca055c076769d36978a81bc5714adccfa2c6a36#diff-440b55a9ff9e0486020fc49197f3278d14446306fc0c5d5107b48139c20bfda6R88-R91) inside the `setSource` method?

KhudaDad414: I am saying that if there is an error with the file, we expect the command to throw an error, `catch` method inside the class will log the error, and `finally` method will report it.

Since The `setSource` is called from the `finally` method, it is unnecessary to do anything inside of it's `catch` block. 

If we log it, we have logged it twice, if we initiate a report we have reported it twice.

One time by command trying to access the file and throwing the error and one time by us trying to access it here.


Hope it is clear now.

Amzani (approved):
LGTM

KhudaDad414:
@fmvilas waiting for your review. 🙇‍♂️

Amzani (approved):
LGTM

PR #1338, opened 2024-04-03
fix: update @smoya/multi-parser and @asyncapi/parser to include parser v3.0.12

smoya:


smoya:
@Souvikns @derberg @magicmatatjahu this PR is fixing your recent errors in CI failing with:

> Error: src/base.ts(46,43): error TS2345: Argument of type 'AsyncAPIDocumentInterface' is not assignable to parameter of type 'AsyncAPIDocument'.

Example: https://github.com/asyncapi/cli/actions/runs/8543665709/job/23408013867?pr=1324

The issue is that multiparser uses parser, and CLI has both parser and multiparser as dependency (which could be removed in favor of multiparser only). So there was a version missmatch between those parsers.

derberg (approved):

smoya:
error-windows-gif
Shall we just skip that and merge this by using Maintainer's force?

derberg:
did some fixes in CI

updating the branch and lets 🤞🏼

PR #1339, opened 2024-04-04
ci: add install step in testing that in the end skips cache

derberg:
doing it first here to make sure it works, then will rollout to others


derberg:
have to force merge, as it is also failing CI like others

this PR is fixing CI - I hope

PR #1340, opened 2024-04-04
ci: use `npm ci` and force to use npm 8 with node 18 on windows

derberg:
we need to use npm 8 when using node 18 on windows due to some cache issues

reported bug https://github.com/actions/runner-images/issues/9630

even updates to package-lock did nothing


.github/workflows/if-nodejs-pr-testing.yml
```diff
@@ -56,23 +57,10 @@ jobs:
         uses: actions/setup-node@v4
         with:
           node-version: "${{ steps.lockversion.outputs.version }}"
-          cache: 'npm'
-          cache-dependency-path: '**/package-lock.json'
       - if: steps.packagejson.outputs.exists == 'true'
         name: Install dependencies
-        id: first-installation
+        shell: bash
         run: npm ci
```
Souvikns, line 63: should we just `--force` this?

derberg:
@Souvikns yeah, forcing install is scary, especially across package-lock versions

Souvikns (approved):

Souvikns (approved):

Issue #1371, opened 2024-04-17
[FEATURE] Automatic Version Update Notification


KhudaDad414:
### Why do we need this improvement?

Right now, if you're using the CLI, there's no prompt telling you when a new version is available. This means you've got to keep an eye out yourself on GitHub or npm to make sure you're not missing out. 

Having to do this manually is a bit of a hassle and might lead to missing out on cool new features or important fixes.

### How will this change help?

Adding a little notification about new versions when you use the CLI could make a big difference. Users can get new features and bug fixes sooner.

### How could it be implemented/designed?

A few things to consider when implementing this:

- It should be automatic.
- Alert if a new version is available: If there’s a new version, the CLI should show a friendly message. Something like, "A new major/minor version is available. Update by running `npm update/upgrade -g @asycapi/cli`."
- The user should be able to disable it entirely.
- Frequency: Nobody likes a nag. We should make sure it only checks for updates/shows the message every so often (like once a week), so it's not annoying.

### 🚧 Breaking changes

No

### 👀 Have you checked for similar open issues?

- [X] I checked and didn't find a similar issue

### 🏢 Have you read the Contributing Guidelines?

- [X] I have read the [Contributing Guidelines](https://github.com/asyncapi/.github/blob/master/CONTRIBUTING.md)

### Are you willing to work on this issue?

No, someone else can work on it.

AayushSaini101:
@KhudaDad414 I want to work on this. I have some questions before: 
* Where we will get the notification message ? when the user install cli first time on their machine ? or when the user execute particular command using the cli ?

KhudaDad414:
When the user runs a command.

Just like the 'npm' message when a new version is available.

KhudaDad414:
When the user runs a command.

Just like the 'npm' message when a new version is available.

PR #1393, opened 2024-04-27
feat: add log to show the version status of cli

AayushSaini101:
Resolves: #1371


Shurtu-gal:
@AayushSaini101 I suppose it the message should be infrequent so as not to bother users.

AayushSaini101:
> @AayushSaini101 I suppose it the message should be infrequent so as not to bother users.

@Shurtu-gal Thanks, used the probaility method to show the message when the probaility is less than 0.5. 

Sample Testing of message: 

Screenshot 2024-04-27 at 10 05 09 AM

src/utils/versionVerification.ts
```diff
@@ -0,0 +1,27 @@
...
+  const packageJson = JSON.parse(await fs.readFile('package.json', 'utf-8') as string);
+  const currentVersion = `v${packageJson.version}`;
+  // Compare the versions
+  if (latestVersion !== currentVersion) {
+    // Avoid sending message frequently
+    const randomProbability = Math.random();
+    if (randomProbability < 0.5) {return `There is a newer version available: ${latestVersion}. You are currently using ${currentVersion}.`;}
+  }
+}
+checkLatestVersion();
```
Shurtu-gal, line 27: Is there any reason the function is being called here?

AayushSaini101: thanks @Shurtu-gal no used, I was testing some changes removed

AayushSaini101: @Shurtu-gal If we are doing infrequent messages, how can we test this 🤔

Amzani:
> @AayushSaini101 I suppose it the message should be infrequent so as not to bother users. 
Agree, especially if we don't have (yet) an auto-update feature.

Amzani:
@AayushSaini101 why not just use https://github.com/oclif/plugin-warn-if-update-available, instead of having the random check for simplicity the plugin provides a periodic check that is more common. 

We can check every 24h for instance.

Some other useful plugins we can add:
- https://github.com/oclif/plugin-version
- https://github.com/oclif/plugin-update

AayushSaini101:
> @AayushSaini101 why not just use https://github.com/oclif/plugin-warn-if-update-available, instead of having the random check for simplicity the plugin provides a periodic check that is more common.
> 
> We can check every 24h for instance.
> 
> Some other useful plugins we can add:
> 
> * https://github.com/oclif/plugin-version
> * https://github.com/oclif/plugin-update
Thanks @Amzani I am exploring them

KhudaDad414 (commented):
regarding the message that we show, I am seeing the following format in a few places including `Next.js`, `Storybook`, `Pnpm`. 

Screenshot 2024-04-29 at 11 53 10

AayushSaini101:
> https://github.com/oclif/plugin-warn-if-update-available

@Amzani @KhudaDad414 Added plugin and configuration to run in 24 hour.

Shurtu-gal:
Great work @AayushSaini101.

KhudaDad414 (approved):
LGTM

Shurtu-gal (approved):
LGTM 🚀

KhudaDad414:
@AayushSaini101  the tests seem to be cancelled. can you push an empty commit to re-run the tests?

peter-rr (commented):
LGTM 🚀

PR #1412, opened 2024-05-08
fix: enable to use nodejs and java templates with asyncapi v3

derberg:
with current code we get https://github.com/asyncapi/nodejs-template/issues/215#issuecomment-2098550341 error

nodejs template already supports v3: https://github.com/asyncapi/nodejs-template/releases/tag/v3.0.0
joava template as well: https://github.com/asyncapi/java-template/releases/tag/v0.3.0


Shurtu-gal (approved):
Thanks for the PR!!

Issue #1131, opened 2024-01-30
Ability to score quality of async api specification file


Ferror:
#### Reason/Context

Many optional yet valuable parameters in the Async API specification can improve the developer experience. Like `example`, `description`, `format`, etc. When working with many specifications, I was asked, "**How good is our documentation?**". How do I measure if my documentation is good, mediocre, insufficient, or 1-100?

I propose a new command like `score` or `evaluate` that will reuse existing warnings and errors from the `validate` command. The command will extend the list of best practices and generate some final/aggregated scores. Hints that will increase the score should be provided.

##### A couple of examples

1. if the specification has **servers**, then it increases the score
2. if the specification has **descriptions**, then it increases the score
3. if the specification has **examples**, then it increases the score
4. if the specification has a **format**, then it increases the score
5. if the specification has **channels**, then it increases the score
6. if the specification has **operations**, then it increases the score
7. if the specification has a **license**, then it increases the score

AayushSaini101:
@Amzani I want to work on this issue:) I have worked similar kind of issue for Jekins https://github.com/jenkins-infra/plugin-health-scoring that provide scoring for each pluing on basis of some parameters

Amzani:
Sure @AayushSaini101 please go ahead.
That's a nice feature to have as well in studio

AayushSaini101:
> Sure @AayushSaini101 please go ahead. That's a nice feature to have as well in studio

@Amzani @derberg @Ferror We need to decide how much score we need to increase for particular parameter. Which parameter will be more proritise as compared to others, I can created a small proposal for that

Ferror:
I think we could start with something simple in the form of MVP.

1. The abstraction for scorecards/metrics (What to score)
2. Each metric may have a scale/weight
3. Deciding on an algorithm to calculate the sum/mediana/average

Feel free to propose something. We could also start with linear calculation (without weights) 😄

AayushSaini101:
> I think we could start with something simple in the form of MVP.
> 
> 1. The abstraction for scorecards/metrics (What to score)
> 2. Each metric may have a scale/weight
> 3. Deciding on an algorithm to calculate the sum/mediana/average
> 
> Feel free to propose something. We could also start with linear calculation (without weights) 😄


[...]

Issue #1214, opened 2024-03-07
CLI new UI/UX implementation


Amzani:
## Problem

After https://github.com/asyncapi/cli/issues/872 completion, we now have a [visual direction](https://www.figma.com/file/9yt0aJh7cVoJHxXzDDb2lt/CLI-UI%2FUX-Project?type=design&node-id=195%3A762&mode=design&t=XeFNVy8nWiOjPVeC-1) of what needs to be implemented in the CLI.

## Scope

- [x] https://github.com/asyncapi/cli/issues/1215
- [x] #1379
- [x] #1381 
- [x] #1382 
- [x] #1385

Shurtu-gal:
Great @Amzani, we can make some decisions in parallel too.

There was a great proposal by @fmvilas on new framework, [clack](https://github.com/natemoo-re/clack)

Amzani:
Great @Shurtu-gal,
I don't know clack, does it replace oclif? 
In case we need to change Oclif, this might help https://github.com/asyncapi/server-api/issues/485#issuecomment-1961317382

Shurtu-gal:
@Amzani I believe we can take it as a reference, and build our own design system to not face problems down the line.

aeworxet:
### Bounty Issue's service comment

Text labels: `bounty/2024-Q2`, `bounty/advanced`, `bounty/coding`
First assignment to third-party contributors: [2024-03-22 00:00:00 UTC+12:00](https://time.lol/#2024-03-22T00:00:00+12:00)
End Of Life: [2024-08-31 23:59:59 UTC-12:00](https://time.lol/#2024-08-31T23:59:59-12:00)

@asyncapi/bounty_team

utnim2:
would like to work on this

Amzani:
@utnim2 @Shurtu-gal already started to work on the foundations, I think once he finish you can be part of it, as we need to implement DX in remaining commands.

aeworxet:
@Amzani Please assign @Shurtu-gal and do not disperse the assignment, as unlike usual GitHub issues, the Bounty Issue can be worked on by only one person.

Shurtu-gal:
@Amzani I was looking into this, and I am faced with some doubts:-
The current output looks something like:- 
![image](https://github.com/asyncapi/cli/assets/100484401/2933b659-5d02-4b98-a62a-49406fa1eff1)

[...]

PR #1200, opened 2024-02-23
feat: new CLI architecture (part 1)

Amzani:
Resolves https://github.com/asyncapi/server-api/issues/485
Resolves https://github.com/asyncapi/cli/issues/781

This PR introduces changes discussed in https://github.com/asyncapi/server-api/issues/485#issuecomment-1961317382 and https://github.com/asyncapi/cli/issues/781#issuecomment-1990571897

### New architecture/flags extractions

```
.
├── adapters
│   ├── api
│   │   ├── controllers
│   │   │   ├── generator.controller.ts
│   │   │   └── ping.controller.ts
│   │   └── core
│   │       ├── app.ts
│   │       ├── constants.ts
│   │       ├── exceptions
│   │       ├── interfaces.ts
│   │       ├── logs
│   │       ├── middlewares
│   │       └── server-api.d.ts
│   └── cli
│       ├── commands
│       │   ├── bundle.ts
│       │   ├── config
│       │   ├── convert.ts
│       │   ├── diff.ts
│       │   ├── generate
│       │   ├── new
│       │   ├── optimize.ts
│       │   ├── start
│       │   └── validate.ts
│       └── core
│           ├── base.ts
│           ├── flags
│           ├── global.d.ts
│           ├── globals.ts
│           ├── hooks
│           ├── models
│           └── parser.ts
├── api.ts
├── core
│   └── services
│       ├── archiver.service.ts
│       └── generator.service.ts
├── errors
│   ├── context-error.ts
│   ├── diff-error.ts
│   ├── generator-error.ts
│   ├── specification-file.ts
│   └── validation-error.ts
├── index.ts
├── logs
├── ports
│   └── generator.interface.ts
└── utils
    ├── ajv.ts
    ├── generator.ts
    ├── logger.ts
    └── temp-dir.ts
```

This architecture will enable CODE Maintainers split discussed in https://github.com/asyncapi/cli/issues/781

## To test the API

```
./bin/run start api -p 3001
curl --location 'http://localhost:3001/v1/generate' \
--data-raw '{
[...]

PR #1413, opened 2024-05-08
feat: added new evaluate command to check the score of the asyncapi document

AayushSaini101:
Resolves: #1131

* Started implementing new command to evaluate the asyncapi file:
* Use four parameters to describe the score of file:
 
```
 if (document?.info().hasDescription()) {
      scoreEvalate+=0.15;
    }
    if (document?.info().hasLicense()) {
      scoreEvalate+=0.25;
    }
    if (!document?.servers().isEmpty()) {
      scoreEvalate+=0.25;
    }
    if (!document?.channels().isEmpty()) {
      scoreEvalate+=0.35;
    }
    const finalScore = (scoreEvalate/1)*100;
```


Shurtu-gal:
@AayushSaini101 do we need a new command for the linked issue ?

AayushSaini101:
> #1131

I guess it would be if we add new command for this, What do you think ? @Amzani @peter-rr @Souvikns

Shurtu-gal:
> > #1131
> 
> I guess it would be if we add new command for this, What do you think ? @Amzani @peter-rr @Souvikns

Was seeing the incorrect issue then 👍🏻

src/commands/evaluate.ts
```diff
@@ -0,0 +1,60 @@
...
+    }
+    if (!document?.servers().isEmpty()) {
+      scoreEvaluate+=0.25;
+    }
+    if (!document?.channels().isEmpty()) {
+      scoreEvaluate+=0.35;
+    }
+    const finalScore = (scoreEvaluate/1)*100;
+
+    this.log('The score of the asyncapi document is ', +finalScore);
```
Amzani, line 58: @AayushSaini101 check if we can use the new UI library for this (see https://github.com/asyncapi/cli/issues/1214)

Amzani:
> @AayushSaini101 do we need a new command for the linked issue ?

This might be used under `validate` command, however, for a better DX and maintainability having a dedicated command looks like a better approach.

Shurtu-gal:
> > @AayushSaini101 do we need a new command for the linked issue ?
> 
> This might be used under `validate` command, however, for a better DX and maintainability having a dedicated command looks like a better approach.

Agreed, how about adding it as a flag `--evaluate`. Any way parsing will be done for both validate and evaluate. Could think some more ideas 🤔.

peter-rr:
IMO, I would opt for just adding a flag to `validate` command since the feature to be added should work like an extension for the validation process. Also I think `--score` would be semantically more clear and easier to understand than `--evaluate`. So the resulting command could be something like `asyncapi validate --score`.

WDYT? @AayushSaini101 @Amzani @Shurtu-gal

src/commands/evaluate.ts
```diff
@@ -0,0 +1,60 @@
...
+    if (document?.info().hasLicense()) {
+      scoreEvaluate+=0.25;
+    }
+    if (!document?.servers().isEmpty()) {
+      scoreEvaluate+=0.25;
+    }
+    if (!document?.channels().isEmpty()) {
+      scoreEvaluate+=0.35;
+    }
+    const finalScore = (scoreEvaluate/1)*100;
```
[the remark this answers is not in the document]
AayushSaini101: @peter-rr In the initial stage, the total weight of the coefficient is 1 that's why we need to divide the sum by 1, to get 100 Score for a valid asyncapi document.

Amzani:
> IMO, I would opt for just adding a flag to validate command since the feature to be added should work like an extension for the validation process. Also I think --score would be semantically more clear and easier to understand than --evaluate. So the resulting command could be something like asyncapi validate --score.

Evaluation and Validation are closely related, having `asyncapi validate --score` looks good to me as long as we can make the new feature easily maintainable, and users can discover it.

Shurtu-gal:
/update

There are some issues with sonar-cloud as well @AayushSaini101.

AayushSaini101:
@Shurtu-gal @Amzani @Souvikns I have added score flag in the existing validate command, and added a test case to verify the score and command. Ready for review. : ) Thanks

src/commands/validate.ts
```diff
@@ -35,4 +61,26 @@ export default class Validate extends Command {
       process.exitCode = 1;
     }
   }
+
+  private async calculateScore(filePath:string) {
```
Amzani, line 65: I suggest to extract the score logic and move it under `models` or `utils` for now. It will make this https://github.com/asyncapi/cli/pull/1200 easy.

Amzani (changes requested):

src/commands/validate.ts
```diff
@@ -4,6 +4,24 @@ import { validate, validationFlags, ValidateOptions, ValidationStatus } from '..
 import { load } from '../models/SpecificationFile';
 import { specWatcher } from '../globals';
 import { watchFlag } from '../flags';
+import { Parser} from '@asyncapi/parser/cjs';
```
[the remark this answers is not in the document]
AayushSaini101: Thanks, I missed that Done @Amzani

src/commands/validate.ts
AayushSaini101: Moved to Utils Thanks

src/utils/parsingFile.ts
```diff
@@ -0,0 +1,49 @@
+import { load } from 'js-yaml';
```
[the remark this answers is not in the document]
AayushSaini101: Reverted thanks

src/utils/scoreCalculator.ts
```diff
@@ -0,0 +1,20 @@
...
+  }
+  if (!document?.servers().isEmpty()) {
+    scoreEvaluate += 0.25;
+  }
+  if (!document?.channels().isEmpty()) {
+    scoreEvaluate += 0.35;
+  }
+  const finalScore = (scoreEvaluate / 1) * 100;
+  
+  return `The score of the asyncapi document is ${finalScore}`;
```
Shurtu-gal, line 19: How about returning the score instead of a log string. As this makes it easier to use later in another scenarios if we want.

Shurtu-gal (changes requested):

src/utils/scoreCalculator.ts
AayushSaini101: @Shurtu-gal In my opinion, proper logging would be better that enhanve the developer experience as well. @Amzani What do you think on this ?

Shurtu-gal: Yeah, I am not suggesting to take out the log message, just return the score from this function. And put logging where this function is consumed.

This way this function can be reused if needed.

Shurtu-gal (approved):
LGTM from my side. 🚀

Let's wait for @Amzani as well.

src/commands/validate.ts
```diff
@@ -12,17 +13,27 @@ export default class Validate extends Command {
     help: Flags.help({ char: 'h' }),
     watch: watchFlag(),
     ...validationFlags(),
+    score: Flags.boolean({
+      description: 'Compute the score of the AsyncAPI document',
```
Amzani, line 17: Users might might not understand how score is calculated, could we add more description and details about that ?

Amzani (changes requested):
Great PR! only thing missing for me is describing how score is calculated for users.

src/commands/validate.ts
AayushSaini101: Done @Amzani

AayushSaini101:
@Shurtu-gal @Amzani Ready for review. Hopefull for the last review cycle :)

test/integration/validate.test.ts
```diff
@@ -245,4 +245,23 @@ describe('validate', () => {
         done();
       });
   });
+  // The score of the asyncapi document is 50
```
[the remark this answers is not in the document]
AayushSaini101: Removed thanks @peter-rr

peter-rr (approved):
LGTM 🚀

Amzani (approved):
LGTM 🚀

Shurtu-gal (approved):
LGTM

Issue #1433, opened 2024-05-17
[BUG] latest download URL is not working


Praveencompro:
### Description

In one of our github actions, we were downloading the Debian package and then installing it.
For downloading the latest package we were using the URL https://github.com/asyncapi/cli/releases/latest/download/asyncapi.deb.
This URL was previously working but it's not working now.
Now, it's downloading a file with text "Not Found", refer attached image.

### Expected behavior

The URL https://github.com/asyncapi/cli/releases/latest/download/asyncapi.deb should download the latest package.

### Screenshots

![image](https://github.com/asyncapi/cli/assets/56534471/62390beb-740c-406f-a784-a7d322351878)

### How to Reproduce

1. Run the below command in bash and then check the file content.
`curl -OL https://github.com/asyncapi/cli/releases/latest/download/asyncapi.deb`

### 🥦 Browser

None

### 👀 Have you checked for similar open issues?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Contributing Guidelines?

- [X] I have read the [Contributing Guidelines](https://github.com/asyncapi/.github/blob/master/CONTRIBUTING.md)

### Are you willing to work on this issue ?

None

Amzani:
Somehow the release job failed https://github.com/asyncapi/cli/actions/runs/9091771147/job/24987034747

AayushSaini101:
@Amzani I am on it

Shurtu-gal:
Could get these to open again as well :- 

https://github.com/oclif/oclif/issues/1175

Amzani:
This can help as well 
https://github.com/oclif/oclif/issues/145#issuecomment-523517048

KhudaDad414:
> This can help as well [oclif/oclif#145 (comment)](https://github.com/oclif/oclif/issues/145#issuecomment-523517048)

just changing the `windows-latest` to `ubuntu-latest` should resolve the issue based on that comment.


https://github.com/asyncapi/cli/blob/50d9fb1e0598991801aa6532ddd89833a88a748e/.github/workflows/upload-release-assets.yml#L30


https://github.com/asyncapi/cli/blob/50d9fb1e0598991801aa6532ddd89833a88a748e/.github/workflows/upload-release-assets.yml#L26

Amzani:
Not solved yet

PR #1440, opened 2024-05-21
fix: upload release assets not working

Amzani:
Fixes https://github.com/asyncapi/cli/issues/1433


Shurtu-gal (commented):
@Amzani I believe we would need to add installation for `nsis` and `7zip` as well as its needed for pack:win.

image

Shurtu-gal (approved):
Great Idea 🚀

Issue #1427, opened 2024-05-15
[FEATURE] Improve storage of `.asyncapi-analytics` config file


peter-rr:
### Why do we need this improvement?

Main ideas taken from @asyncapi/developer_experience_wg  [slack discussion](https://asyncapi.slack.com/archives/C06G92DN05D/p1713226100765919):
cc @smoya @fmvilas 

1. We are handling the same analytics config file for [integration tests](https://github.com/asyncapi/cli/blob/master/test/integration/config/analytics.test.ts) and [source code](https://github.com/asyncapi/cli/blob/01f068a62de34d6deeb41d7f00b1ec2fd4b104d8/src/commands/config/analytics.ts#L21), which might lead to potential conflicts.

2. Also we are storing that analytics config file at home directory, which is not a recommended practice.

### How will this change help?

1. We can avoid potential conflicts by not mixing testing data with real user data.

2. Not polluting the user's home directory.

### How could it be implemented/designed?

1. Handling **different config files** for testing and source code.

2. The analytics config file should have a **configurable path**, and also be interchangeable between the different apps (like CLI, Studio), so we can know when the user is jumping from one app to another.

### 🚧 Breaking changes

No

### 👀 Have you checked for similar open issues?

- [X] I checked and didn't find a similar issue

### 🏢 Have you read the Contributing Guidelines?

- [X] I have read the [Contributing Guidelines](https://github.com/asyncapi/.github/blob/master/CONTRIBUTING.md)

### Are you willing to work on this issue?

No, someone else can work on it.

peter-rr:
@Amzani How can I add this issue to the DX working group's Kanban?

Amzani:
@peter-rr it's automatically added

peter-rr:
@fmvilas @smoya I'm trying to find out a good solution for what we talked about [here](https://asyncapi.slack.com/archives/C06G92DN05D/p1713226100765919), but I'm kind of stuck right now 🤔 

Regarding the usage of a configurable path for the analytics config file, should we interact with the user to get the file path desired and then use the value passed as a parameter? Could you provide any suggestion or example about this topic? Thanks, folks 😄

smoya:
> @fmvilas @smoya I'm trying to find out a good solution for what we talked about [here](https://asyncapi.slack.com/archives/C06G92DN05D/p1713226100765919), but I'm kind of stuck right now 🤔
> 
[...]

PR #1451, opened 2024-05-27
feat: make analytics config file's path configurable

peter-rr:
This PR will allow us to make the storage of `.asyncapi-analytics` config file configurable.

**Related issue(s)**
Relates to #1427 

cc @smoya @fmvilas

### DIFF
--- a/docs/metrics_collection.md
+++ b/docs/metrics_collection.md
@@ -39,6 +39,11 @@ We are making use of [New Relic API](https://docs.newrelic.com/docs/apis/intro-a
 
 Metrics won't be collected in CI environments, or when the "CI" env variable is set up to "true".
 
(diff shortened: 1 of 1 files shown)


Shurtu-gal:
> However, as I recommended in the Slack thread you shared, the best way to work with tests is to create an interface for the storage of the config, and swap it in tests (from in file, to in memory).

Will we be doing this as well?

peter-rr:
> > However, as I recommended in the Slack thread you shared, the best way to work with tests is to create an interface for the storage of the config, and swap it in tests (from in file, to in memory).
> 
> Will we be doing this as well?

👍 Yeah, I've opened another PR to cover the changes related to tests: https://github.com/asyncapi/cli/pull/1452

Souvikns (approved):

PR #1456, opened 2024-06-05
fix: update @asyncapi/parser to v3.0.16 and @smoya/multi-parser to v5.0.8

smoya:


peter-rr (commented):
LGTM 👍 Let me update the multi-parser dependency at `@smoya/asyncapi-adoption-metrics` as well.

Souvikns (approved):

PR #1466, opened 2024-06-13
fix: update @smoya/asyncapi-adoption-metrics to 2.4.9 version

peter-rr:
This PR will update `@smoya/asyncapi-adoption-metrics` dependency to the latest version.

PR #1469, opened 2024-06-19
fix: update @smoya/asyncapi-adoption-metrics to 2.4.9 version [DUPLICATED]

peter-rr:
This PR will update @smoya/asyncapi-adoption-metrics dependency to the latest version.

Duplicates #1466


peter-rr:
Not valid anymore since https://github.com/asyncapi/cli/pull/1466 has been finally merged.

PR #1472, opened 2024-06-20
feat: support multiple versions of generator

### LINKED ISSUE #1388: Support for multiple versions of generator
So generator reach out v2 dropping support for some node versions: https://github.com/asyncapi/generator/releases/tag/v2.0.0

Normally according to https://github.com/asyncapi/cli/issues/573 we should bump generator in cli and release new major of CLI. Basically generator drops support for node prior v18 and cli still supports older versions

but this is not a problem

problem is the generator functionality that we have, to prevent templates from failing with new generator version that may have breaking changes, we have such setting: https://github.com/asyncapi/java-template/blob/master/package.json#L102

so if we bump v2 generator in CLI, all the users of templates, will be blocked until templates developers will update config to support v2. Makes sense

so yeah:
- do we start supporting both generators here, and have a logic that basing on the config of the template, different generator is used
- or should we assume people use CLI properly, they use always a fixed version, and the same with templates - always specific version. So they will not notice issues, and basically upgrade only when they feel a need and other projects are ready?

## Scope

- [ ] Discover how we can support both generators (v2 and < v2)
- [ ] Implementation
- [ ] Add new tests
- [ ] Communicate the change

peter-rr:
This PR will allow us to find a solution to support different versions (v1 and v2) of generator in the CLI. Based on the config of the templates a different version of generator will be used.

**Related issue(s)**
Resolves #1388


derberg:
@peter-rr do you plan to continue with this one?

peter-rr:
Closing this PR in favour of https://github.com/asyncapi/cli/pull/1513

PR #1477, opened 2024-07-02
fix: add missing file path flag to instructions

andrew-rosca:


andrew-rosca:
I realized afterwards that only the HTML template uses this `-f` flag to specify the input file path. This is inconsistent with the other templates, which don't require a flag.

That's a problem with the HTML template though, so I'm closing this PR.

PR #1376, opened 2024-04-19
feat!: generate models for v3 through Modelina CLI

jonaslagoni:
**Description**

This PR enables https://github.com/asyncapi/cli/issues/883 to generate models through Modelina CLI and replaces https://github.com/asyncapi/cli/pull/1155.

**Related issue(s)**
Fixes https://github.com/asyncapi/cli/pull/1376
Blocked by https://github.com/asyncapi/cli/pull/1228
Blocked by https://github.com/asyncapi/modelina/pull/1952
Blocked by https://github.com/asyncapi/modelina/pull/2008
Blocked by https://github.com/asyncapi/modelina/pull/2010
Blocked by https://github.com/asyncapi/modelina/pull/2009

PR #1479, opened 2024-07-12
fix: modelina supporting asyncapi v3

Souvikns:
**Description**
Removing the check for spec v3 as now modelina support AsyncAPI spec v3.

**Related issue(s)**

See also https://github.com/asyncapi/cli/pull/1376

### DIFF
--- a/src/commands/generate/models.ts
+++ b/src/commands/generate/models.ts
@@ -30,9 +30,7 @@ export default class Models extends Command {
     }
 
     const inputFile = (await load(file)) || (await load());
-    if (inputFile.isAsyncAPI3()) {
-      this.error('Generate Models command does not support AsyncAPI v3 yet, please checkout https://github.com/asyncapi/modelina/issues/1376');
-    }
+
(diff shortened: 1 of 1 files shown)


Shurtu-gal (dismissed):
LGTM 🚀

Shurtu-gal:
@Souvikns could you push a empty commit. The sonarcloud analysis seems to be stuck.

Souvikns:
Sonarcloud is still not working @Shurtu-gal

Shurtu-gal (approved):

derberg:
@Shurtu-gal I have no idea. My fix is go to Sonarcloud, remove project, add project and works  Then again it fails. I suggest you folks to make it a non-mandatory check, so a check that even if fails, you may merge.

we need to work on https://github.com/asyncapi/.github/issues/306

but also in this issue you will see a hint from Sergio, that it might be just because of some silly change sonar did on their side with check name

didacrios (approved):

adammfrank:
Can this be removed in generate fromTemplate as well?

PR #1487, opened 2024-07-28
chore: migrate github-action to cli

akshatnema:
**Description**

PR to add the github action for cli from https://github.com/asyncapi/github-action-for-cli.

**Related issue(s)**
https://github.com/asyncapi/github-action-for-cli/issues/396


aeworxet:
@asyncapi/bounty_team

github-action/lib/index.js
Shurtu-gal, line 1: @akshatnema this has been already changed and we are using bash script right now.

akshatnema: So, it's not needed in action, now right?

Shurtu-gal: Yeah, you have to add the bash script.

akshatnema: So, do I have to convert this node js script into a bash script? Because I don't see any bash script present in github action repo.

Shurtu-gal:
@akshatnema any blockers being faced by you?

derberg:
but the GH action do not need package.json as it is bash and docker based only. So no need for monorepo afaik - I think we talked about it

akshatnema:
@derberg @Shurtu-gal  ok so should we do it in a manner that we completely remove the versioning from github action? Since, there is no need of a separate package.json, our whole github action depends on the working of `entrypoint.sh` and the versioning for github action scripts can be added to existing package.json of cli, right?

Shurtu-gal:
> @derberg @Shurtu-gal ok so should we do it in a manner that we completely remove the package.json from github action? Since, there is no need of a separate package.json, our whole github action depends on the working of `entrypoint.sh` and the versioning for github action scripts can be added to existing package.json of cli, right?

That's correct. If I remember correctly the package.json was there just to support the workflows already present.

Shurtu-gal:
@akshatnema sonarcloud is complaining about the Dockerfile building on a image with root as the default user. Maybe the user being added could be used for further steps through
```
USER myuser
```

I may have missed out on this during migrating to bash.

Shurtu-gal (commented):
Could you add the test-workflow as well to see if the action is working as intended?

.github/workflows/test-action.yml
```diff
@@ -0,0 +1,344 @@
...
+          regex='docker:\/\/asyncapi\/github-action-for-cli:([0-9.]+)';
+          [[ $action =~ $regex ]];
+          action_version=${BASH_REMATCH[1]};
+          echo "action_version=$action_version" >> $GITHUB_OUTPUT
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+      - name: Build Docker image and export
+        uses: docker/build-push-action@v5
+        with:
+          context: .
```
Shurtu-gal, line 52: The context should be different I believe as the DockerFile is inside github-action folder.

Shurtu-gal:
@aeworxet this issue is pretty complicated as some things related to containerizing part is still not working and we still have to do the automation of version locking and bumping. @akshatnema is also pretty new to the CLI repo so is it possible to reclassify this bounty as `advanced` and extend it for 1-2 weeks.

Shurtu-gal (approved):
LGTM. The version automation part would be done in a separate PR.

PR #1492, opened 2024-08-07
fix: update package `@asyncapi/bundler` to version `0.6.1`

aeworxet:
This PR updates the `@asyncapi/bundler` package to version `0.6.0`.

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -5,7 +5,7 @@ weight: 30
 
 ## Node and npm
 
-To use the AsyncAPI CLI tool, you must install NPM and Node.js version 10 or higher. To check if you already have both installed, run the following commands in your terminal:
+To use the AsyncAPI CLI tool, you must install NPM and Node.js version 16 or higher. To check if you already have both installed, run the following commands in your terminal:
 
 ```sh
 # check if node is installed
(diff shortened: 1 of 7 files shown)


Shurtu-gal:
@aeworxet the tests are currently not passing.

aeworxet:
`Bundler` v0.6.0 [uses](https://github.com/asyncapi/bundler/blob/master/src/util.ts#L161) [`structuredClone()`](https://developer.mozilla.org/en-US/docs/Web/API/structuredClone), which was [added only in Node 17](https://developer.mozilla.org/en-US/docs/Web/API/structuredClone#browser_compatibility).

Should it be stated in [CLI's documentation](https://github.com/asyncapi/cli/blob/master/docs/installation.md?plain=1#L8) that the minimal version of Node.js is v18 or is it too early to make users upgrade to v18 and I should rewrite Bundler's functionality with a [polyfill](https://www.npmjs.com/package/@ungap/structured-clone) or without `structuredClone()` at all (good ol' `JSON.parse(JSON.stringify())`) (downgrade?)

Souvikns:
Well according to the website v20 is the current LTS version, but still v16 is wildly used so I think we should still support v16 for some time in the near future.

aeworxet:
Replaced in `Bundler` native `structuredClone()` with a [polyfill](https://www.npmjs.com/package/@ungap/structured-clone) to keep compatibility with Node.js v16.

I will update this PR as soon as https://github.com/asyncapi/bundler/pull/183 is merged and `Bundler` `v0.6.1` is released.

Souvikns (approved):
LGTM

PR #1513, opened 2024-09-06
feat: add flag for using generator v2

Souvikns:
**Description**
In this PR I am adding a new Flag to use generator v2+, I am trying to change as little as possible from users original version of CLI. Currently most temapltes are not supporting v2 of generator, so it is better to have flag `--use-new-generator`, once more templates start supporting v2 of generator we can update the flag. 

```bash
asyncapi generate fromTemaplate ./spec.yaml @asyncapi/html-template --use-new-generator
```

**Related issue(s)**

See also #1388


check/asyncapi.md
Shurtu-gal, line 1: Where are these being used?

Shurtu-gal (approved):

Shurtu-gal:
@Souvikns Please merge if ready.

aeworxet:
@asyncapi/bounty_team

peter-rr (approved):
LGTM 👍

PR #1543, opened 2024-10-13
feat: add workflow to release github-action

akshatnema:
**Description**
Added a workflow that will complete following tasks:
- Release the latest version of github-action in Dockerhub
- Update action.yml with latest version of docker image

**Related issue(s)**


aeworxet:
@asyncapi/bounty_team

.github/workflows/release-action-docker.yml
```diff
@@ -0,0 +1,56 @@
...
+        run: |
+          echo ${{env.DOCKER_PASSWORD}} | docker login -u ${{env.DOCKER_USERNAME}} --password-stdin
+          npm run action:docker:build
+          docker tag asyncapi/github-action-for-cli:latest asyncapi/github-action-for-cli:${{ steps.version.outputs.value }}
+          docker push asyncapi/github-action-for-cli:${{ steps.version.outputs.value }}
+          docker push asyncapi/github-action-for-cli:latest
+
+      - name: Update action.yml with version
+        run: |
+          VERSION=${{ steps.version.outputs.value }}
```
Shurtu-gal, line 39: The update to this action.yml won't be a part of the release though if this is done like this.

Shurtu-gal (changes requested):

.github/workflows/release-action-docker.yml
akshatnema: @Shurtu-gal  Yeah, do you want to run this workflow as part of pre-release or this update should be done before we build docker image of github-action?

Shurtu-gal: @akshatnema pre-release

.github/workflows/release-action-docker.yml
```diff
@@ -0,0 +1,56 @@
+name: Release Github action for CLI
+
+on:
+  release:
+    types:
+      - prereleased
```
Shurtu-gal, line 6: I believe this only works when you do a manual pre-release.
image

Shurtu-gal (changes requested):

.github/workflows/release-action-docker.yml
akshatnema: Yeah, so you can then proceed with the release with the release, once the workflow is successful, right?

Shurtu-gal: But the release is no longer automated if it has to be done manually. Let me correct the misunderstanding pre-release event will only work when a pre-release is done, which is done nowhere as far as I know.

Shurtu-gal (approved):

Issue #1540, opened 2024-10-03
Better UX support for using --use-new-generator flag. 


Souvikns:
I'm afraid there might be issues with implementation in https://github.com/asyncapi/cli/pull/1513 as it assumes kinda that v2 is the last major generator version.

`--use-new-generator` scares me 😄 and generating without it, for template that needs new generator just gives `Generator Error: Installation failed`. Weren't you able to intercept the error, that would say that template do not support old generator, and in the error you could drop some good info for the user? like to use the flag

also look at this strange behavior with logs, when I use the `--use-new-generator`  flag:

```
asyncapi generate fromTemplate https://raw.githubusercontent.com/asyncapi/spec/refs/heads/master/examples/streetlights-kafka-asyncapi.yml https://github.com/derberg/html-template#bumphtml --force-write --use-new-generator -o dupa
┌  AsyncAPI Generator
│
◒  Generation in progress. Keep calm and wait a bit...(⠂⠂⠂⠂⠂⠂⠂⠂⠂⠂⠂⠂⠂⠂⠂⠂⠂⠂) ⠋ idealTree:generato
◐  Generation in progress. Keep calm and wait a bit(⠂⠂⠂⠂⠂⠂⠂⠂⠂⠂⠂⠂⠂⠂⠂⠂⠂⠂) ⠋ idealTree:generator-v
◓  Generation in progress. Keep calm and wait a bit(⠂⠂⠂⠂⠂⠂⠂⠂⠂⠂⠂⠂⠂⠂⠂⠂⠂⠂) ⠋ idealTree:generator-v
◑  Generation in progress. Keep calm and wait a bit(⠂⠂⠂⠂⠂⠂⠂⠂⠂⠂⠂⠂⠂⠂⠂⠂⠂⠂) ⠋ idealTree:generator-v
◒  Generation in progress. Keep calm and wait a bit(⠂⠂⠂⠂⠂⠂⠂⠂⠂⠂⠂⠂⠂⠂⠂⠂⠂⠂) ⠋ idealTree:generator-v
◐  Generation in progress. Keep calm and wait a bit(⠂⠂⠂⠂⠂⠂⠂⠂⠂⠂⠂⠂⠂⠂⠂⠂⠂⠂) ⠋ idealTree:generator-v
[...]

PR #1552, opened 2024-10-20
feat: added the progress bar for generate command.

### LINKED ISSUE #1410: [FEATURE] Adding a progress bar for CLI Commands to show the progress status

It would be good if we can add progress bar when we execute following command to make CLI more user attractive and increase developer experience: 

Progress bar can be added while:
* Model Generation 
* Pulling templates from private repositories 
* Generating templates 

It would make CLI more user attractive and increase developer experience:  

### Screenshots

Sample Progress Bar: 
image

Set up the progress start and progress end between the process start 

### 🚧 Breaking changes

No

ItshMoh:
**Description**
Added a progress bar for the generate command in the cli.

https://www.loom.com/share/72534057ea2644f997bae71fb4c79957?sid=7e05ac94-c362-4860-9dfb-333ff07bfff4

**Related issue(s)**
fixes #1410


AayushSaini101:
Thanks for the PR @ItshMoh Please include this in all the commands of the CLI and fix Quality Gate

ItshMoh:
> Thanks for the PR @ItshMoh Please include this in all the commands of the CLI and fix Quality Gate

Sure @AayushSaini101

Shurtu-gal:
@ItshMoh how is the progress being calculated ?

ItshMoh:
> @ItshMoh how is the progress being calculated ?

Hey @Shurtu-gal  I have updated the logic. So in the new logic here i have divided it into 3 steps. The first step is Initializing generator , Generating Files and the last for completion of generation. After the step the progress bar increases.
Do you have anything extra that i can add?

Shurtu-gal:
> Hey @Shurtu-gal I have updated the logic. So in the new logic here i have divided it into 3 steps. The first step is Initializing generator , Generating Files and the last for completion of generation. After the step the progress bar increases. Do you have anything extra that i can add?

What I am not understanding is how is a progress bar different from the current spinner we have if we don't have any way to measure the current progress in generation. 

Note: @Souvikns has also proposed removing the spinner altogether https://github.com/amplication

ItshMoh:
>
> What I am not understanding is how is a progress bar different from the current spinner we have if we don't have any way to measure the current progress in generation.
> 
> Note: @Souvikns has also proposed removing the spinner altogether https://github.com/amplication

Hey @Shurtu-gal Should i remove the spinner alltogether ?

Shurtu-gal:
@Souvikns do require this feature to be there as we don't have any set time or some kind of progress parameter I guess this would rather just be a aesthetic thing.

Souvikns:
Yeah what we have right now work, but having a progress bar is causing some problems with new generator as well. So we have to consider that as well. @ItshMoh do consider this https://github.com/asyncapi/cli/issues/1540 as well, if you can add a progress bar and not create the same problem with the new generator that there is no issue.

Shurtu-gal:
Any updates @ItshMoh?

ItshMoh:
Hey @Shurtu-gal I am not able to  work on it currently.  So closing the PR.

PR #1555, opened 2024-10-22
fix: install latest npm 

Souvikns:
**Description**

For some reason, gh action skipped when `matrix.ox = 'windows-latest`, so changed the if condition in the PR.


aeworxet:
@asyncapi/bounty_team

Shurtu-gal (approved):

PR #1565, opened 2024-11-04
fix: update action.yml

Shurtu-gal:
**Description**

- Update action.yml for release of action purpose.


Souvikns:
@Shurtu-gal slack notify is failing.

Shurtu-gal:
> @Shurtu-gal slack notify is failing.

This does not pertain to this PR but is rather an issue in our workflows. I would make a issue in `.github` and try to fix this.

Meanwhile I believe this can be ignored here.

Souvikns (approved):

PR #1605, opened 2024-12-21
fix: prepare action for release

Shurtu-gal:
Changing name of the action as apparently name should be unique.

image


Shurtu-gal:
Please take a look @Souvikns

Souvikns (approved):

Issue #1627, opened 2025-01-14
[FEATURE] introduce new command `asyncapi start preview`


derberg:
### Why do we need this improvement?

ATM we have `asyncapi start studio`. It is cool, starts local Studio editor, so people can edit in browser. Problem is that users still locally have the same limitation as https://studio.asyncapi.com/ - no support in web app for references pointing to the local file system.

I think we need new command: `asyncapi start preview`

- it can still use studio
- it would have to take a AsyncAPI file before starting studio and bundle local references )using the same mechanism as `asyncapi bundle`) and do it in memory (not save to file) and pass to studio
- studio would have to be started with `readOnly=true` param to hide editor (it has some limits though https://github.com/asyncapi/studio/issues/591)

And the preview would have to be hot reloaded on any change in the file on local

> example with local file reference: https://www.asyncapi.com/docs/tutorials/kafka/configure-kafka-avro#reusing-existing-avro-schemas

### How will this change help?

Solve problem that to preview my local AsyncAPI document on local, that uses local file references, I need to generate docs or use special plugins.

### How could it be implemented/designed?

already described in main description.

although maybe instead of using studio better just use react component directly? 🤔 something to validate during implementation. This is how preview was done in https://github.com/asyncapi/vs-asyncapi-preview

### 🚧 Breaking changes

No

### 👀 Have you checked for similar open issues?

- [X] I checked and didn't find a similar issue

### 🏢 Have you read the Contributing Guidelines?

- [X] I have read the [Contributing Guidelines](https://github.com/asyncapi/.github/blob/master/CONTRIBUTING.md)

### Are you willing to work on this issue?

No, someone else can work on it.

GauravSaini01:
I am working on this.

Paribesh01:
@derberg  The readOnly parameter cannot be used with locally stored files; it must be used with a URL. So, when starting this 
new command with a local file, the readOnly mode will not work.

derberg:
yes, `readOnly` works with URL only:
- that could be improved in https://github.com/asyncapi/studio
- or we could start a tiny server to host the preview file, and use `url` query param on studio to point to localhost. We already for example start a websocket for `asyncapi start studio` command

what I'm thinking about is - should we use studio, or maybe better use https://github.com/asyncapi/asyncapi-react (used in studio for preview) directly. It is super simple to use like https://github.com/asyncapi/asyncapi-react/blob/master/docs/usage/web-component.md

I would say requirement is:
[...]

PR #1633, opened 2025-01-17
feat: new command preview

Paribesh01:
Issue Solved : #1627 

PR Summary:
This PR introduces the asyncapi start preview command to solve the limitation of local references in the asyncapi start studio command. It allows users to preview AsyncAPI documents with local file references directly in Studio.

What was done:
Bundled the reference files into one and served it locally.
Passed the bundled document to Studio via a URL with a readOnly=true parameter to hide the editor.

How it helps:
This change eliminates the need for workarounds, enabling users to seamlessly preview local AsyncAPI documents with references  file.


Shurtu-gal:
This feature has been implemented through #1753

Issue #1626, opened 2025-01-14
[BUG] Inaccurate error in `asyncapi optimize`


derberg:
### Describe the bug.

I have AsyncAPI document with broken message reference in line 39:

```yml
asyncapi: 3.0.0

info:
  title: Account Service
  version: 1.0.0
  description: This service is in charge of processing user signups

servers: 
  development:
    host: loca{dupa}lhost:5672
    description: Development AMQP broker.
    protocol: amqp
    protocolVersion: 0-9-1

  staging:
    host: rabbitmq-staging.in.mycompany.com:5672
    description: RabbitMQ broker for the staging environment.
    protocol: amqp
    protocolVersion: 0-9-1

  production:
    host: rabbitmq.in.mycompany.com:5672
    description: RabbitMQ broker for the production environment.
    protocol: amqp
    protocolVersion: 0-9-1

channels:
  userSignedup:
    address: user/signedup
    messages:
      UserSignedUp:
        $ref: '#/components/messages/UserSignedUp'
operations:
  sendUserSignedup:
    action: send
    channel:
      $ref: '#/channels/userSignedup'
    messages:
      - $ref: '#/components/messages/UserSignedUp'
components:
  messages:
    UserSignedUp:
      payload:
        type: object
        properties:
          displayName:
            type: string
            description: Name of the user
          email:
            type: string
            format: email
            description: Email of the user
```

When I run: `asyncapi optimize ./asyncapi-client.yaml` I get error like
```bash

[
  {
    code: 'asyncapi3-operation-messages-from-referred-channel',
    message: 'Operation message does not belong to the specified channel.',
    path: [ 'operations', 'sendUserSignedup', 'messages', '0' ],
    severity: 0,
    range: { start: [Object], end: [Object] }
  }
]
ValidationError: There is no file or context with name "./asyncapi-client.yaml".

```

The array with validation error info is correct. I mean, maybe object should be rendered instead of `[Object]` but in general it is ok.

The problem is with the last line:
`ValidationError: There is no file or context with name "./asyncapi-client.yaml".`

This error just doesn't make sense, the path reference was correct, otherwise validation would not catch the error with the message

### Expected behavior

Remove the misleading error message

### Screenshots

no

### How to Reproduce

already described in description

### 🥦 Browser

None

### 👀 Have you checked for similar open issues?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Contributing Guidelines?

[...]

PR #1634, opened 2025-01-18
fix: resolve error in asyncapi optimize command

GauravSaini01:
## **Related issue(s)**
https://github.com/asyncapi/cli/issues/1626

### Right File Path
Image

### Wrong File Path
- With `./asyncapi.yaml`
Automatic pick `./asyncapi.yaml` file and handle it similarly to the right file path case
- Without ` ./asyncapi.yaml`
Image

### Wrong Context name
Image
### Right Context name
- pick file that is associated with **context name** and handle it similarly to the right file path case

### `asyncapi optimize` without [spec-file]
- If current context is set
Pick the file associated with the current context name and handle it similarly to the right file path case
- If `./asyncapi.yaml` present
Automatic pick `./asyncapi.yaml` file and behave similar to Right File Path case
- without `./asyncapi.file` and current context
Image

@AayushSaini101 @Shurtu-gal please review my PR


derberg:
@GauravSaini01 rule number one when contributing is to followup on automation results. Maintainers will not fix problems for you. So if there is PR title linting failing, tests failing, you need to proactively research solution and fix things to make sure CI/CD automation is green and maintainers can do the review

/ptal 
solution here relates to code fixes purely, nothing on the command level so in my opinion doesn't really matter as I'm not a maintainer here. Just letting you know that I will not be involved in review 😄

Shurtu-gal:
Could attach a demo of the solution fixing problem in the original issue?

GauravSaini01:
@Shurtu-gal
The behavior of the command after merging the PR is already explained in the description for reference.

GauravSaini01:
@AayushSaini101
Is this PR okay, or should I make some changes?

test/integration/optimize.test.ts
Shurtu-gal, line 1: Could you add the test case for the new error type you have introduced?

Shurtu-gal (changes requested):
Rest of the PR looks good. Good job.

test/integration/optimize.test.ts
GauravSaini01: Sure! I’ll get on it, thanks!

GauravSaini01:
@Shurtu-gal 
Please take a look at this and let me know if you have any suggestions or changes.

AayushSaini101 (approved):
LGTM

AayushSaini101 (approved):

Shurtu-gal (approved):
Lgtm

PR #1631, opened 2025-01-15
feat: make asyncapi start studio consistent

lmalkam:
Description
Solves Issue : #1623 
Removed the flag -f, now user can directly input the filename from cli without using -f
Hints are presented in the cli when --help is used

PR #1635, opened 2025-01-18
fix: made studio command consistent with other commands

### LINKED ISSUE #1623: `asyncapi start studio` is not consistent with other commands
To start studio with `asyncapi start studio` you need to specify path to AsyncAPI document.

Assumption, my document location is `./asyncapi-client.yaml`.

With all other CLI commands, you just provide the path:
- `asyncapi validate ./asyncapi-client.yaml`
- `asyncapi format ./asyncapi-client.yaml`
- `asyncapi pretty ./asyncapi-client.yaml`

Now, in case of Studio, you need to do: `asyncapi start studio -f ./asyncapi-client.yaml`

And when you execute `asyncapi start studio` without pointing file, you get generic error that is not true:
```
error locating AsyncAPI document: These are your options to specify in the CLI what AsyncAPI file should be used:
```

### Fix suggestion

- remove `-f` flag from `asyncapi start studio` and make it work like in other commands
- it would probably also make sense to not make it required to point document, I think maybe better start studio, just without any file opened in it by default + drop in termina a hint that user can open specific file if they want, and how

SahilDahekar:
**Description**

- Updated `asyncapi start studio` command to input spec file as a argument rather than from a  `-f` flag.
- Added instruction in console on how to open studio with own spec file . This instruction is only visible when no spec file arg is specified .
- Updated docs to reflect the change in `asyncapi start studio` command usage.
- Added a default `asyncapi.yml` file to open in case no spec file is passed as input argument.

### **Screenshots**

- Default spec file `asyncapi.yml` opens when not specified with `start studio` command with terminal instruction on how to open own spec file .

Screenshot from 2025-01-18 18-09-54

Output :
```bash
To open a specific AsyncAPI file in the Studio, use: asyncapi start studio [your-spec-file-location]

Studio is now running at http://localhost:3210?liveServer=3210&studio-version=0.20.2.
You can open this URL in your web browser, and if needed, press Ctrl + C to stop the process.
Watching changes on file asyncapi.yml
```

Screenshot from 2025-01-18 18-09-25

- Instruction is not logged when using own spec file .

Screenshot from 2025-01-18 18-10-32

Output :
```bash
Studio is now running at http://localhost:3210?liveServer=3210&studio-version=0.20.2.
You can open this URL in your web browser, and if needed, press Ctrl + C to stop the process.
Watching changes on file /home/sahil/test.yaml
```

**Related issue(s)**

Resolves #1623

### DIFF
--- a/asyncapi.yml
+++ b/asyncapi.yml
@@ -1,0 +1,9 @@
+asyncapi: '3.0.0'
+info:
+  title: Welcome to AsyncAPI Studio
+  version: '1.0.0'
+  description: This is a sample AsyncAPI document to help you get started. Define your async APIs, channels, messages, and bindings here.
+  contact:
+    name: AsyncAPI 
+    url: https://www.asyncapi.com
+    email: support@asyncapi.org
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -786,10 +786,12 @@ starts a new local instance of Studio
 
 ```
 USAGE
-  $ asyncapi start studio [-h] [-f <value>] [-p <value>]
+  $ asyncapi start studio [SPEC_FILE] [-h] [-p <value>]
+
+ARGUMENTS
+  SPEC-FILE  spec path, url, or context-name
 
 FLAGS
-  -f, --file=<value>  path to the AsyncAPI file to link with Studio
   -h, --help          Show CLI help.
   -p, --port=<value>  port in which to start Studio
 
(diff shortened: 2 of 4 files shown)


derberg:
/ptal

suggestion to maintainers to close this PR as it is duplicate of https://github.com/asyncapi/cli/pull/1631 and actually author of https://github.com/asyncapi/cli/pull/1631 is doing the right thing of having discussion in the issue on the scope

PR #1637, opened 2025-01-19
feat: added new command for asyncapi start preview

### LINKED ISSUE #1627: [FEATURE] introduce new command `asyncapi start preview`

Lemeri123:
Added code to implement the asyncapi start preview command

closes #1627 

cc. @derberg


Souvikns:
@Lemeri123 the tests are failing, can you have a look

Shurtu-gal:
Closing this due to non-respondence by contributor and it being a bounty issue.

PR #1638, opened 2025-01-20
feat: remove -f flag from asyncapi start command

### LINKED ISSUE #1623: `asyncapi start studio` is not consistent with other commands

Adi-204:
**Description**
Remove -f flag from asyncapi start command 
asyncapi start command can now work without filename also

Fixes #1623 

Case 1 Filename given in command by user.
Screenshot 2025-01-20 200812

Case 2 Filename not given in command by user but asyncapi.yaml default file is present in current working directory.
Screenshot 2025-01-20 200724

Case 3 Filename not given in command by user and asyncapi.yaml default file is not present in current working directory. Open studio with empty file.
Screenshot 2025-01-20 200949

Studio for Case 3
image

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -38,33 +38,35 @@ USAGE
 # Commands
 
 <!-- commands -->
-* [`asyncapi bundle`](#asyncapi-bundle)
-* [`asyncapi config`](#asyncapi-config)
-* [`asyncapi config analytics`](#asyncapi-config-analytics)
-* [`asyncapi config context`](#asyncapi-config-context)
-* [`asyncapi config context add CONTEXT-NAME SPEC-FILE-PATH`](#asyncapi-config-context-add-context-name-spec-file-path)
-* [`asyncapi config context current`](#asyncapi-config-context-current)
-* [`asyncapi config context edit CONTEXT-NAME NEW-SPEC-FILE-PATH`](#asyncapi-config-context-edit-context-name-new-spec-file-path)
-* [`asyncapi config context init [CONTEXT-FILE-PATH]`](#asyncapi-config-context-init-context-file-path)
-* [`asyncapi config context list`](#asyncapi-config-context-list)
-* [`asyncapi config context remove CONTEXT-NAME`](#asyncapi-config-context-remove-context-name)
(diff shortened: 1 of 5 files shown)


derberg (commented):
it looks pretty good, although please take into account I'm not a maintainer here, so in depth review is needed from their side. I'm not involved here for some time, afaik you should not use console.log directly.

anyway, from code, without manual testing, I see you fulfilled requirements. I hope you did manual testing.

of course there is also a case of missing unit tests

Shurtu-gal:
After a brief look over the PR, I found the following:

- I couldn't understand the reason behind removing `-f` flag as if the only issue is not having default file, then default can be added to flag itself.
- If file passed doesn't exist it should not open the studio and just throw an error.

derberg:
as specified in linked issues, studio command works like this now: `asyncapi start studio -f ./asyncapi-client.yaml` which is totally out of sync with other commands. Studio is the only command that requires you to use `-f`, so if you use multiple commands (like I do) whenever you use studio, you always face error as you do asyncapi start studio ./asyncapi-client.yaml`, and then need to remind yourself what is the right way

> If file passed doesn't exist it should not open the studio and just throw an error.
studio is also useful to start locally without any file, to see it in action

Shurtu-gal:
Okay, makes sense.

Shurtu-gal (approved):
LGTM

PR #1645, opened 2025-01-25
feat: introduced asyncapi start preview command

GauravSaini01:
## **Related issue(s)**
https://github.com/asyncapi/cli/issues/1627

## **Description**
Screenshot 2025-01-25 215858

AsyncAPI can include local references files (e.g.  `$ref: './temp/feature.yaml#/components/messages/UserSignedUp'`)
If any local reference file or the asyncapi.yaml file changes, the studio will hot reload automatically.

Screenshot 2025-01-25 215640

Note : 
 - In the studio, bundled files are passed through the `asyncapi-bundled.yaml` file instead of in memory.
 - The `readOnly` feature is not implemented  

@AayushSaini101 @Shurtu-gal please review my PR

### DIFF
--- a/src/commands/start/preview.ts
+++ b/src/commands/start/preview.ts
@@ -1,0 +1,151 @@
+import Command from '../../core/base';
+import { start as startStudio } from '../../core/models/Studio';
+import { load } from '../../core/models/SpecificationFile';
+import bundle from '@asyncapi/bundler';
+import path from 'path';
+import fs from 'fs';
+import { Args } from '@oclif/core';
+import chokidar from 'chokidar';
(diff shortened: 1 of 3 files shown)


AayushSaini101:
@GauravSaini01 can you add some test cases as well ?

GauravSaini01:
Hey @AayushSaini101,

I added the following test:
image
However, the terminal seems to be stuck after passing all tests successfully, which is why the PR CI testing fails. I’ve spent time trying to resolve it, but I’m still not sure what might be causing it. Could you please help me figure out what’s causing the issue and how I can improve the test to avoid this problem?

Thanks

PR #1651, opened 2025-02-07
fix: starting of studio fixed when using example with new file

### LINKED ISSUE #1648: [BUG] Creating a asyncapi specification file form examples with different name fails to open it in studio.
### Describe the bug.

When using `asyncapi new -e any-example` to generate a new asyncapi specification file based on an example file after providing a different file name than the example file and selecting `yes` to open it in studio cli fails to do it.

Example

using `anyof-asyncapi.yml` example file and providing it a different name as `different` than the example file name and selecting `yes` to open it in studio we get `SpecificationFileError: File differnt does not exist.` even after the file has been created.

but if the same file name is provided the studio opens perfectly.

### Expected behavior

Studio should open with the custom file name given by user.

### Screenshots

Image

Image

### How to Reproduce

1. Create an new file using `asyncapi new -e anyof-asyncapi.yml`
2. Give it a different file name than the example.
3. Select yes when asked to open in studio.
4. Gives `Specification Error`

### 🥦 Browser

None

neoandmatrix:
**Description**

This PR introduces the changes to fix the starting of studio when used with an example while creating a new file.

Approach:-
Upon debugging found that to start studio the extension had also to be passed with file name so that was added to be passed if not passed by default. `yaml` was passed as it was default extension used by cli when creation of files if no extension is specified.

**Related issue(s)**
Resolves #1648

### DIFF
--- a/.changeset/wild-rockets-care.md
+++ b/.changeset/wild-rockets-care.md
@@ -1,0 +1,5 @@
+---
+"@asyncapi/cli": patch
+---
+
+fix: starting of studio fixed when using example with new file
--- a/src/commands/new/file.ts
+++ b/src/commands/new/file.ts
@@ -127,6 +127,7 @@ export default class NewFile extends Command {
     selectedTemplate = selectedTemplate || DEFAULT_ASYNCAPI_TEMPLATE;
 
     await this.createAsyncapiFile(fileName, selectedTemplate);
+    fileName = fileName.includes('.') ? fileName : `${fileName}.yaml`;
     if (openStudio) { startStudio(fileName, flags.port || DEFAULT_PORT);}
   }
 


AayushSaini101:
@neoandmatrix please add screenshots for the validation thanks

neoandmatrix:
> @neoandmatrix please add screenshots for the validation thanks

My apologies forgot to add it beforehand.

Here is the result after changes. Even without extension of file given studio opens it correctly.
image

AayushSaini101:
> > @neoandmatrix please add screenshots for the validation thanks
> 
> My apologies forgot to add it beforehand.
> 
> Here is the result after changes. Even without extension of file given studio opens it correctly. image

@neoandmatrix can you add some test cases as well ?

neoandmatrix:
> > > @neoandmatrix please add screenshots for the validation thanks
> > 
> > My apologies forgot to add it beforehand.
> > 
> > Here is the result after changes. Even without extension of file given studio opens it correctly. image
> 
> @neoandmatrix can you add some test cases as well ?

Sure will do.

neoandmatrix:
@AayushSaini101 I tried to add test for above changes but the problem that i am facing is that the changes in this PR allow opening of studio which if i do through tests open the studio which opens fine but the tests don't continue further and get stuck. 
Also previous tests start to fail and i believe this test won't be possible to be run in a CI/CD environment as studio won't open in that.
Looking at the already present tests they also don't include tests related to opening of studio.

Any guidance on to where should i look for this or if doing something wrong, will be helpful.

Thanks.

AayushSaini101:
> @AayushSaini101 I tried to add test for above changes but the problem that i am facing is that the changes in this PR allow opening of studio which if i do through tests open the studio which opens fine but the tests don't continue further and get stuck. Also previous tests start to fail and i believe this test won't be possible to be run in a CI/CD environment as studio won't open in that. Looking at the already present tests they also don't include tests related to opening of studio.
> 
> Any guidance on to where should i look for this or if doing something wrong, will be helpful.
> 
> Thanks.

@Souvikns @Shurtu-gal is it possible to test these changes somehow, we don't have any test cases for studio before also ?

Shurtu-gal:
Only very specific things can be tested such as file being changed should send websocket events. So these things can be verified. Rest I don't believe can be tested. 

No error condition can be checked though.

*If we really need to test* then would need to look at puppeteer or sth else.

AayushSaini101:
> Only very specific things can be tested such as file being changed should send websocket events. So these things can be verified. Rest I don't believe can be tested.
> 
> No error condition can be checked though.
> 
> _If we really need to test_ then would need to look at puppeteer or sth else.

make sense, we add this as well in GSOC project` then would need to look at puppeteer or sth else.` thanks

AayushSaini101 (approved):

neoandmatrix:
> > Only very specific things can be tested such as file being changed should send websocket events. So these things can be verified. Rest I don't believe can be tested.
> > 
> > No error condition can be checked though.
> > 
> > _If we really need to test_ then would need to look at puppeteer or sth else.
> 
> make sense, we add this as well in GSOC project` then would need to look at puppeteer or sth else.` thanks

Sure will do some research on how to achieve the same.

Thanks for guidance.

AayushSaini101 (approved):

PR #1677, opened 2025-02-19
feat: asyncapi generate fromTemplate can now work in interactive mode

### LINKED ISSUE #1660: [FEATURE] `asyncapi generate fromTemplate` requires arguments despite interactive mode

Currently, the asyncapi generate command has a --no-interactive flag, meaning that by default, it should run in interactive mode. However, the arguments ASYNCAPI and TEMPLATE are still required, even though the interactive mode is already designed to prompt the user for these values. This contradicts the purpose of interactive mode.

1. The current implementation forces users to provide ASYNCAPI and TEMPLATE even when they intend to use interactive mode.
2. This makes the CLI experience inconsistent because interactive mode is supposed to allow users to provide inputs dynamically.
3. Removing these constraints will make the CLI more user-friendly, aligning it with common interactive CLI patterns.

1. Users will be able to run `asyncapi generate fromTemplate` without needing to remember or type the required arguments.
2. The interactive mode will guide users through input selection instead of throwing an error.
3. This will make the tool more intuitive, especially for new users.
4. It improves flexibility while maintaining backward compatibility for those who prefer using arguments explicitly.

### Screenshots

Image

Update the CLI argument parser to make ASYNCAPI and TEMPLATE optional. If these arguments are missing and --no-interactive is not used, prompt the user interactively.

### 🚧 Breaking changes

No

Adi-204:
**Description**
Made the 2 args for command `asyncapi generate fromTemplate` not required to make the CLI more user-friendly, aligning it with common interactive CLI patterns.
Users will be able to run `asyncapi generate fromTemplate` without needing to remember or type the required arguments.

Screenshot 2025-02-19 122607

**Related issue(s)**
Fixes #1660

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -38,33 +38,35 @@ USAGE
 # Commands
 
 <!-- commands -->
-* [`asyncapi bundle`](#asyncapi-bundle)
-* [`asyncapi config`](#asyncapi-config)
-* [`asyncapi config analytics`](#asyncapi-config-analytics)
-* [`asyncapi config context`](#asyncapi-config-context)
-* [`asyncapi config context add CONTEXT-NAME SPEC-FILE-PATH`](#asyncapi-config-context-add-context-name-spec-file-path)
-* [`asyncapi config context current`](#asyncapi-config-context-current)
-* [`asyncapi config context edit CONTEXT-NAME NEW-SPEC-FILE-PATH`](#asyncapi-config-context-edit-context-name-new-spec-file-path)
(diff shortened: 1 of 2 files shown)


AayushSaini101 (approved):

AayushSaini101:
LGTM

Shurtu-gal:
Is non interactive still possible if we pass the arguments?

Adi-204:
@Shurtu-gal Sir yes it works completely fine without --no-interactive flag.

image

PR #1716, opened 2025-03-12
Update async-cli package to use ES modules

### LINKED ISSUE #1698: [BUG] <description> Async-cli package is using commonJS module which is not appropriate for many dependencies to update
### Describe the bug.

Currently this async-cli package is using commonjs module which is not good , as many dependcies latest versions are shifted to module js and the node also recommend to shift to modulejs 

### Expected behavior

Package should be shifted to module js so that the dependecies can be upgraded to new versions 

### Screenshots

Image

### How to Reproduce

1. First i update the older dependecies 
2. then i update the code for the updated dependencies
3. but it was not working as the dependencies are using ES module but project was using common js which was cauing error during building

### 🥦 Browser

None

Recxsmacx:
Fixes #1698

Update `async-cli` package to use ES modules instead of CommonJS.

* **Update `tsconfig.json` files:**
  - Change `module` to `ESNext` in `tsconfig.json`, `assets/create-glee-app/templates/default/tsconfig.json`, and `assets/create-glee-app/templates/tutorial/tsconfig.json`.
  - Update `moduleResolution` to `node` in the same files.

* **Update `package.json`:**
  - Add `"type": "module"`.

* **Update import/export syntax:**
  - Modify import statements in `src/commands/bundle.ts`, `src/commands/convert.ts`, `src/commands/generate/fromTemplate.ts`, `scripts/fetch-asyncapi-example.js`, `scripts/releasePackagesRename.js`, and `src/index.ts` to use ES module syntax.

### DIFF
--- a/assets/create-glee-app/templates/default/tsconfig.json
+++ b/assets/create-glee-app/templates/default/tsconfig.json
@@ -4,7 +4,7 @@
     "target": "es6",
     "esModuleInterop": true,
     "moduleResolution": "node",
-    "module": "commonjs",
+    "module": "ESNext",
     "skipLibCheck": true
   }
 }
--- a/assets/create-glee-app/templates/tutorial/tsconfig.json
+++ b/assets/create-glee-app/templates/tutorial/tsconfig.json
@@ -4,6 +4,6 @@
     "target": "es6",
     "esModuleInterop": true,
     "moduleResolution": "node",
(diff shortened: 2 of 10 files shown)


Shurtu-gal:
Hey @Recxsmacx, this is not something we would be doing right now. Please feel free to contribute to other issues though.

Issue #1685, opened 2025-03-01
[FEATURE] Add Command-line Tab Completion to AsyncAPI CLI


gambhirsharma:
### Why do we need this improvement?

While using the AsyncAPI CLI for the first time, I found it frustrating to repeatedly use the `--help` flag to look up commands. To address this, I proposed adding a feature to improve the user experience. 

I initially started a [discussion](https://github.com/asyncapi/cli/discussions/1668), but since it hasn't received any responses, I decided to create an issue instead. I also checked for duplicate issues and couldn't find any, so I went ahead with opening this one.

### How will this change help?

This feature will significantly improve the user experience when working with the AsyncAPI CLI by:

- **Making the CLI more discoverable** - users can press tab to see what commands are available
- **Reducing cognitive load** - users don't need to memorize all commands and options
- **Speeding up command entry** - especially for long command names or complex option combinations
- **Bringing the CLI in line with industry standards** - most popular CLI tools support tab completion
- **Improving accessibility for new users** - making it easier to explore the tool's capabilities

### How could it be implemented/designed?

As the AsyncAPI CLI uses the oclif framework, adding tab completion is straightforward using the `@oclif/plugin-autocomplete` plugin.

### 🚧 Breaking changes

No

### 👀 Have you checked for similar open issues?

- [x] I checked and didn't find a similar issue

### 🏢 Have you read the Contributing Guidelines?

- [x] I have read the [Contributing Guidelines](https://github.com/asyncapi/.github/blob/master/CONTRIBUTING.md)

### Are you willing to work on this issue?

Yes I am willing to submit a PR!

Shurtu-gal:
This had been on my mind for a while now. Found out about https://github.com/oclif/plugin-autocomplete a while back but didn't get a chance to play around with it.

I think this will be pretty cool to have. WDYT? @Souvikns @AayushSaini101

gambhirsharma:
@AayushSaini101 Should I start working on the PR?

AayushSaini101:
I will take this issue for bounty

Devnil434:
Can I start my work on this issue

aeworxet:
### Bounty Issue's service comment

Text labels: `bounty/2025-Q2`, `bounty/medium`, `bounty/coding`
First assignment to regular contributors: [2025-03-21 00:00:00 UTC+12:00](https://time.lol/#2025-03-21T00:00:00+12:00)
[...]

PR #1724, opened 2025-03-14
feat: add autocomplete feature in cli

AayushSaini101:
Proof of working: 
Screenshot 2025-03-14 at 8 29 06 AM
Relates to: #1685


AayushSaini101:
> @AayushSaini101 I don't think this is ths way autocomplete should be done can we run it as part of installation process?

Sure. Let me check during the installation process

Souvikns:
Awesome feature in my opinion, I am always forgetting command 😆

AayushSaini101:
1 test case is failing, checking the cause
Here is the outcome of adding script and a new plugin for autocomplete
Screenshot 2025-03-18 at 11 25 07 AM

package.json
```diff
@@ -163,6 +165,7 @@
     "pretest": "npm run build",
     "pretest:coverage": "npm run build",
     "posttest": "rimraf ./test.asyncapi-cli",
+    "postinstall": "node ./scripts/enableAutoComplete.js",
```
Shurtu-gal, line 168: Does this script runs after packaging as well?
Also what about docker image. Does this still work there?
Same question for chocolatey.

scripts/enableAutoComplete.js
```diff
@@ -0,0 +1,102 @@
+/* eslint-disable @typescript-eslint/no-var-requires */
+
+const { spawnSync } = require('child_process');
+const os = require('os');
+const fs = require('fs');
+const path = require('path');
+
+const shellConfigs = {
```
Shurtu-gal, line 8: What about their derivatives?

scripts/enableAutoComplete.js
```diff
@@ -0,0 +1,102 @@
...
+  },
+  bash: {
+    rcFile: path.join(os.homedir(), '.bashrc'),
+    detectFile: path.join(os.homedir(), '.bashrc'),
+    postMessage: 'Run: source ~/.bashrc',
+    action: (output, rcFile) => fs.appendFileSync(rcFile, `\n# AsyncAPI CLI Autocomplete\n${output}\n`),
+  },
+  fish: {
+    rcFile: path.join(os.homedir(), '.config', 'fish', 'completions', 'asyncapi.fish'),
+    detectFile: path.join(os.homedir(), '.config', 'fish', 'config.fish'),
```
Shurtu-gal, line 23: Detection using $SHELL or some other way would be better. Not everyone has their config like this.

Shurtu-gal (changes requested):

scripts/enableAutoComplete.js
AayushSaini101: After running the npm install 
Here is the result 
image
The process.env.SHELL returning /bin/sh the because many system use by default to run these kind of scripts. 

but if we use the same concept as in the PR, we can easily detect the default shell of the os. Result
image

Shurtu-gal:
Fish: https://github.com/oclif/plugin-autocomplete/pull/38

Shurtu-gal:
@AayushSaini101 You can take a look at this comment, this solution should work for all shell I believe.

https://github.com/oclif/plugin-autocomplete/pull/38#issuecomment-2761570609

https://github.com/carapace-sh/carapace

AayushSaini101:
> @AayushSaini101 You can take a look at this comment, this solution should work for all shell I believe.
> 
> [oclif/plugin-autocomplete#38 (comment)](https://github.com/oclif/plugin-autocomplete/pull/38#issuecomment-2761570609)
> 
> https://github.com/carapace-sh/carapace

@Shurtu-gal yes this can work, but to automate all the thing is very complex and with some small changes in the cli commands we need to do changes very much, in this approach, we need to use external binary that is carapace-spec, when i try to download first time, my mac doesn't allow to open directly i cannot able to access it, then i need to change the settings from privacy section of the mac **manually**  then need to configure in the path.  If you check this image, we are passing hardcore commands, flags and arguments to the carapace-spec
image
This is the yaml document containing the structure: 
image

The dependency is very much in the case of second plugin, my take. 
* Automate zshrc and bash using the first plugin 
* Add the instructions to manually enable autocomplete feature in rest of the other shells.

AayushSaini101:
@Shurtu-gal In order to maintain the UX properly, allowing autocomplete for [.bashrc', '.bash_profile', '.profile',.zshrc'] shells

scripts/enableAutoComplete.js
```diff
@@ -0,0 +1,143 @@
...
+      console.warn(`⚠️ Ignored error while checking path ${potentialPath}: ${error.message}`);
+    }
+  }
+
+  throw new Error('CLI executable not found. Ensure AsyncAPI CLI is installed.');
+}
+
+function generateAutocompleteScript(shell) {
+  const executablePath = findCliExecutable();
+  const result = spawnSync(executablePath, ['autocomplete', 'script', shell], {
```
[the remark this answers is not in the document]
AayushSaini101: Done thanks

docs/installation.md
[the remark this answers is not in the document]
AayushSaini101: Done thanks

Souvikns (approved):

docs/installation.md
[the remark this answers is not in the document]
Shurtu-gal: Could you add the changes to the installation as well. Mainly revolving around the fact that we are adding stuff to their profile. This becomes something like a privacy thing so adding a log message in the enable.js file as well as mention it in the docs would be the best.

docs/autocompleteEnabled.md
```diff
@@ -0,0 +1,77 @@
+---
+title: 'AutoCompleteEnabled Guide'
```
Shurtu-gal, line 2:
```suggestion
title: 'Auto-complete setup'
```

docs/installation.md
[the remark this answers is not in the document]
AayushSaini101: I will update the docs

Amzani (dismissed):
Nice PR, LGTM

Shurtu-gal (approved):
Thanks for the requested changes.

.changeset/breezy-pandas-arrive.md
```diff
@@ -0,0 +1,5 @@
+---
+"@asyncapi/cli": minor
+---
+
+Remove `--renderer` flag and make `React` as the de-facto renderer, deprecating `Nunjucks`
```
Shurtu-gal, line 5: Please change the changelog

Shurtu-gal (changes requested):

aeworxet:
@asyncapi/bounty_team

PR #1768, opened 2025-04-14
fix(docs): update link to preview.ts in usage documentation

neoandmatrix:
This PR fixes the docs link for the new command `asyncapi start preview`. The broken link happened as the command files and docs were added in the same PR hence the correct link to command reference in docs can only be added after the command is added by the merging of that PR.

Related issues:-
None

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -785,7 +785,7 @@ DESCRIPTION
   starts a new local instance of Studio in minimal state bundling all the refs of the schema file and with no editing allowed. 
 ```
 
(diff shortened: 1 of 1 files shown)


docs/usage.md
Shurtu-gal, line 1: As far as I am aware this is updated automatically, so your changes will be overwritten.

Shurtu-gal (changes requested):

docs/usage.md
neoandmatrix: Currently the link in docs doesn't work. The one on the asyncapi.com.

PR #1803, opened 2025-06-15
docs: added new generator command

### LINKED ISSUE #1741: [📑 Docs]: Improvement for the usage of `--use-new-generator` flag.
### What Dev Docs changes are you proposing?

Currently, 2 generator versions are being maintained under AsyncAPI as mentioned in the https://github.com/asyncapi/cli/issues/1388 . But currently, the CLI docs lack a clear mention on when to use hence creating confusion. Hence the docs need to be updated properly to differentiate that when this flag is to be used or not.

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

mahimonga:
Resolves #1741


neoandmatrix:
Hey @mahimonga is this PR ready for review? As If it is I don't think this was the expected change with which I had opened the issue which this is PR resolves. 

Currently as we have 2 generator versions being maintained the documentation of cli needs to be updated for the `generate` command section to reflect and refer the difference in two versions of generator and when to use which with different templates to avoid confusion. 

Most of this is also documented in the dedicated generator docs only on `cli` side some clarity is required.

But if it is still in progress please mark it in draft so as to avoid confusion.

Thanks.

mahimonga:
\u

AayushSaini101:
LGTM thanks @mahimonga

Shurtu-gal:
@AayushSaini101 changes to `usage.md` is not done manually but instead done through automation. These changes must be made in code for now else they would be overridden.

Issue #1796, opened 2025-06-10
[FEATURE] Pass authentication/authorization to $ref HTTP resolvers


fmvilas:
### Why do we need this improvement?

It is impossible to work with files that have `$ref` pointing to a URL that's not public. In large organizations, it's pretty common to rely on schema definitions in a Schema Registry server and a GitHub repo with common definitions. However, they usually require some sort of authentication. In most cases, a simple way to specify an HTTP `authorization` header should suffice.

### How will this change help?

By providing a mechanism in the CLI to pass an HTTP `authorization` header, we'll be unblocking the users in large organizations, which in turn are AsyncAPI target users (AsyncAPI doesn't really make much sense in small companies).

### How could it be implemented/designed?

This improvement could be done in the form of a global config. Since secrets are often stored in environment variables, it would be great to add support for specifying which env var to read from.

Ideas:

```
# This makes all $refs pointing to any file in the myorg/myrepo repo to use the
# HTTP header Authorization: Bearer super-secret-here
asyncapi config auth add github.com/myorg/myrepo/**/*.* bearer super-secret-here

# This makes all $refs pointing to any file in the myorg/myrepo repo to use the
# HTTP header Authorization: Bearer {{$MY_TOKEN}}
asyncapi config auth add github.com/myorg/myrepo/**/*.* bearer $MY_TOKEN --env
```

### 🚧 Breaking changes

No

### 👀 Have you checked for similar open issues?

- [x] I checked and didn't find a similar issue

### 🏢 Have you read the Contributing Guidelines?

- [x] I have read the [Contributing Guidelines](https://github.com/asyncapi/.github/blob/master/CONTRIBUTING.md)

### Are you willing to work on this issue?

None

fmvilas:
This requires a PR in the github.com/asyncapi/parser-js too, so we can pass this info to the resolvers. It may even requires us to build a custom HTTP resolver.

AayushSaini101:
@fmvilas Sounds interesting would like to take

aeworxet:
The maintainers of https://github.com/asyncapi/parser-js have been inactive for half a year already; any PR there would be stalled indefinitely, and this Bounty Issue will be impossible to complete due to this trivial matter.

First, an active maintainer needs to be appointed there.
Second, this maintainer needs to fix the testing (I and @derberg explored it and can provide information on how to fix it).
Only then will there be sense in submitting new PRs there.

[...]

PR #1810, opened 2025-06-21
feat:  add new custom resolver to fetch the reference from private repo

AayushSaini101:
Resolves: #1796 

* Create a custom resolver to fetch the reference from the private repo 
* Add new command auth add for storing the repo name and repo token. 
* Tests and documents need to be updated

### DIFF
--- a/.changeset/1810.md
+++ b/.changeset/1810.md
@@ -52,5 +52,30 @@ Co-authored-by: Fran Méndez <fmvilas@gmail.com>
 Co-authored-by: Fran Méndez <fmvilas@gmail.com>
 - 8943766: updateValidationservice
 - aa37d14: remove invalid comment
+- 3b882ed: Add new custom resolver to fetch the reference from private template and add new command auth add
+
+# Conflicts:
+#	package-lock.json
+#	src/core/parser.ts
+- 811bc08: update
+
+# Conflicts:
+#	asyncapiTest.yml
+#	index.js
+#	src/apps/api/middlewares/validation.middleware.ts
+#	src/apps/cli/commands/diff.ts
+#	src/apps/cli/commands/generate/models.ts
+#	src/apps/cli/commands/validate.ts
+#	src/domains/services/config.service.ts
+#	src/domains/services/validation.service.ts
+- bb0fb6c: remove config service test file
+- 32298a9: remove invalid statements
+- ad3b3b8: add missing dependencies
+- 7662fca: remove warnings
+- 40b4b50: update validation-service
+- d7db432: remove invalid file
+- 98dfd4c: fix lint issue
+- 75dda7b: update validation-service
+- 5c23530: fix lint issue
 
 


AayushSaini101:
Tests and documentation need to be updated, i will update after getting ack from this approach : ) 
* Spec containing private repo: 
image


* Trying to parse using the existing version of the asyncapi cli: 
image


### New feature: 

* Add auth value for the github token: 
image


**Repo URL format**: https://github.com/AayushSaini101/AayushSaini101/tree/main

* Trying to fetch the private ref: 

image

cc: @fmvilas @Souvikns @Shurtu-gal

src/core/parser.ts
```diff
@@ -35,13 +36,73 @@ const formatExtensions: Record<DiagnosticsFormat, string> = {
...
+
+  const [org, repo, branch] = segments;
+  return `https://github.com/${org}/${repo}/tree/${branch}`;
+}
+
+const customResolver = {
+  schema: 'https',
+  order: 1,
+
+  canRead: (uri: URI) => uri.hostname() === 'raw.githubusercontent.com',
```
fmvilas, line 71: Why checking for GitHub here? 🤔 References can be in many places, not just GitHub, and even in such a case the GitHub deployment might be in a private domain, which is pretty common in medium to large companies.

src/core/parser.ts
```diff
@@ -35,13 +36,73 @@ const formatExtensions: Record<DiagnosticsFormat, string> = {
...
+
+  read: async (uri: URI) => {
+    const url = uri.toString();
+    const normalized = normalizeRawGitUrl(uri);
```
fmvilas, line 75: What if the URL is not pointing to a GitHub repo? This could perfectly be a Schema Registry URL, an S3 bucket, some random URL, or even a GitLab or BitBucket URL.

src/core/parser.ts
```diff
@@ -35,13 +36,73 @@ const formatExtensions: Record<DiagnosticsFormat, string> = {
...
+    const matched = entries.find(entry =>
+      micromatch.isMatch(normalized, entry.pattern)
+    );
+
+    const headers: Record<string, string> = {};
+    if (matched) {
+      const token = matched.token.startsWith('$')
+        ? process.env[matched.token.slice(1)] || ''
+        : matched.token;
+      headers['Authorization'] = `Bearer ${token}`;
```
fmvilas, line 86: Some systems don't use Bearer tokens. Some might accept `jwt` or `api-key` or actually whatever the API designer wants to put there. We shouldn't hardcode it, that's why in my definition of the task I added some alternative examples.

fmvilas (changes requested):

src/core/parser.ts
AayushSaini101: sure @fmvilas i will update it

AayushSaini101:
@fmvilas For the private Github URL should be in this format: like https://api.github.com/repos/AayushSaini101/AayushSaini101/contents/user-signedup.yaml, not direct URL like we simple copy the path of the URL from the private repo, this is the github url, not sure about the private repo, in short we need to compare each and every url and create special check to find whether it is matching with the stored pattern or not, we can start with the github and gitlab pattern then we can slowly slowly add the comparison format as we grow thanks

aeworxet:
@asyncapi/bounty_team

fmvilas:
> @fmvilas For the private Github URL should be in this format: like https://api.github.com/repos/AayushSaini101/AayushSaini101/contents/user-signedup.yaml, not direct URL like we simple copy the path of the URL from the private repo, this is the github url, not sure about the private repo, in short we need to compare each and every url and create special check to find whether it is matching with the stored pattern or not, we can start with the github and gitlab pattern then we can slowly slowly add the comparison format as we grow thanks

@AayushSaini101 I'm not sure I follow. I think you can translate from one to another, right? For instance, if I give you https://github.com/AayushSaini101/a-repo/blob/master/asyncapi.yaml (which is the URL I can copy from the browser), you can transform it to https://api.github.com/repos/AayushSaini101/a-repo/contents/asyncapi.yaml or whatever format you need it to be, right? The idea is to make it really easy for the user.

Likewise, if my company has a Github installation behind the https://source.mycompany.com URL, then a URL like https://source.mycompany.com/AayushSaini101/a-repo/blob/master/asyncapi.yaml should also be possible to resolve.

In short, picture this, you're working at a company and have an AsyncAPI or JSON/YAML file in a private repo. You want to reference it using $ref and it should just work out of the box. Sure, the user has to provide the necessary details to resolve these URLs but it should be fairly easy once it's configured.

AayushSaini101:
> [github.com/repos/AayushSaini101/a-repo/contents/asyncapi.yaml](https://api.github.com/repos/AayushSaini101/a-repo/contents/asyncapi.yaml)

yes, i can tranform/configured the URL for the Github because i know the format for the accepted URL in the case of the Github, but for the case of another services like registry or any private registry what should be the correct URL that we need to consider for the custom resolver, that's why for this issue we can enable feature Github then in future according to the requirements we can add bit by bit @fmvilas

AayushSaini101:
> > @fmvilas For the private Github URL should be in this format: like https://api.github.com/repos/AayushSaini101/AayushSaini101/contents/user-signedup.yaml, not direct URL like we simple copy the path of the URL from the private repo, this is the github url, not sure about the private repo, in short we need to compare each and every url and create special check to find whether it is matching with the stored pattern or not, we can start with the github and gitlab pattern then we can slowly slowly add the comparison format as we grow thanks
> 
> @AayushSaini101 I'm not sure I follow. I think you can translate from one to another, right? For instance, if I give you https://github.com/AayushSaini101/a-repo/blob/master/asyncapi.yaml (which is the URL I can copy from the browser), you can transform it to https://api.github.com/repos/AayushSaini101/a-repo/contents/asyncapi.yaml or whatever format you need it to be, right? The idea is to make it really easy for the user.
> 
> Likewise, if my company has a Github installation behind the https://source.mycompany.com URL, then a URL like https://source.mycompany.com/AayushSaini101/a-repo/blob/master/asyncapi.yaml should also be possible to resolve.
> 
> In short, picture this, you're working at a company and have an AsyncAPI or JSON/YAML file in a private repo. You want to reference it using $ref and it should just work out of the box. Sure, the user has to provide the necessary details to resolve these URLs but it should be fairly easy once it's configured.

@fmvilas User need to provide the direct valid url in the asyncapi spec that it  https://api.github.com/repos/AayushSaini101/a-repo/contents/asyncapi.yaml , so that it would be easy to direct authenticate without chaning in any other kind of format. wdyt ?

AayushSaini101:
> > > @fmvilas For the private Github URL should be in this format: like https://api.github.com/repos/AayushSaini101/AayushSaini101/contents/user-signedup.yaml, not direct URL like we simple copy the path of the URL from the private repo, this is the github url, not sure about the private repo, in short we need to compare each and every url and create special check to find whether it is matching with the stored pattern or not, we can start with the github and gitlab pattern then we can slowly slowly add the comparison format as we grow thanks
> > 
> > 
> > @AayushSaini101 I'm not sure I follow. I think you can translate from one to another, right? For instance, if I give you https://github.com/AayushSaini101/a-repo/blob/master/asyncapi.yaml (which is the URL I can copy from the browser), you can transform it to https://api.github.com/repos/AayushSaini101/a-repo/contents/asyncapi.yaml or whatever format you need it to be, right? The idea is to make it really easy for the user.
> > Likewise, if my company has a Github installation behind the https://source.mycompany.com URL, then a URL like https://source.mycompany.com/AayushSaini101/a-repo/blob/master/asyncapi.yaml should also be possible to resolve.
> > In short, picture this, you're working at a company and have an AsyncAPI or JSON/YAML file in a private repo. You want to reference it using $ref and it should just work out of the box. Sure, the user has to provide the necessary details to resolve these URLs but it should be fairly easy once it's configured.
> 
> @fmvilas User need to provide the direct valid url in the asyncapi spec that it https://api.github.com/repos/AayushSaini101/a-repo/contents/asyncapi.yaml , so that it would be easy to direct authenticate without chaning in any other kind of format. wdyt ?

@fmvilas can you check the last question, i guess then after we can resume on this thanks a lot : )

fmvilas:
> @fmvilas User need to provide the direct valid url in the asyncapi spec that it https://api.github.com/repos/AayushSaini101/a-repo/contents/asyncapi.yaml , so that it would be easy to direct authenticate without chaning in any other kind of format. wdyt ?

Honestly, I think that would be good enough but since we're still here designing the feature, enabling the user to simply put URLs like https://github.com/AayushSaini101/a-repo/blob/master/asyncapi.yaml (which is the URL they can copy from the browser) would produce a higher DX. That's not to say the user can't put the https://api.github.com/repos/AayushSaini101/a-repo/contents/asyncapi.yaml URL. An amazing DX is one that anticipates this kind of things IMHO. And it's really a low effort thing.

AayushSaini101:
> > @fmvilas User need to provide the direct valid url in the asyncapi spec that it https://api.github.com/repos/AayushSaini101/a-repo/contents/asyncapi.yaml , so that it would be easy to direct authenticate without chaning in any other kind of format. wdyt ?
> 
> Honestly, I think that would be good enough but since we're still here designing the feature, enabling the user to simply put URLs like https://github.com/AayushSaini101/a-repo/blob/master/asyncapi.yaml (which is the URL they can copy from the browser) would produce a higher DX. That's not to say the user can't put the https://api.github.com/repos/AayushSaini101/a-repo/contents/asyncapi.yaml URL. An amazing DX is one that anticipates this kind of things IMHO. And it's really a low effort thing.

Thanks for the suggestions @fmvilas, from now i am resuming the pending work on this PR that is allowing private pull from the Github Repo : )

src/apps/cli/commands/validate.ts
```diff
@@ -16,7 +16,7 @@ import {
 
 export default class Validate extends Command {
   static description = 'validate asyncapi file';
-  private validationService = new ValidationService();
+  private validationService = new ValidationService({}, true);
```
fmvilas, line 19: This looks really confusing. What's that empty object and that `true` value? What do they mean?

My suggestion is that you merge the two. If you still want to keep parserOptions separated, in the ValidationService, you can do something like:

```js
{
  parserOptions: ...,
  customAuth: true
}
```

This way it'll be much clearer what these things mean.

Now that said, is there any case we should not be using the customAuth in the CLI? Shouldn't it be always enabled? Also, the parserOptions object already have a way to provide the custom resolver, why do we need another parameter? 🤔

src/domains/services/validation.service.ts
```diff
@@ -59,16 +62,76 @@ const validFormats = [
   'pretty',
 ];
 
+type AuthEntry = {
+  pattern: string;
+  token: string;
+};
+
+const CONFIG_FILE = path.join(os.homedir(), '.asyncapi', 'config.json');
```
fmvilas, line 70: Can't we import this from somewhere else? I don't think it's the responsibility of the ValidationService to know how to construct the config directory.

fmvilas (changes requested):
Left some suggestions. I'd also add that this custom resolver should work with all the commands, not just when doing the validation.

src/domains/services/validation.service.ts
AayushSaini101: Make sense i will do : )

src/core/parser.ts
fmvilas: Just to be sure we're understanding each other. If I'm understanding it correctly, right now we're assuming all URLs are GitHub URLs, which is wrong because that would make parser stop working with custom URLs. In the `normalizeRawGitUrl` function we should be checking if it's actually a GitHub URL, otherwise, leave it intact. Mappings for GitLab and Bitbucket can wait but breaking all URLs except the GitHub ones is definitely not acceptable.

AayushSaini101: yes, i need to change the functionnaming properly i will make it generic so that we can normalise other url as well in the future based on the requirements

AayushSaini101:
Summary for all the conversations 

* For every private URL that the user passes in the Asyncapi document has to be normalised in the particular format in order to fetch the information properly, in order to maintain the better DX we will allow the user to pass the default URL 

For example in the case of Github Private URL we have to convert into normalised URL: 
Spec private URL: 'https://github.com/AayushSaini101/AayushSaini101/blob/main/user-signedup.yaml#/payload 
Normalised URL:  `https://api.github.com/repos/${owner}/${repo}/contents/${path}?ref=${branch}`; 

The function should be generic so that in the future, we can extend it properly 

* No need to use another parameter for the customAuth 
* The format of the secerts: 
  _asyncapi config auth add github.com/myorg/myrepo/**/*.* bearer super-secret-here_
* All the secrets should be stored in the file so that we can check whether the secrete is present in the file or not 


This is all the final requiments : cc: @fmvilas

fmvilas:
> For every private URL that the user passes in the Asyncapi document has to be normalised in the particular format in order to fetch the information properly, in order to maintain the better DX we will allow the user to pass the default URL

This only applies for GitHub URLs. For all other URLs, private or not, the URL shouldn't change.

src/domains/services/validation.service.ts
AayushSaini101: Created a new service thanks : )

AayushSaini101:
@fmvilas completed the implementation and tested also with cc: @neoandmatrix

Shriya-Chauhan:
@AayushSaini101 Tested it. Working fine.

neoandmatrix:
> @fmvilas completed the implementation and tested also with cc: @neoandmatrix 
> 

Yes @fmvilas tested this and it's working fine.

src/apps/cli/commands/diff.ts
```diff
@@ -27,7 +27,7 @@ const { readFile } = fs;
 
 export default class Diff extends Command {
   static description = 'Find diff between two asyncapi files';
-  private validationService = new ValidationService();
+  private validationService = new ValidationService({});
```
fmvilas, line 30: This doesn't seem necessary anymore, right?

src/domains/services/config.service.ts
```diff
@@ -0,0 +1,123 @@
...
+    if (!config.auth) {
+      config.auth = [];
+    }
+
+    // Find if an entry with the same pattern already exists
+    const index = config.auth.findIndex(e => e.pattern === entry.pattern);
+
+    if (index !== -1) {
+    // If found, merge existing entry with new entry
+      // eslint-disable-next-line security/detect-object-injection
```
fmvilas, line 63: Is there a good reason to disable security/detect-object-injection? We're dealing with auth and tokens, shouldn't we be careful here?

src/domains/services/validation.service.ts
```diff
@@ -32,6 +32,108 @@ import { ParseOptions } from '@asyncapi/parser';
...
+  /**
+   * Convert GitHub web URL to API URL
+   */
+  convertGitHubWebUrl: (url: string): string => {
+    // Remove fragment from URL before processing
+    const urlWithoutFragment = url.split('#')[0];
+    
+    // Handle GitHub web URLs like: https://github.com/owner/repo/blob/branch/path
+    // eslint-disable-next-line no-useless-escape
+    const githubWebPattern = /^https:\/\/github\.com\/([^\/]+)\/([^\/]+)\/blob\/([^\/]+)\/(.+)$/;
```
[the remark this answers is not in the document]
AayushSaini101: Thanks : )

AayushSaini101:
> I think you should, at least, test that the resolver is working well for some GitHub URLs and also that the resolver is not interfering with other non-GitHub URLs.

Added some test cases thanks ; )

fmvilas:
@AayushSaini101 let me know if you need guidance.

AayushSaini101:
> @AayushSaini101 let me know if you need guidance.

@fmvilas tried to added some test cases thanks: : )

Shurtu-gal (approved):
The code looks fine to me. Just wanted to know the use cases for this (is it just limited to github private repo).

Also @fmvilas what should be the proposed way of testing the auth?

test/unit/services/config.service.test.ts
```diff
@@ -0,0 +1,88 @@
+import { expect } from 'chai';
```
fmvilas, line 1: This file is definitely not needed. It's not testing any part of our code.

test/unit/services/validation.service.test.ts
```diff
@@ -129,4 +172,80 @@ describe('ValidationService', () => {
...
+      } else {
+        try {
+          await fs.unlink(CONFIG_FILE);
+        } catch (err) {
+          // File doesn't exist, ignore
+        }
+      }
+    });
+
+    it('should validate document with private GitHub reference', async () => {
```
fmvilas, line 204: This is testing that the document fails to validate because the $ref is in a private URL and we haven't configured how to resolve it. IMHO, the name of the test is not reflecting what this is testing.

fmvilas: I suggest something like the following:

```suggestion
    it('should fail to validate document with private GitHub reference when not properly configured', async () => {
```

test/unit/services/validation.service.test.ts
```diff
@@ -129,4 +172,80 @@ describe('ValidationService', () => {
...
+
+      const specFile = new Specification(asyncAPIWithPrivateGitHubRef);
+      const options = {
+        'diagnostics-format': 'stylish' as const
+      };
+
+      const result = await validationService.validateDocument(specFile, options);
+
+      // The validation succeeds but the document is invalid due to unresolved ref
+      expect(result.success).to.equal(true);
```
fmvilas, line 220: If this test is passing we're doing something wrong. `result.success` shouldn't be `true` because a key part of the document couldn't even be resolved.

test/unit/services/validation.service.test.ts
```diff
@@ -129,4 +172,80 @@ describe('ValidationService', () => {
...
+      if (result.success) {
+        expect(result.data).to.have.property('status');
+        expect(result.data?.status).to.equal('invalid');
```
fmvilas, line 223: Having `result.success === true` and `result.data?.status === 'invalid'` is definitely weird. Don't you think so?

fmvilas (changes requested):
Left comment on the tests.

IMHO, you should mock the parser functionality that's trying to make an HTTP request. When you're testing the private URL, it should fake a 401 Unauthorized HTTP response. Perhaps a 404 Not found response too, since some servers will give you that a 404 instead of a 401.

The way tests are right now, they're not testing the new feature. Let me know if you need guidance for testing.

fmvilas:
> The code looks fine to me. Just wanted to know the use cases for this (is it just limited to github private repo).

To clarify, this is not just for GitHub private repos. The URL translation feature is just for GitHub though.
 
> Also @fmvilas what should be the proposed way of testing the auth?

We should mock the resolver and fake the status code and response we get. We should be testing that unless we pass the `Authorization` header with a specific value (a random value), the resolver will respond with 401 or 404. Basically, imitate what a real server would do without making the actual HTTP call.

AayushSaini101:
> > The code looks fine to me. Just wanted to know the use cases for this (is it just limited to github private repo).
> 
> To clarify, this is not just for GitHub private repos. The URL translation feature is just for GitHub though.
> 
> > Also @fmvilas what should be the proposed way of testing the auth?
> 
> We should mock the resolver and fake the status code and response we get. We should be testing that unless we pass the `Authorization` header with a specific value (a random value), the resolver will respond with 401 or 404. Basically, imitate what a real server would do without making the actual HTTP call.

Thanks for the clarification @fmvilas I thought we just needed to add for the GitHub support only, it can extend other private URL as well: 
```
} else {
      if (authInfo) {
        headers['Authorization'] = `${authInfo.authType} ${authInfo.token}`;
        Object.assign(headers, authInfo.headers); // merge custom headers
      }
      const res = await fetch(url, { headers });
      if (!res.ok) {
        throw new Error(`Failed to fetch URL: ${url} - ${res.statusText}`);
      }
      return await res.text();
    }
```

src/domains/services/validation.service.ts
```diff
@@ -32,6 +32,120 @@ import { ParseOptions } from '@asyncapi/parser';
...
+    let url = uri.toString();
+    
+    // Default headers
+    const headers: Record<string, string> = {
+      'User-Agent': 'AsyncAPI-CLI'
+    };
+
+    const authInfo = await ConfigService.getAuthForUrl(url);
+
+    if (url.includes('github.com') && url.includes('/blob/')) {
```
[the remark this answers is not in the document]
AayushSaini101: Added thanks

test/unit/services/validation.service.test.ts
fmvilas: Don't you find it weird that `success === true` but the document is invalid?

src/domains/services/validation.service.ts
```diff
@@ -32,6 +32,154 @@ import { ParseOptions } from '@asyncapi/parser';
...
+    // Default headers
+    const headers: Record<string, string> = {
+      'User-Agent': 'AsyncAPI-CLI'
+    };
+
+    const authInfo = await ConfigService.getAuthForUrl(url);
+
+    if (isValidGitHubBlobUrl(url)) {
+      url = createHttpWithAuthResolver().convertGitHubWebUrl(url);
+    }
```
fmvilas, line 134: You are creating a resolver just to call the `convertGitHubWebUrl` function. Extract that function instead.

```suggestion
    if (isValidGitHubBlobUrl(url)) {
      url = convertGitHubWebUrl(url);
    }
```

src/domains/services/validation.service.ts
```diff
@@ -32,6 +32,154 @@ import { ParseOptions } from '@asyncapi/parser';
...
+    
+    // Only require authentication for GitHub URLs
```
fmvilas, line 136: This comment is not true. We should not require authentication **only** for GitHub URLs.

src/domains/services/validation.service.ts
```diff
@@ -32,6 +32,154 @@ import { ParseOptions } from '@asyncapi/parser';
...
+    if (isValidGitHubUrl(url)) {
+      if (authInfo) {
+        headers['Authorization'] = `${authInfo.authType} ${authInfo.token}`;
+        Object.assign(headers, authInfo.headers); // merge custom headers
+      } else {
+        throw new Error(`Cannot resolve URL: ${url} - No authentication configured for this GitHub repository`);
+      }
+    }
```
fmvilas, line 144: You don't need this because we're not applying authorization just on GitHub URLs. We apply authorization if we find the URL in the config. That's it.


```suggestion
    if (authInfo) {
      headers['Authorization'] = `${authInfo.authType} ${authInfo.token}`;
      Object.assign(headers, authInfo.headers); // merge custom headers
    }
```

src/domains/services/validation.service.ts
```diff
@@ -32,6 +32,154 @@ import { ParseOptions } from '@asyncapi/parser';
...
+      const res = await fetch(url, { headers });
+      if (!res.ok) {
+        throw new Error(`Failed to fetch GitHub URL: ${url} - ${res.statusText}`);
+      }
+      return await res.text();
+    } else {
+      if (authInfo) {
+        headers['Authorization'] = `${authInfo.authType} ${authInfo.token}`;
+        Object.assign(headers, authInfo.headers); // merge custom headers
+      }
```
fmvilas, line 173: Now you don't need this here because I moved it outside.

```suggestion
```

src/domains/services/validation.service.ts
AayushSaini101: Make sense thanks

src/domains/services/validation.service.ts
```diff
@@ -32,6 +32,120 @@ import { ParseOptions } from '@asyncapi/parser';
...
+      throw new Error(`No download URL found in GitHub API response for: ${url}`);
+    } else if (url.includes('raw.githubusercontent.com')) {
+      headers['Accept'] = 'application/vnd.github.v3.raw';
+      const res = await fetch(url, { headers });
+      if (!res.ok) {
+        throw new Error(`Failed to fetch GitHub URL: ${url} - ${res.statusText}`);
+      }
+      return await res.text();
+    } else {
```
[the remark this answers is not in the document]
AayushSaini101: Resolved thanks : )

AayushSaini101:
Need to open new PR, squashing and cleaning the commits something got wrong : (

Issue #1752, opened 2025-04-03
[FEATURE] Reduce number of dependencies and install size (v2.17 needs 1700 dependencies, takes 1.7Gb of space)


peitschie:
### Why do we need this improvement?

The `@asyncapi/cli` tool (v2.17.0) requires 1700 additional packages and occupies about 1.7Gb of space on install.

This provides a significant barrier to use within more controlled environments, as the auditing and review of the dependencies is completely impractical.

It also has a significant impact on install size, application performance, and even tool choice.

For example, I'm unable to install this on some machines as they simply don't have enough RAM available to complete the installation without special overrides.

### How will this change help?

Reducing the number of dependencies significantly would make this tool a much more attractive option, and likely will reduce the maintenance burden of keeping the dependencies up to date.

It would also increase accessibility to the package.

### Screenshots

See attachment for the current result of running `npm install @asyncapi/cli` in a completely clean directory:
[dependencies.txt](https://github.com/user-attachments/files/19578086/dependencies.txt)

### How could it be implemented/designed?

Some related discussion seems to be in #1657, #1707 and #1703 (these all list identifying and removing redundant dependencies and imports)

1. Review direct dependencies and determine which are not providing significant features and might be suitable to inline/reimplement.
2. Review heavy-weight dependencies to evaluate alternative tools or alternative approaches that don't require the massive dependency tree

### 🚧 Breaking changes

Yes

### 👀 Have you checked for similar open issues?

- [x] I checked and didn't find a similar issue

### 🏢 Have you read the Contributing Guidelines?

- [x] I have read the [Contributing Guidelines](https://github.com/asyncapi/.github/blob/master/CONTRIBUTING.md)

### Are you willing to work on this issue?

Yes I am willing to submit a PR!

neoandmatrix:
Hey @Shurtu-gal was investigating this one and as per above comment comparing the changes between the versions mentioned above the i.e 2.16.7 and 2.16.8 the studio's new version which uses next.js was implemented, hence i believe a seperate issue there needs to open for discussion on reducing size if possible, WDYT or something else can also be done?

Shurtu-gal:
Could you generate a list dependency sizes @neoandmatrix

neoandmatrix:
> Could you generate a list dependency sizes @neoandmatrix 

Okay, will update.

neoandmatrix:
Main root
[...]

Issue #1693, opened 2025-03-05
[FEATURE] Addition of tests for the commands that depend on browser interaction.


neoandmatrix:
### Why do we need this improvement?

Currently none of the commands like the one related to `studio` can be tested properly as there is a lack of mechanism to test these command that open a browser.

### How will this change help?

This change will allow us to properly test the commands like `asyncapi start studio` and another commands like `asyncapi new` which require interaction with browser.

### How could it be implemented/designed?

As this is browser interaction related we can use libraries like `Puppeteer` but first discussions with maintainers are required to figure out a correct approach.

### 🚧 Breaking changes

No

### 👀 Have you checked for similar open issues?

- [x] I checked and didn't find a similar issue

### 🏢 Have you read the Contributing Guidelines?

- [x] I have read the [Contributing Guidelines](https://github.com/asyncapi/.github/blob/master/CONTRIBUTING.md)

### Are you willing to work on this issue?

Yes I am willing to submit a PR!

AayushSaini101:
Good Enhancement for the Gsoc program, we can take during that time

Shurtu-gal:
Thanks @neoandmatrix

AayushSaini101:
@neoandmatrix can you define the scope of this issue ?

neoandmatrix:
Hey @Shurtu-gal @AayushSaini101  did some research on this and here is what i believe can be covered in this :- 

First as we need to increase test coverage with this one hence the testing needs to be integrated with current tests only. Now current tests use `mocha` so these tests will be needed to be integrated with that only.

Second i believe the following can be tested currently :- 
- When we test studio with `start studio` command the presence of elements such as `sidebar` , `editor` etc. can be checked to confirm that correct mode is opened.
- When we test studio with `start preview` command absence of elements such as `editor` and presence of `preview` panel can be tested to confirm that correct mode is opened.

One missing thing is testing websockets as they might be bit tricky to test. First to test websocket one would need to check that if files are modified correctly or not which involves reading from files and also comparing them and testing them both way might be bit difficult and in real-time during tests so these can be added later or after above are implemented in my opinion.

Another thing is what framework to use for this. Currently i explored [cypress](https://www.cypress.io/) , [puppeteer](https://pptr.dev/) and [playwright](https://playwright.dev/) out of these I found **cypress** to be best supported with `mocha` and `chai` currently.

Puppeteer also have an [example of usage](https://pptr.dev/examples#testing) with `mocha` but is quite outdated .

[...]

PR #1815, opened 2025-07-06
fix: made studio start on different port if one is already in use

### LINKED ISSUE #1797: [ENHANCEMENT] Allow starting of other instance of studio if one is already running.

Currently if `asyncapi studio` is running and one tries to start another instance of studio with a different file the command is stuck if different ports are not passed manually and no message is displayed.

Will allow opening of multiple instance of `studio` for different files.

### Screenshots

Image

Allow trying of different ports if not passed manually.

### 🚧 Breaking changes

No

neoandmatrix:
**Description**

The PR adds the functionality to automatically start a new instance of studio on a different port if the current port intended to be used is already in use.

image

An info is displayed that the specified port was already in use hence a different port is being used to the user.

**Related issue(s)**
Resolves #1797


aeworxet:
@asyncapi/bounty_team

neoandmatrix:
Hey @AayushSaini101 @Shurtu-gal if possible can this be reviewed and merged as #1824 is dependent on this.

Thanks

neoandmatrix:
Hey @AayushSaini101 @Shurtu-gal any update on this ?

Shurtu-gal:
I guess we should not go ahead with this iterative checking of ports. Can use stuff like: https://expressjs.com/en/api.html#app.listen

Basically mentioning port as 0 gets the OS to allocate a port.

neoandmatrix:
> I guess we should not go ahead with this iterative checking of ports. Can use stuff like: https://expressjs.com/en/api.html#app.listen

Is this specific to express.js? I believe we are not using express.

And also it might cause inconsistency as OS will always assign a random port with this method not only for case if port is already in use.

Shurtu-gal:
We are using nodejs http which has the same functionality.

neoandmatrix:
Okay but assigning a random port every time , would it be a good practice and experience ? Also it will make testing harder i believe.

neoandmatrix:
One concern for me with the random port approach is testing these commands through puppeteer. As we need to navigate to the url for testing and since we include port in the url query parameters it being random would make it extremely difficult to navigate that site and test the instance as url would be random every time.

 Inferring it from logs is extremely difficult as it would require a lot of parsing which make tests un-reliable.

Passing the port through flag is also not possible as we can only pass strings in parameter in tests but port require a number which cannot be passed.

neoandmatrix:
Hey @Shurtu-gal with respect to above discussion how should i proceed with this ?

Shurtu-gal:
I suppose we can wait for @AayushSaini101 or @Souvikns input also.

AayushSaini101:
> https://expressjs.com/en/api.html#app.listen

> I guess we should not go ahead with this iterative checking of ports. Can use stuff like: https://expressjs.com/en/api.html#app.listen
> 
> Basically mentioning port as 0 gets the OS to allocate a port.

I guess we can go for iterating port allocation to the server, explicitly allocating the port is not a proper way, wdyt @Souvikns

Shurtu-gal:
> I guess we can go for iterating port allocation to the server

Setting up server and checking each port is not the way to go according to me.

> explicitly allocating the port is not a proper way

I couldn't understand this.

aeworxet:
`npm install portfinder` installs four packages and takes 1.3 MB.
Are these quantities and size a significant concern in this case?

```
$ ls
async
debug
ms
portfinder

$ du -h
32K	./debug/src
68K	./debug
444K	./async/dist
112K	./async/internal
1.1M	./async
20K	./ms
24K	./portfinder/lib
40K	./portfinder
1.3M	.
```

aeworxet:
@Souvikns, do you have insights on https://github.com/asyncapi/cli/pull/1815#issuecomment-3139798032 ?

The package `portfinder`, presumably, can easily resolve this issue, but as I understand it, @Shurtu-gal is not very certain about iterating the ports, and I was addressing https://github.com/asyncapi/cli/issues/1752 with the question of whether the quantity and size of the packages are concerning.

neoandmatrix:
@Shurtu-gal would it be a problem if the custom port passed using the -p flag can be taken as a string instead of number and necessary checks are added that the passed port is within the correct port range of allowed ports on a device and is an postive integer. 

Internally then we convert it to a number and pass it to the studio commands and for other cases we pass 0 to allocate an available port as suggested by you.

As for the UX of the user there would be no changes.

If this is good it would allow me resolve both this issue and #1693.

Shurtu-gal:
Yep @neoandmatrix I would like to back this approach. Let's wait for @AayushSaini101 and @Souvikns thoughts as well.

Shurtu-gal:
> passed port is within the correct port range of allowed ports on a device and is a positive integer.

Can check if the port is being used as well I suppose.

neoandmatrix:
> Can check if the port is being used as well I suppose.

Yes but it would require setting up a temporary server. Would that be fine?

Shurtu-gal:
Won't starting studio also show the `EADDRINUSE` error. We can utilise that I suppose. 

Also I will be resolving the conflicts in evening, a new architecture has been Implemented.

neoandmatrix:
> Won't starting studio also show the EADDRINUSE error. We can utilise that I suppose.

Yes it would but only after server has tried starting up once and it would throw the error in .catch() only from where starting the server i believe is not possible only error message can be displayed to user.

Shurtu-gal:
> Yes it would but only after server has tried starting up once and it would throw the error in .catch() only from where starting the server i believe is not possible only error message can be displayed to user.

If user is explicitly defining the port only error message should suffice.

neoandmatrix:
Sure then it can be done.

neoandmatrix:
@AayushSaini101 @Souvikns can you please have a look at the new suggested approach for resolution of this issue and let me know if i should proceed with the changes for same or not?

Thanks.

neoandmatrix:
@Shurtu-gal let me update this PR with the new approach and then we can discuss from there on as this has been unfortunately delayed too much.

neoandmatrix:
Hey @Shurtu-gal i have updated the PR with new approach and everything seems to be working fine and as expected. Only this linter in testing is failing. Can you please have a look at the new changes?

Thanks.

aeworxet:
When the code is working not only fine but also as expected, it's a relief for the developer.

src/apps/cli/commands/start/preview.ts
```diff
@@ -23,7 +23,9 @@ export default class PreviewStudio extends Command {
 
     let filePath: string | undefined = args['spec-file'] ?? flags.file;
 
-    const previewPort = flags.port;
+    const previewPort = parseInt(flags.port ?? '0',10);
+
+    // to add : add check here
```
Shurtu-gal, line 28: Anything else you wish to add here?

Shurtu-gal:
> Only this linter in testing is failing. Can you please have a look at the new changes?

The changes are looking fine by me. I believe you would be integrating these changes into the tests as well?

Regarding linter: fixing this https://github.com/asyncapi/cli/blob/7817c6d672679f26096d0a99f48b5a7015ceee1c/src/utils/temp-dir.ts#L13-L16 should work.

neoandmatrix:
> The changes are looking fine by me. I believe you would be integrating these changes into the tests as well?

In studio testing yes these will be added but not random port allocation as that will be done by OS so I don't think that's needed to be tested explicitly or separately.

neoandmatrix:
Hey @AayushSaini101 @Shurtu-gal have resolved all the error and tests are also passing now. Can you please have a look?

Thanks

PR #1824, opened 2025-07-22
chore: add tests for studio and preview commands.

### LINKED ISSUE #1693: [FEATURE] Addition of tests for the commands that depend on browser interaction.

neoandmatrix:
**Description**

The following PR adds a process to test the command that require a browser interaction. Currently 2 such commands are available `start studio` and `start preview`. Puppeteer is  being used for this purpose to silently open and test the browser instances of these commands.

Some additional change was required for this to be achieved and that is addition of a `-B` flag that makes these command not to start a browser by default that is only start the server and then it's upto user to navigate to that url manually to see the instance of studio. This was required as opening the browser by default will start another process in the system that  will interfere with the tests and making them unreliable.

Since these commands start a `http` server which needs to be closed very safely and precisely as it originates as a sub-process of running tests and don't exit until forced. Hence a `/close` endpoint was introduced to safely close this server and let the tests complete without any issues. I had also tested other approaches such as `server.unwrind()` or `app.close().then()` or even using `exec` commands to shut down the server but none were working or were reliable to be used. Hence `process.exit()` was used.

At last `closeStudioServer()` was placed in the `temp-dir.test.ts` tests as the temp-dir tests run at last and the http server can only be closed after all the tests have been completed not before.

The following screenshot shows the new tests being working correctly and being included in the code coverage.

image
--- 
image
--- 
image

Also for visual representation these the elements whose presence is being tested. (The red rectangle ones.) I have tried to keep it as minimum and enough as possible for us to get to know that all the studio commands are fine and working as expected.

Screenshot From 2025-08-30 19-21-31

Screenshot From 2025-08-30 19-22-17

**Related issue(s)**

Resolves #1693


neoandmatrix:
Hey @Shurtu-gal @AayushSaini101 added tests for `studio` and `preview ` command through browser interaction and are working fine. currently had to change the `preview` command default port as during tests it was clashing with the `studio` command port which would not happen once #1815 is merged and would revert after that.

Also i tried testing the websocket also but currently found it to be unreliable and causing issues hence i suggest to skip them for now if possible as it would require quite a bit of testing and debugging along with mocking lots of stuff as websockets in these commands modify the file and it contents in realtime and mocking them as whole is quite difficult for now.

Rest tests are working good and seem to pass here also.

Please have a look and i believe as i have introduced a new flag documentation might also be required to be updated ?

Thanks.

aeworxet:
@asyncapi/bounty_team

neoandmatrix:
Hey @AayushSaini101 have added the tests for both the commands. Can you please have a look and let me know if other changes are required?

Also later I realised that testing the websocket connection specifically and explicitly is of not benefit as when the studio starts from websocket only it gets the contents of the passed file. So if it is starting successfully and connected then it already means that the websocket is working fine. And also testing those explicitly would require setting up server again and we already have 3 servers during the tests and adding another one would not only complicate managing those and also slow down tests.

Shurtu-gal (approved):
LGTM. Thanks for the hard work.

AayushSaini101 (approved):
LGTM : )

PR #1832, opened 2025-08-01
feat: add --compile flag support to generate fromTemplate command Issue#1829

### LINKED ISSUE #1829: Enable `--compile` for `generate fromTemplate` command
[This](https://github.com/asyncapi/cli/blob/master/src/commands/generate/fromTemplate.ts#L89) needs to be extended with new flag called: `--compile`

Would be good to also bump to use latest generator: `2.7.1`

It's going to work only with `AsyncAPINewGenerator` that has it enabled.

By default `compile` option in the library is set to `true` and would be great to allow people to pass `--compile=false`. This should speed up generation for templates like `@asyncapi/html-template` that are released with transpiled files and compilation is not needed.

sarthakNITT:
**Description**

fixed #1829
added compile flag in ./src/commands/generate/fromTemplate.ts
added compile flag in ./src/commands/core/flags/generate/fromTemplate.flags.ts

### DIFF
--- a/.changeset/1832.md
+++ b/.changeset/1832.md
@@ -1,0 +1,10 @@
+---
+'@asyncapi/cli': minor
+---
+
+feat: add --compile flag support to generate fromTemplate command Issue#1829
+
+- e7b086d: bump modelina-cli #1818
+- f1b604c: fixed issue#1829
+
+


sarthakNITT:
Hi @AayushSaini101,

Can you please review the pr.

derberg (commented):
hey, I'm not a maintainer here, just reported an issue but I can already give you feedback that maintainers will for sure do as well:

- `compile` is not an entirely new flag, it should be passed into generator like any other generator option through `options`
- you are mixing 2 issues in one PR which is never good. Focus on compile and remove modelina-cli upgrade to a separate PR

PR #1846, opened 2025-08-27
ci: update label check to meet api spec

Shurtu-gal:
**Description**

- Turns out PR labels cannot be filtered normally.
- https://docs.github.com/en/actions/reference/workflows-and-actions/expressions#example-using-an-object-filter


Shurtu-gal:
@AayushSaini101 could you review this once?

PR #1868, opened 2025-10-06
fix:output flag renamed and writing to file functionality added to diff command 

### LINKED ISSUE #1656: [BUG] inconsistent uses of `-o` flag in some commands
### Describe the bug.

In some commands like `asyncapi bundle`, `asyncapi format` etc the -o flag is written explicitly in flags file while in below 3 commands it is used from `validationFlags()` . 
According to the code the output flag inside `validationFlags()` is responsible to write any diagnostics (if present) in file but the description of this flag is - 'The output file name. Omitting this flag the result will be printed in the console.' which might confuse user with other commands output flags.

1. `asyncapi diff OLD NEW` - In this command we have 2 flags with -o 
  -o, --output=<value>               The output file name. Omitting this flag the result will be printed in the console.
  -o, --overrides=<value>            path to JSON file containing the override properties

Case 1 - Write only -o 

Image

It is assuming that I am trying to use --overrides

Case 2 - Provide --output as flag

Image

It is printing correct output in logs but no file was created.

2. `asyncapi validate [SPEC-FILE] [PROXYHOST] [PROXYPORT]` -  
In this command if we use -o flag than if their are any diagnostics than it will be printed in file otherwise file will be empty

Case 1 : Empty File

Image

Case 2: If Diagnostics present than shown in output 

Image

So command is working fine but in documentation it is written: 
  -o, --output=<value>               The output file name. Omitting this flag the result will be printed in the console.
But about validation of file is still written in logs only because of `logGovernanceMessage()` in file `src\core\parser.ts`

3. `asyncapi generate models LANGUAGE FILE` 

The usage of -o flag in this command is working absolutely fine a folder is created and which contains model file.

Image

But this is created because of `ModelinaFlags` it also has output flag and was not created with help of `validationFlags` function.

### Expected behavior

Solutions I am thinking to fix the issues - 

1. I think we should change field `output` inside `validationFlags` to `output-diagnostics` and flag from `-o` to `--output-diagnostics` so user dont get confuse.

2.  For `asyncapi diff OLD NEW`  command we will have 3 flags `--output-diagnostics` for write diagnostics in file , `-o` for writing console print inside file and `-O` (maybe some other flag becuase all flags are in lowercase we can discuss) for override operation.

3. For  `asyncapi validate [SPEC-FILE] [PROXYHOST] [PROXYPORT]` command the output print in console is simple and we don't require it to copy in file so this command will have only ` --output-diagnostics` flag (which is currently present with `-o`).

4. For  `asyncapi generate models LANGUAGE FILE`  command -o is working correctly because of `ModelinaFlags` so if we just do above changes it will also have 2 different flags `-o` and `--output-diagnostics`.

### Screenshots

Provided above.

### How to Reproduce

Already mention above.

### 🥦 Browser

Google Chrome

No

neoandmatrix:
**Description**

This PR introduces the changes to fix the inconsistent naming of flags in commands `diff` , `validate` and `generate models`. The `-o` or `--output` flag present in these was inherited from `parserFlags()` and these commands also had a separate declaration and usage of `-o`  and `--output` flags among themselves and thus creating a confusion. 

With this PR the `-o` flag from `parserFlags()` is renamed to `-s` or `--save-output` which makes the name more verbose and descriptive along with eliminating any conflicts with declaration of flags with other commands.

In the `asyncapi diff` command which uses `parserFlags` and hence inherits the `-s` flag for saving the output functionality but previously lack implementation of it. Hence that was also added with this PR.

**Related issue(s)**

Resolves #1656

### DIFF
--- a/src/apps/cli/commands/diff.ts
+++ b/src/apps/cli/commands/diff.ts
@@ -53,6 +53,7 @@ export default class Diff extends Command {
     let markdownSubtype = flags['markdownSubtype'];
     const watchMode = flags['watch'];
     const noError = flags['no-error'];
+    const writeOutput = flags['save-output'];
     let firstDocument: Specification, secondDocument: Specification;
 
     checkAndWarnFalseFlag(outputFormat, markdownSubtype);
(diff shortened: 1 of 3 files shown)


neoandmatrix:
Hey @Shurtu-gal can you please have a look at the new flags and suggest if some more enhancements can be done.

Thanks.

neoandmatrix:
@Shurtu-gal added tests and also tested them locally, works fine. Please have a look. 
Thanks.

Shurtu-gal (approved):
Thanks for the awesome work @neoandmatrix

PR #1875, opened 2025-10-20
feat: add new custom resolver to fetch the reference from private repo

AayushSaini101:
Resolves: #1796 

* Create a custom resolver to fetch the reference from the private repo 
* Add new command auth add for storing the repo name and repo token. 
* Tests and documents need to be updated

### DIFF
--- a/.changeset/1875.md
+++ b/.changeset/1875.md
@@ -4,5 +4,13 @@
 
 feat: add new custom resolver to fetch the reference from private repo
 
-- 88ceb3d: add new custom resolver to fetch the reference from private repo
+- 88ceb3d:  add new custom resolver to fetch the reference from private repo
+- 4974358: Update src/domains/services/validation.service.ts
+
+Co-authored-by: Fran Méndez <fmvilas@gmail.com>
+- c764ee6: update instruction in the test case for the validation success
+- 2b9457f: fix test-cases
+- 86bbfc4: fix lint issue across the project
+- 6813ef0: Update 1875.md
+
 


test/unit/services/validation.service.test.ts
```diff
@@ -129,4 +172,81 @@ describe('ValidationService', () => {
...
+
+      const specFile = new Specification(asyncAPIWithPrivateGitHubRef);
+      const options = {
+        'diagnostics-format': 'stylish' as const
+      };
+
+      const result = await validationService.validateDocument(specFile, options);
+
+      // The validation succeeds but the document is invalid due to unresolved ref
+      expect(result.success).to.equal(true);
```
AayushSaini101: @fmvilas success is true defines the command is successfully executed

fmvilas:
Please next time don't force-push 🙏  It completely removes the whole history and makes it harder to review.

src/domains/services/validation.service.ts
```diff
@@ -32,6 +32,123 @@ import { ParseOptions } from '@asyncapi/parser';
...
+      'User-Agent': 'AsyncAPI-CLI',
+    };
+
+    const authInfo = await ConfigService.getAuthForUrl(url);
+
+    if (isValidGitHubBlobUrl(url)) {
+      url = convertGitHubWebUrl(url);
+    }
+
+    // Only require authentication for GitHub URLs
```
fmvilas:
```suggestion
```

aeworxet:
@asyncapi/bounty_team

Shurtu-gal:
@fmvilas 
> Also, think about it, "validation succeeds" but it's "invalid" (i.e., not valid). Isn't it a contradiction?

So, how all services currently work, is if there is a uncaught or unresolveable error it returns an error response. The `success` basically means that the service ran properly. Inside the data the status of the document is shown if its `valid | invalid`.

This can be changed though if you have some suggestions.

You can see it here:
https://github.com/asyncapi/cli/blob/master/src/domains/services/base.service.ts#L1-L25

fmvilas:
> @fmvilas
> 
> > Also, think about it, "validation succeeds" but it's "invalid" (i.e., not valid). Isn't it a contradiction?
> 
> So, how all services currently work, is if there is a uncaught or unresolveable error it returns an error response. The `success` basically means that the service ran properly. Inside the data the status of the document is shown if its `valid | invalid`.
> 
> This can be changed though if you have some suggestions.
> 
> You can see it here: https://github.com/asyncapi/cli/blob/master/src/domains/services/base.service.ts#L1-L25

I see it's not an issue of this PR but it's definitely weird, don't you think so? Think about the devs than don't know the code base and are only using the CLI.

Shurtu-gal:
> I see it's not an issue of this PR but it's definitely weird, don't you think so? Think about the devs than don't know the code base and are only using the CLI.

- Agree with the weirdness part, any suggestions on how it could be improved.
- Regarding the devs statment, this success or failure is not shown to CLI users. If the doc is invalid it just shows the doc is invalid. The service output is irrelevant here I suppose?

image

fmvilas:
Yeah, if CLI users are not suffering it, that's fine. However, I think it's a fine line. If you're not aware of this difference, it's easy to push wrong code in the future.

As per how to fix it, I'd just make it fail. If something doesn't resolve, trigger the error. A call to a method called `validate()` should return `success: true` if, and only if, the document is valid. If a reference is not resolved, then the document isn't valid, right? Or, at the very least, we couldn't finish the validation. This reminds me of people responding with 200 OK and `{ "error": "..." }` in HTTP APIs. If there's an error, is not OK, so don't use the 200 code. Same here.

The above is if you want to preserve the current API. I'd go one step beyond and simply remove the `success` field altogether. If the call to `validate()` doesn't throw an error, it means it's valid. Otherwise, you can catch the error(s).

Issue #1881, opened 2025-10-27
Fix Sonar Cloud issues in the CLI Project


AayushSaini101:
There are more than 10+ issues in the CLI Project that is detected by Sonar Cloud in the project, Minimise as we can in the project

AayushSaini101:
https://sonarcloud.io/project/issues?id=asyncapi_cli&pullRequest=1875&issueStatuses=OPEN,CONFIRMED&sinceLeakPeriod=true

kajal-jotwani:
Can I work on this issue?

batchu5:
I'm interested to work on this, can you assign me this issue

lightning-sagar:
@AayushSaini101 I am working on this!!

Shurtu-gal:
@AayushSaini101 could you review the PRs when you get some time. Also some of the sonar cloud issues might need intervention in sonarcloud portal itself.

Shurtu-gal:
> [@AayushSaini101](https://github.com/AayushSaini101) could you review the PRs when you get some time. Also some of the sonar cloud issues might need intervention in sonarcloud portal itself.

@AayushSaini101 could you please review the PRs once as well the portal for false positives (which most of them are 😓)

lightning-sagar:
Hi @AayushSaini101 @Souvikns 
I noticed that a similar PR addressing this issue was recently merged... since I opened mine earlier, i wanted to check if there is anything missing or incorrect in my implementation...
It will be great if there is any feedback so I can improve and better align with the issue expectations... I mainly want to learn how I can do better next time...

AayushSaini101:
> > [@AayushSaini101](https://github.com/AayushSaini101) could you review the PRs when you get some time. Also some of the sonar cloud issues might need intervention in sonarcloud portal itself.
> 
> [@AayushSaini101](https://github.com/AayushSaini101) could you please review the PRs once as well the portal for false positives (which most of them are 😓)

@Shurtu-gal sorry i will check thanks

Shurtu-gal:
Closed this by mistake when I tagged #1881 instead of #1885

dasakash26:
@Shurtu-gal can i work on these sonar cloud issues??

tanish35:
@AayushSaini101 @Shurtu-gal Can I also fix some of the issues present

Shurtu-gal:
Please avoid linking issue to pr this closing it.

DarshanCode2005:
[...]

PR #1917, opened 2025-12-05
fix: remove unnecessary await from startPreview call

ashutoshsao:
This PR resolves a SonarCloud code quality issue as part of the effort to fix SonarCloud issues reported in https://github.com/asyncapi/cli/issues/1881.

**Description**

- Removed unnecessary `await` keyword from `startPreview` call in the preview command
- The `startPreview` function returns `void`, not a Promise, so awaiting it was incorrect and triggered a linter error: "Unexpected `await` of a non-Promise (non-"Thenable") value"
- This change aligns the preview command with the pattern used in the studio command, where `startStudio` (also returning `void`) is called without `await` on line 58 of `studio.ts`

**Testing**

The change is minimal and syntactically correct:
- Removed `await` from line 40 in `src/apps/cli/commands/start/preview.ts`
- The function signature in `Preview.ts` confirms `startPreview` returns `void`
- The change matches the existing pattern in `studio.ts` where `startStudio` is called without `await`
- No functional changes - the preview command behavior remains the same, only the linter error is resolved

**Documentation**

No documentation updates required:
- The documentation in `docs/usage.md` is auto-generated from the command definitions
- This is an internal code fix that doesn't change the user-facing API or behavior
- The command usage, flags, and descriptions remain unchanged

**Related issue(s)**

Fixes https://github.com/asyncapi/cli/issues/1881


Souvikns:
Adding the permalink of the function definition of `startPreview` 

https://github.com/asyncapi/cli/blob/33a8dc30ccbd93d5f877a104eed83a212517e678/src/domains/models/Preview.ts#L36

ashutoshsao:
:)

PR #1923, opened 2025-12-12
fix: remove lefthook

Shurtu-gal:
**Description**

- Remove Lefthook. It hampers productivity due to tests taking too long, leading to disabling of the same.
- Also creates multitudes of CI problems. Example: https://github.com/asyncapi/cli/actions/runs/20174523790/job/57920469719#step:9:43

**Related issue(s)**


AayushSaini101 (approved):
LGTM

PR #1925, opened 2025-12-15
fix: update changeset action and cli

Shurtu-gal:
**Description**

- Release workflow still doesn't work. Trusted publishing is setup correctly.
- PR to check if updating changesets version fixes it.

**Related issue(s)**


Shurtu-gal:
Unsetting node token as has been suggested in https://github.com/actions/setup-node/issues/1440 and https://github.com/actions/setup-node/issues/1445#issuecomment-3628907108

PR #1953, opened 2026-01-05
feat: enable --compile flag

### LINKED ISSUE #1829: Enable `--compile` for `generate fromTemplate` command

Siddhi-sahu:
**Description**
hopefully fixes #1829 

I am new here, and still trying to understand the codebase.
I have opened the pr but I am not entirely sure if I am going into the right direction. Any little push would be appreciated.

Thank you for your time.
@Shurtu-gal 
**Related issue(s)**


Varadraj75:
Hi @Siddhi-sahu,

I reviewed the PR and tested it locally.

###  Looks good

* `--compile` flag is implemented correctly (default `true`, `allowNo: true`)
* Flag is properly wired into `GeneratorOptions`
* Verified locally that `--compile` and `--no-compile` appear and work as expected
* Generator version is already `3.0.1`, so the requirement is satisfied

###  Pending

* Add integration tests in `test/integration/generate/fromTemplate.test.ts`:

  * default behavior
  * `--no-compile` / `--compile=false`
* Add a changeset (`npx changeset`, minor bump for `@asyncapi/cli`)
* Please confirm whether changes to `entrypoint.sh` and `package-lock.json` are intentional

Overall, the implementation looks solid. Let me know if you need any help with the tests or changeset.

Shurtu-gal:
Please add options accordingly here https://github.com/asyncapi/cli/blob/master/src/apps/cli/internal/base/BaseGeneratorCommand.ts#L47

I don't the compile option right now does anything meaningful till it's passed to generator.

Siddhi-sahu:
thanks for your patience @Shurtu-gal . I think I understand now. You mean, compile needs to be explicitly forwarded to `AsyncAPIGenerator`? if this is correct, I'll update `GeneratorService` accordingly.

Issue #1959, opened 2026-01-08
update node version in docker file for github action


derberg:
cli v5 should use node 24

https://github.com/asyncapi/cli/blob/master/github-action/Dockerfile requires and update in line 2 and 29

should be released as patch and new action version must be released to public

dasakash26:
Hey @derberg ,

I'd like to take this one, is it just bumping the Dockerfile to node:24-alpine in both stages.
Anything I should watch out for besides the usual CONTRIBUTING.md stuff? (like specific alpine tag or extra tests)

Thanks! :)

Varadraj75:
Sounds good @dasakash26 ,
If you need help validating the Node 24 update or testing the GitHub Action after the Dockerfile change, feel free to tag me.

derberg:
@Shurtu-gal any comments from you before @dasakash26 jumps into implementation? hints on how they should test the fix?

Shurtu-gal:
For testing the changes docker image can be run directly with the arguments. Same happens in PR checks as well.

Shurtu-gal:
For local dry-run the following can be used. https://github.com/asyncapi/cli/blob/master/docs/github-action.md#local-dry-run

Varadraj75:
Hi @Shurtu-gal , 
Can I do a local dry-run to validate the changes as well?

Shurtu-gal:
> Hi [@Shurtu-gal](https://github.com/Shurtu-gal) , Can I do a local dry-run to validate the changes as well?

Sure, please go ahead. Tests also went well https://github.com/asyncapi/cli/actions/runs/20814394014/job/59786351237, let's hope release works  🤞🏻

Shurtu-gal:
Another miss from my side 🙇🏻 
@dasakash26 could you please create a similar PR for other Dockerfiles as well https://github.com/search?q=repo%3Aasyncapi%2Fcli+path%3ADockerfile&type=code.

Please add changeset for patch release so that release is triggered.

Varadraj75:
Hi @Shurtu-gal ,

I would be happy to help with updating the other Dockerfiles and adding the changeset for the patch release, if that’s okay.
Please let me know if I can take this up or if there’s anything specific you’d like me to keep in mind while working on it.

Thanks!

Shurtu-gal:
You can go ahead @Varadraj75

dasakash26:
Hey @Shurtu-gal ,
[...]

PR #1960, opened 2026-01-08
fix: update server-api image to node 24

dasakash26:
Updates the GitHub Action Dockerfile to use node:24-alpine in both stages.
Local test: the image builds fine (local run hits the expected GITHUB_WORKSPACE error since it's outside CI, pretty normal).
closes: #1959

### DIFF
--- a/.changeset/curvy-brooms-pay.md
+++ b/.changeset/curvy-brooms-pay.md
@@ -1,0 +1,5 @@
+---
+"@asyncapi/cli": patch
+---
+
+Update server-api image to use Node 24.


Shurtu-gal (approved):
Thanks for the PR.

dasakash26:
@Shurtu-gal 
thanks a ton for the approval and help!
tbh i'm still a bit confused about the whole thing, like how to find more issues i can actually handle or figure out the other repos. where do people usually ask noob questions? slack #11_contributing maybe?
thanks again, wanna keep going 😅

Shurtu-gal:
Yeah, that would be a good start.

PR #1986, opened 2026-01-24
fix: re-add missing update-notifier plugin dependency

### LINKED ISSUE #1680: [BUG] No info on availability of new version of cli to user.
### Describe the bug.

Whenever a new version is released of the cli and if after that user runs a command then a info should be given that a new version of the cli is available. It is already implemented [here](https://github.com/asyncapi/cli/blob/master/package.json#L132) but appears to be not working as can be seen from below screenshots.

### Expected behavior

An info should be shown with new available version and current version once every day of running any commands.

### Screenshots

Image

Image

As evident form above screenshots i am using an older version and info should be displayed that a new version is available but no info is shown.

### How to Reproduce

1.  Install any older version of cli.
2. Run any command.
3. No info is displayed that a new version is available even tough it is.

### 🥦 Browser

None

abhi-03-kh:
## Description
This PR restores the `@oclif/plugin-warn-if-update-available` dependency, which was missing from `package.json`.

Although the CLI was configured to warn users about updates (in the `oclif` config), the actual plugin was not installed, causing the notification to silently fail. Re-adding this dependency restores the "Update available!" warning functionality.

Fixes #1680

## Changes
* **fix:** Added `@oclif/plugin-warn-if-update-available` to `dependencies` in `package.json`.

## Verification
1. Pull this branch and install dependencies (`npm install`).

2. Modify `package.json` to force a "needs update" state:
   * Set `"version"` to `"1.0.0"`.
   * Set `oclif.warn-if-update-available.frequency` to `0` (to force an immediate check).
  
3. Run the CLI with `CI=false` (to bypass CI silence checks):
   ```bash
   CI=false ./bin/run_bin --version


abhi-03-kh:
@AayushSaini101 @Shurtu-gal  , any feedback on this when you have time?

PR #1994, opened 2026-01-29
feat(perf): lazy-load studio via dynamic imports to reduce install size

### LINKED ISSUE #1752: [FEATURE] Reduce number of dependencies and install size (v2.17 needs 1700 dependencies, takes 1.7Gb of space)

abhi-03-kh:
## Problem
The CLI currently bundles `@asyncapi/studio` (~1GB+) as a hard dependency. This forces CI/CD users and automation pipelines to download a full React/Next.js web application that they never use.

## Solution
- Moved `@asyncapi/studio` from `dependencies` to `optionalDependencies` in `package.json`.
- Refactored `src/domains/models/Studio.ts` and `Preview.ts` to use dynamic `import()` logic.
- Added graceful error handling: if a user runs `start studio` without the package, they receive a helpful prompt to install it instead of a crash.

## Results
| Scenario | Install Size | Notes |
|----------|------|-------|
| **Before** (Hard Dep) | ~1.6 GB | All users pay the cost |
| **After** (Lite) | **698 MB** | **~57% Reduction** 📉 |
| **After** (With Studio) | ~1.6 GB | Only users who need it pay the cost |

## Verification
- [x] `npm install --omit=optional` results in a 698MB install size.
- [x] `asyncapi validate` runs successfully without Studio installed.
- [x] `asyncapi start studio` fails gracefully with a helpful error message when missing.
- [x] `asyncapi start studio` works correctly (launches v1.1.0) when the package is manually installed.
- [x] Applied fixes for WebSocket handling and Preview command stability.

## Breaking Change
Users who rely on `asyncapi start studio` without explicitly installing `@asyncapi/studio` will now see an error message instructing them to install it, rather than a crash.

Closes #1752

### DIFF
--- a/.changeset/lazy-studio-poc.md
+++ b/.changeset/lazy-studio-poc.md
@@ -1,0 +1,5 @@
+---
+"@asyncapi/cli": minor
+---
+
+feat: lazy-load studio to reduce install size by 57%
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@asyncapi/cli",
-  "version": "5.0.5",
+  "version": "5.0.6",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@asyncapi/cli",
-      "version": "5.0.5",
+      "version": "5.0.6",
       "hasInstallScript": true,
       "license": "Apache-2.0",
       "dependencies": {
@@ -22,7 +22,6 @@
         "@asyncapi/problem": "^1.0.0",
         "@asyncapi/protobuf-schema-parser": "^3.6.0",
         "@asyncapi/raml-dt-schema-parser": "^4.0.24",
-        "@asyncapi/studio": "^1.1.0",
         "@clack/prompts": "^0.11.0",
         "@oclif/core": "^4.8.0",
         "@oclif/plugin-autocomplete": "^3.2.39",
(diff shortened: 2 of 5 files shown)


Stanzilla:
I'd be willing to take over the work/MR if the maintainers agree with the approach @AayushSaini101 et all

Issue #1987, opened 2026-01-24
[BUG] Request body validation is skipped for some paths or HTTP methods


ScriptShah:
### Describe the bug.

When using request validation in the CLI, request body validation is skipped or reported as unsupported for certain paths or HTTP methods, even when a valid request body schema is defined for the operation.

As a result, invalid request bodies may pass validation without errors, or the CLI may indicate that validation is not supported, leading to incomplete or misleading validation results.

### Expected behavior

If a request body schema is defined for an operation, the CLI should validate the request body against that schema for all supported paths and HTTP methods.

Invalid request bodies should consistently fail validation with clear and actionable error messages.

### Screenshots

N/A – this is a validation logic issue and does not produce visual UI output.

### How to Reproduce

1. Create or use an AsyncAPI document with a defined request body schema.
2. Select an operation that maps to a path or HTTP method currently reported as unsupported by the validator.
3. Run the CLI request validation using a request body that violates the schema.
4. Observe that validation is skipped or reported as unsupported instead of failing.

### 🖥️ Device Information [optional]

- Operating System (OS): Windows 10 / Linux / macOS
- AsyncAPI CLI version: latest
- Node.js version: 22.x

### 👀 Have you checked for similar open issues?

- [x] I checked and didn't find similar issue

### 🏢 Have you read the Contributing Guidelines?

- [x] I have read the [Contributing Guidelines](https://github.com/asyncapi/.github/blob/master/CONTRIBUTING.md)

### Are you willing to work on this issue ?

Yes I am willing to submit a PR!

1164148-7-eng:
/attempt #1987

I've claimed this issue. Working on it now! 🚀

l8888888:
I've fixed this issue in PR #2128

The problem was:
1. Unsafe access to `requestBody.content['application/json']` causing undefined errors
2. Incorrect error thrown when no request body schema is defined

The fix:
- Uses optional chaining to safely check content type
- Skips validation (instead of erroring) when no schema exists
- This is correct behavior - not all endpoints require request bodies

PR #2128 now fixes both #1940 and #1987, completing the entire MICROGRANT #2125. ✅

zanabalofficial:
I'd like to work on this request body validation bug. I can investigate why validation is skipped for certain paths/methods and fix the logic to ensure consistent validation across all operations. Let me know!

1164148-7-eng:
/attempt #1987

I've claimed this issue. Working on it now! 🚀

[...]

Issue #1940, opened 2025-12-28
[BUG] CLI fails for GitHub URLs with slash-based branches and multi-dot spec files


akash202004:
### Describe the bug.

<p>The CLI has two validation bugs causing valid inputs to fail.</p>

<h3>1. GitHub URL Parsing</h3>
<ul>
  <li>Regex assumes branch names do not contain <code>/</code></li>
  <li>Fails for valid URLs like:</li>
</ul>

<pre>
https://github.com/org/repo/blob/feature/new-validation/spec.yaml
</pre>

<ul>
  <li>Incorrect parsing leads to 404 errors</li>
</ul>

<h3>2. File Extension Detection</h3>
<ul>
  <li>Uses <code>name.split('.')</code></li>
  <li>Fails for:</li>
</ul>

<pre>
my.asyncapi.yaml
asyncapi
</pre>

<h3>Expected Behavior</h3>
<p>All valid GitHub URLs and specification files should be accepted.</p>

<h3>Actual Behavior</h3>
<p>Valid inputs are rejected or cause runtime failures.</p>

### Expected behavior

<h3>Expected behavior</h3>
<p>The CLI should correctly accept and process:</p>
<ul>
  <li>GitHub file URLs from <strong>any valid branch</strong>, including branches with slashes (e.g. <code>feature/new-validation</code>)</li>
  <li>Specification files with <strong>multiple dots</strong> in the filename (e.g. <code>my.asyncapi.yaml</code>)</li>
  <li>Filenames <strong>without extensions</strong> without crashing or behaving unpredictably</li>
</ul>
<p>Valid inputs should not be rejected during validation.</p>

### Screenshots

<h3>Screenshots</h3>
<p>N/A — this is a validation and parsing logic issue. The failure occurs during URL parsing and file validation, not in the UI.</p>

### How to Reproduce

<h3>How to reproduce</h3>
<ol>
  <li>
    Run the CLI with a GitHub file URL from a branch containing a slash, for example:
    <pre>https://github.com/org/repo/blob/feature/new-validation/spec.yaml</pre>
  </li>
  <li>
    Observe that the CLI fails due to incorrect branch parsing and returns a <code>404</code> error.
  </li>
  <li>
    Run the CLI with a specification file named:
    <pre>my.asyncapi.yaml</pre>
  </li>
  <li>
    Observe that the file is rejected due to incorrect extension detection.
  </li>
</ol>

### 🖥️ Device Information [optional]

- Operating System (OS):
[...]

Issue #2125, opened 2026-04-19
Set of issues for resolution under the MICROGRANT Program 2026-05


AayushSaini101:
[MICROGRANT] Set of issues for resolution under the MICROGRANT Program 2026-04

This is an Aggregated MICROGRANT Issue consisting of GitHub issues

https://github.com/asyncapi/cli/issues/1940
https://github.com/asyncapi/cli/issues/1987

that are subject to resolution in the MICROGRANT Program's calendar quarter round 2026-05 .

1164148-7-eng:
/attempt #2125

I've claimed this issue. Working on it now! 🚀

1164148-7-eng:
/attempt #2125

I've claimed this issue. Working on it now! 🚀

aeworxet:
@1164148-7-eng, please read https://github.com/asyncapi/community/issues/2295#issuecomment-3963443341. Attempts to force-push through will not increase your chances of participating in the Bounty Program.

aeworxet:
### Bounty Issue's service comment

Text labels: `bounty/2026-05`, `bounty/medium`, `bounty/coding`
First assignment to regular contributors: [2026-04-25 00:00:00 UTC+12:00](https://time.lol/#2026-04-25T00:00:00+12:00)
End Of Life after: [2026-05-31 23:59:59 UTC-12:00](https://time.lol/#2026-05-31T23:59:59-12:00)

@asyncapi/bounty_team

##### The Bounty Program is not a Mentorship Program. The accepted level of Bounty Program Participants is **Middle**/**Senior**.
##### Regular contributors should explain in meaningful words how they are going to approach the resolution process when expressing a desire to work on this Bounty Issue.

l8888888:
## ✅ MICROGRANT #2125 Complete

I've submitted PR #2128 that fixes **both** issues in this MICROGRANT:

### Fixed Issues:
1. ✅ #1940 - GitHub URL parsing + file extension detection
2. ✅ #1987 - Request body validation

### Summary of Fixes:

**#1940 - GitHub URL & File Extensions**:
- Support branch names with slashes (e.g., `feature/new-validation`)
- Handle multi-dot filenames (e.g., `my.asyncapi.yaml`)
- Proper handling of files without extensions

**#1987 - Request Body Validation**:
- Safe access to request body content types
- Correct behavior when no schema is defined
- No more false "unsupported" errors

All bugs are now resolved in a single PR. Ready for review! 🚀

**PR**: #2128

[...]

PR #2128, opened 2026-04-21
fix: complete MICROGRANT #2125 - URL parsing, file extensions, and request validation

l8888888:
## Description

Fixes #1940 and #1987 (Complete MICROGRANT #2125)

This PR addresses **all bugs** in the MICROGRANT issue #2125:

### 1. GitHub URL Parsing Bug (#1940)

**Problem**: The regex pattern assumed branch names don't contain `/`, causing failures for valid URLs like:
```
https://github.com/org/repo/blob/feature/new-validation/spec.yaml
```

**Solution**: Rewrote the URL parsing logic to:
- Parse owner, repo, and the combined branch+path
- Work backwards from the filename to separate branch from file path
- Correctly handle branch names with slashes

### 2. File Extension Detection Bug (#1940)

**Problem**: Used `name.split('.')[1]` which fails for:
- Files with multiple dots: `my.asyncapi.yaml` → incorrectly gets "asyncapi" instead of "yaml"
- Files without extensions: `asyncapi` → undefined, causes crash

**Solution**: 
- Use `.pop()` to get the last segment after splitting by `.`
- Add proper undefined check for files without extensions
- Convert to lowercase for case-insensitive comparison

### 3. Request Body Validation Bug (#1987)

**Problem**: Request body validation was skipped or reported as "unsupported" for certain paths/methods, even when a valid schema was defined.

**Root causes**:
1. Unsafe access to `requestBody.content['application/json']` caused undefined errors
2. Missing schema incorrectly threw "validation not supported" error instead of skipping

**Solution**:
- Added safe access with optional chaining (`content?.['application/json']`)
- Changed logic to skip validation (not error) when no schema is defined
- This is correct behavior - not all endpoints require request bodies

## Changes

- `src/domains/services/validation.service.ts`: Updated GitHub URL parsing logic
- `src/domains/models/SpecificationFile.ts`: Fixed file extension detection
- `src/apps/api/middlewares/validation.middleware.ts`: Fixed request body validation logic

## Testing

The fixes handle these cases correctly:
- ✅ `https://github.com/org/repo/blob/feature/new-validation/spec.yaml`
- ✅ `my.asyncapi.yaml`
- ✅ `asyncapi` (properly rejected as invalid)
- ✅ Request body validation for all paths/methods
- ✅ No false "unsupported" errors

## Checklist

- [x] Fixes all bugs in MICROGRANT #2125
- [x] Addresses #1940 (GitHub URL + file extension)
- [x] Addresses #1987 (request body validation)
- [x] Maintains backward compatibility
- [x] Follows existing code style
- [x] Ready for review

### DIFF
--- a/.changeset/fix-github-url-parsing.md
+++ b/.changeset/fix-github-url-parsing.md
@@ -1,0 +1,5 @@
+---
+"@asyncapi/cli": patch
+---
+
+fix: support slash-based branch names and multi-dot file extensions in GitHub URL parsing
--- a/src/apps/api/middlewares/validation.middleware.ts
+++ b/src/apps/api/middlewares/validation.middleware.ts
@@ -54,13 +54,16 @@ async function compileAjv(options: ValidationMiddlewareOptions) {
 
   const requestBody = method.requestBody;
   if (!requestBody) {
-    return;
+    return; // No request body defined - validation not needed
   }
 
-  let schema = requestBody.content['application/json'].schema;
-  if (!schema) {
-    return;
+  // Check if application/json content type exists
+  const jsonContent = requestBody.content?.['application/json'];
+  if (!jsonContent || !jsonContent.schema) {
+    return; // No JSON schema defined - validation not needed
   }
+
+  let schema = jsonContent.schema;
 
   schema = { ...schema };
   schema['$schema'] = 'http://json-schema.org/draft-07/schema';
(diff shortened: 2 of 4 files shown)


l8888888:
## Note about MICROGRANT #2125

I noticed that #1940 is part of the MICROGRANT issue #2125, which also includes #1987.

This PR currently fixes **only #1940** (GitHub URL parsing and file extension bugs).

**Question for maintainers**: 
- Should I also fix #1987 (request body validation) in this PR to complete the MICROGRANT?
- Or should #1987 be handled separately?

I'm happy to extend this PR to include both fixes if that's preferred. Please let me know! 🙏

PR #2143, opened 2026-04-24
fix: add 10s timeout and HEAD request to registry validation (#2027)

Angelebeats:
Fix [BUG] CLI hangs indefinitely when --registry-url points to an unreachable host (#2027). 

/claim #2027


Angelebeats:
??????????,??????????????????,??!

Issue #1839, opened 2025-08-22
[BUG] external files in same directory are not resolved via $ref anymore, since v3.3.0


pstorch:
### Describe the bug.

We've externalized parts of the asyncapi contract into separate files and reference them via `$ref`. 
These files are located in the same directory.

src/contract/asyncapi.yaml:

```
...
components:
  messages:
    MyMessage:
      title: "My message"
      traits:
        - $ref: "my-event-headers.yaml#/components/messageTraits/event-headers-definition"
      headers:
        type: object
        properties:
          x_origin:
            const: my_project
          x_event_type:
            allOf:
              - $ref: "my-event-models.yaml#/components/schemas/EventTypeName"
              - const: EVENT_TYPE_NAME
          x_grouping_key:
            $ref: "my-event-models.yaml#/components/schemas/Id"
      payload:
        $ref: "my-event-models.yaml#/components/schemas/MyPayload"
      bindings:
        kafka:
          key:
            $ref: "my-event-headers.yaml#/components/schemas/GroupingKeyHeader"
...
```

Since asyncapi-cli v3.3.0 this doesn't work anymore if all files are in a subdirectory like `src/contract`. 
As a workaround we have to copy all files into the current working directory. Then it works.
I've also tried to prefix the path reference with `./`, but it didn't work either.

### Expected behavior

Read all referenced files in the same directory as in version 3.2.0

### Screenshots

n.a.

### How to Reproduce

Create an asyncapi contract in a subdirectory and reference other yaml files in the same directory.
Run the asyncapi-cli generator.

### 🖥️ Device Information [optional]

- Operating System (OS):
- Browser:
- Browser Version:

### 👀 Have you checked for similar open issues?

- [x] I checked and didn't find similar issue

### 🏢 Have you read the Contributing Guidelines?

- [x] I have read the [Contributing Guidelines](https://github.com/asyncapi/.github/blob/master/CONTRIBUTING.md)

### Are you willing to work on this issue ?

None

einstux:
We are also hit by this issue.

Kartikayy007:
Picking up this issue

Shurtu-gal:
This issue still exists. @neoandmatrix could you please put the screenshots here.

neoandmatrix:
> This issue still exists. [@neoandmatrix](https://github.com/neoandmatrix) could you please put the screenshots here.

[...]

PR #2144, opened 2026-04-24
fix: resolve $ref for external files (#1839)

Angelebeats:
Fix [BUG] external files in same directory are not resolved via $ref since v3.3.0 (#1839). 

/claim #1839


Angelebeats:
??????????,??????????????????,??!

PR #2196, opened 2026-05-12
fix: handle slash branches in GitHub URLs and multi-dot file extensions

chengyixu:
## Summary
- Fixes GitHub blob URL parsing to support branch names containing slashes (e.g., `feature/new-validation`)
- Fixes file extension detection to correctly handle multi-dot filenames (e.g., `my.asyncapi.yaml`)

Fixes #1940
Part of microgrant #2125

### 1. GitHub URL branch parsing (`src/domains/services/validation.service.ts`)
The regex `([^\/]+)` for branch names did not allow slashes, causing URLs like `https://github.com/org/repo/blob/feature/new-validation/spec.yaml` to fail with 404 errors.

**Fix**: Match everything after `/blob/` as a single string, then generate candidate branch/path splits. Try each candidate progressively until one succeeds:
- Candidate 0: branch=`main`, path=`src/file.yaml` (most common case, works for simple branches)
- Candidate 1: branch=`feature/new-validation`, path=`spec.yaml` (works for slash branches)

The resolver's `read()` method now tries each candidate in sequence, stopping at the first successful fetch.

### 2. File extension detection (`src/domains/models/SpecificationFile.ts`, `src/apps/cli/commands/new/file.ts`)
The old code used `name.split('.')[1]` which returns the second segment after splitting by dots. For `my.asyncapi.yaml`, this returned `asyncapi` instead of `yaml`.

**Fix**: Use `name.split('.')[parts.length - 1]` to get the actual file extension (last segment after the final dot).

## Test plan
- [x] All existing unit tests pass (60/60)
- [x] All existing integration tests pass (31/31 validate, 2/2 new file)
- [x] Verified regex produces correct candidate URLs for slash branches
- [x] Verified extension detection correctly identifies `yaml` in `my.asyncapi.yaml`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/apps/cli/commands/new/file.ts
+++ b/src/apps/cli/commands/new/file.ts
@@ -175,7 +175,8 @@ export default class NewFile extends Command {
     if (!fileName.includes('.')) {
       fileNameToWriteToDisk = `${fileName}.yaml`;
     } else {
-      const extension = fileName.split('.')[1];
+      const parts = fileName.split('.');
+      const extension = parts[parts.length - 1];
 
       if (extension === 'yml' || extension === 'yaml' || extension === 'json') {
         fileNameToWriteToDisk = fileName;
--- a/src/domains/models/SpecificationFile.ts
+++ b/src/domains/models/SpecificationFile.ts
@@ -237,11 +237,12 @@ export async function fileExists(name: string): Promise<boolean> {
       return true;
     }
 
-    const extension = name.split('.')[1];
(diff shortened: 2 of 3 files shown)


chengyixu:
Closing in favor of #2197 which has the cleaner splitBranchAndPath approach. The new/file.ts fix from this PR will be ported to #2197.

PR #2205, opened 2026-05-19
fix: request body validation is skipped for some paths or HTTP methods

AayushSaini101:
Related: https://github.com/asyncapi/cli/issues/1987

PR #2225, opened 2026-06-03
fix(api): skip request body validation when endpoint has no requestBody schema

### LINKED ISSUE #1987: [BUG] Request body validation is skipped for some paths or HTTP methods

q404365631:
## What
Fixes the case where `validationMiddleware()` reported a false-positive "request body validation is not supported for this path/method" error for endpoints that legitimately have no `requestBody` schema (e.g. `GET` or `DELETE` endpoints, or `POST` endpoints that don't take a JSON body).

## Why
Looking at `compileAjv()` in `validation.middleware.ts`:
- It returns `undefined` either when `method.requestBody` is missing (line 57), or
- when `requestBody.content['application/json'].schema` is missing (line 62).

Both cases mean: **there is no JSON body to validate**, and the request should pass through.

But the middleware then treated any falsy `validate` value as a hard error and threw a 422 with the misleading detail `Request body validation is not supported for "<path>" path with "<method>" method.` — which is exactly what #1987 reports.

The fix: explicitly check for `validate === undefined` (the documented signal for "no body to validate") and `next()` instead of throwing.

## Changes
- `src/apps/api/middlewares/validation.middleware.ts`: Replaced the broad `if (!validate)` error branch with a narrow `if (validate === undefined)` skip branch that calls `next()`.

1 file changed, 4 insertions(+), 7 deletions(-).

Closes #1987

Part of MICROGRANT 2026-05 round (`#2125`).

### DIFF
--- a/.changeset/fix-request-body-validation-1987.md
+++ b/.changeset/fix-request-body-validation-1987.md
@@ -1,0 +1,5 @@
+---
+"@asyncapi/cli": patch
+---
+
+fix(api): skip request body validation when the endpoint has no `body` parameter defined in its OpenAPI spec, avoiding spurious validation errors.
--- a/src/apps/api/middlewares/validation.middleware.ts
+++ b/src/apps/api/middlewares/validation.middleware.ts
@@ -173,13 +173,10 @@ export async function validationMiddleware(
     }
 
     try {
-      if (!validate) {
-        throw new ProblemException({
-          type: 'invalid-request-body',
-          title: 'Invalid Request Body',
-          status: 422,
-          detail: `Request body validation is not supported for "${options.path}" path with "${options.method}" method.`,
-        });
+      if (validate === undefined) {
+        // `validate` is `undefined` only when the endpoint has no `requestBody`
+        // schema. There is nothing to validate, so let the request continue.
+        return next();
       }
 
       await validateRequestBody(validate, req.body);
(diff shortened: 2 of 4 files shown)


AayushSaini101:
Resolved via: https://github.com/asyncapi/cli/pull/2205 Closing the PR thanks lot for your timing thanks

PR #2283, opened 2026-08-19
feat: studio made optional

neoandmatrix:
This PR makes studio an optional package for cli to reduce its size by 500 MB. New flags are added to enable studio be installed optionally.

## Studio availability & behavior by environment

| Environment | Studio present out-of-the-box? | `start studio` / `preview` / `new --studio` | PDF (`html-template`) | Core cmds (`validate`, `convert`, `generate`, `bundle`, `diff`, `optimize`, `new`) |
|---|---|---|---|---|
| **CLI — `npm i -g @asyncapi/cli`** (dev machine) |  No (devDeps not installed for consumers) | Interactive TTY → **prompt** "~450MB, proceed?" → installs to data dir → launches. Later runs: no prompt. |  Works (via html-template's own puppeteer) | Work |
| **Docker — `asyncapi/cli` image** | No (`--omit=dev` excludes it) | No TTY → **auto-declines** with clear error unless `--yes` / `-e ASYNCAPI_STUDIO_AUTO_INSTALL=1`. With those → installs into container. |  Works out of the box (chromium kept in image) | Work |
| **CI — your CLI in a user pipeline** (non-interactive) |  No | No TTY → **auto-declines** unless `--yes` / `ASYNCAPI_STUDIO_AUTO_INSTALL=1` → installs (needs network) | Works | Work |
| **CI — this repo's own test suite** (`npm ci`) | **Yes** (devDeps installed) | Resolves the installed devDep → **no prompt**; studio integration test **runs** | Works | Work |
| **GitHub Action image** (`asyncapi/github-action-for-cli`) | No (`--production`) | Not used in actions; if invoked, no TTY → clear error | Unchanged from before (action image never bundled chromium) | Work |

Docker build and image size has also been reduced to 643 MB from previous 1.6 GB.
image


Shurtu-gal (changes requested):
Could we add a flag to the config so that studio is always installed, if the user wants in case?

Shurtu-gal (approved):

Shurtu-gal (approved):

PR #2302, opened 2026-09-20
fix: print workspace package tags from publish:trusted

princerajpoot20:
### Before merge: 
on npm, `@asyncapi/optimizer` Trusted Publisher should be repo `asyncapi/cli` and workflow `release-with-changesets.yml`.

### Why

`changesets/action` reads lines like `New tag: @asyncapi/cli@6.2.0`. That text used to live in an npm script in `package.json`. npm prints the script before it runs, so the action saw `New tag: $PACKAGE_NAME@...` and crashed. The npm publish step never ran.

This PR also sits on master after turbo was bumped 1 → 2. Turbo 2 needs a `packageManager` field to see npm workspaces. 

### What changed

- Moved tag/release helper into `scripts/publish-trusted.js`. It prints one `New tag:` line per unpublished workspace package. npm publish is still the next workflow step (`changeset publish`).
- Set `"packageManager": "npm@11.6.2"` so turbo 2 can resolve workspaces.

Related https://github.com/asyncapi/optimizer/issues/306


Shurtu-gal (approved):

aeworxet:
@asyncapi/microgrant_team

